PR #7, opened 2015-06-29
Add Jingle-pub as XEP-0358

stpeter:


linuxwolf:
Merged in changeset 7055bfa

PR #9, opened 2015-07-02
XEP-0082 errounously states that CCYY has four digits

linuxwolf:
Submitted to editors mailing list:  http://mail.jabber.org/pipermail/editor/2015-April/000330.html


stpeter:
See https://github.com/xsf/xeps/pull/17 for an alternative approach.

linuxwolf:
PR #17 chosen instead

PR #12, opened 2015-07-06
Gitignore pyc and other temp files

SamWhited:
Ensures that binary files generated by Python, and other temp files that we may end up with in the repo are ignored by Git.

Comprises two gitignore files taken from [this collection](https://github.com/github/gitignore).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,71 @@
+# Python
+# https://github.com/github/gitignore/blob/master/Python.gitignore
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+*.egg-info/
+.installed.cfg
+*.egg
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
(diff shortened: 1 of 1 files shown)


bear:
I'm +1 for this change as a short term fix. The longer term fix will be to move to using xsf-tools repo and remove all tooling from this repo.

small incremental changes FTW

SamWhited:
> The longer term fix will be to move to using xsf-tools repo and remove all tooling from this repo.

I'm torn about this statement: On the one hand, the tooling should almost certainly be developed in its own repo, on the other hand no one wants to clone a repo just to find that they have to clone another repo just to build what's in the first repo. Building an XEP should be as easy as cloning, and running some script, no other steps, or READMEs, or things to search for required.

I suppose that's a discussion for somewhere else though.

dwd:
Shub-modules may help you gain separation without making it too hard to
build. The problem is we don't want the same folks committing to the tools
and the XEPs.

But make can do the shub-module init, etc, and keep everything transparent.

bear:
My goal with that comment was to separate the xep content from the tools to generate the presentation content.

This will allow people who are read-only to have a single goto spot for content while people who need to edit/modify the content will have the tools needed. The edit/modify group is always going to be so much smaller than the consumer group that having them go thru a small "here is your tools repo" install step is not a concern IMO.

dwd:
Actually, I think the content group should be satisfied by the XEPs
website, not this repo at all. But in any case, I agree that any friction
should be placed in the edit line, but we can reduce that friction to very
low levels. We can probably also push rendered XEPs to another repo for
mirroring purposes.

bear:
I was reminded that, for me, editors included authors -- I tend to conflate them but I will try to be more precise.

I point this out because that matches what @dwd just said - author/editor work with the xep repo and consumers work with the generated content which is pushed to the website or other read-only location.

> Actually, I think the content group should be satisfied by the XEPs
> website, not this repo at all. But in any case, I agree that any friction
> should be placed in the edit line, but we can reduce that friction to very
> low levels. We can probably also push rendered XEPs to another repo for
> mirroring purposes.

Issue #10, opened 2015-07-02
gen.py fails to run


linuxwolf:
The XEP rendering script, gen.py, cannot run with the new repo layout.

The errors are because the intermediate XML file is empty.  This appears to happen because gen.py (and its deps) assume the parent directory is the repository, rather than the current directory.

winfried:
On 02-07-15 22:08, Matthew A. Miller wrote:

Hi,

> The XEP rendering script, gen.py, cannot run with the new repo layout.

Sorry I don't have time to dive into it right now, but this problem
should already be solved in: https://github.com/xsf/xsf-tools

Winfried

bear:
@winfried is there a place where we can talk about xsf-tools -- i'm digging into gen.py and I see a lot of small changes that I am just itching to fix/update/change so I'm thinking that some of them have already been thought about.

linuxwolf:
You're welcome to discuss "Editor" (not "editor" (-:  ) tooling in the editor@ MUC room, if you'd like.

PR #13, opened 2015-07-07
add a local helper routine to mimic what getLatestXEPFilename does for a non-mercurial working dir

bear:
work-around for issue #10 

Current gen.py assumes that the parent dir contains a mercurial repo and then walks the revisions to find the latest content for a given filename - this is then copied into a temp directory for the buildXEP() code to process.

This routine reads the contents for the given filename relative to the current working directory because it assumes that it's being called after getLatestXEPFilename() has failed.

This is a very quick work around to the issue, the real fix is to continue the work others have done to get xsf-tools working.

### DIFF
--- a/gen.py
+++ b/gen.py
@@ -323,9 +323,24 @@ def buildPDF( file, nr ):
 		
 	return True
 
+def getTempContents(filename):
+	try:
+		with open(filename, 'r') as h:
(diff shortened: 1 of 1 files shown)


linuxwolf:
Close in favor of #14.

PR #14, opened 2015-07-09
temporary fix to realize the repo root is now the current directory

### LINKED ISSUE #10: gen.py fails to run

linuxwolf:
Should fix #10.


linuxwolf:
Seems to work on xmpp.org.

PR #19, opened 2015-07-15
Exclude note on when last call ends if no <lastcall> 

linuxwolf:
Valuable for when a XEP is currently under consideration by XMPP Council, especially if one or more Council members voted -1 until author addresses issues.


stpeter:
+1 that's a fine little improvement. :-) Merging.

PR #22, opened 2015-07-18
Remove message sentinel from MAM (XEP-0313)

Flowdalic:
The MAM history already says "Switch the sentinel message back to the iq
result.", but this has never happened. This commit does the switch.

This hasn't been done in 6 months. I really like to proceed MAM forwards, and remove the sentinel, [which I consider to be a design flaw](http://mail.jabber.org/pipermail/standards/2015-January/029514.html)

The commit itself is best viewed as word diff: `git show f46e5d --word-diff`.

Summoning @mwild1 and @Kev for review.


linuxwolf:
Kev indicted in council@ (and editor@) MUC he is fine with these changes.

PR #23, opened 2015-07-19
Submit first draft of Mobile Considerations

SamWhited:
This is my initial draft of a new XEP: Mobile Considerations.

It is an informational XEP which supersedes [XEP-0286: XMPP on Mobile Devices](https://xmpp.org/extensions/xep-0286.html) and deals primarily with modern LTE networks.


SamWhited:
Will do; thanks. I want to actually put a number on that too; I've got a few other experiments I'd like to run if I can find the time (and come up with the devices / a proper methodology for testing it), but wanted to go ahead and get this out there and get some feedback.

dwd:
Sam, you're welcome to just edit XEP-0286 directly and get rid of any content that's not worth keeping. Saves creating a new XEP, and then having an older and a newer XEP that overlap.

SamWhited:
@dwd I started to do it that way at first, but I had ended up rewriting a lot of it anyhow, or cutting out most of the data (since it was taken with such an old phone), so I ended up just rewriting and marking this one as superceding the old one. I'm happy to go either way though; I debated keeping the 3G information, since there's still a very wide 3G deployment, so using the old one would make more sense if that were the case. I ended up not keeping it because most networks that I looked up in the US and Europe either already have LTE networks, or were in the process of upgrading (and the 3G state machine is far more complicated than the LTE one).

dwd:
I think that the vast majority of both handsets and coverage is still 3G, actually, but even if that's the case, I think you should just overwrite 286.

fippo:
gentleman, can you move this discussion to standards@ pretty please? Might get you a wider audience and I like the archives more.

linuxwolf:
Much of this discussion really does belong on standards@xmpp.org; I hope someone at least summarizes there sometime very soon.

In the meantime, there is still a protoXEP pending.  I do not see any problems with finishing the submission process on it; the XMPP Council can decide to formally object to publication as a separate XEP.  @SamWhited do you still want this to go through the submission process, or would you rather work with @dwd on updating XEP-0286?

linuxwolf:
merged as bddd8a4177e55ba7f663b6ce113d630b8bad17bc.

PR #29, opened 2015-07-24
Mobile considerations

SamWhited:
Update the numbers in the compression section of the mobile considerations document to include real world numbers (thanks to my boss who let me release them) instead of some perfect-world tests that I did. Will try to add more later.

Also add a few more recommended XEPs and fix some typos.

### DIFF
--- a/xep-0286.xml
+++ b/xep-0286.xml
@@ -183,7 +183,9 @@
 		to those developing mobile clients, or considering support for them in
 		servers.
 	</p>
-	<p>&xep0138; provides stream level compression.</p>
+	<p>&xep0138; provides stream level compression with the ZLIB codec.</p>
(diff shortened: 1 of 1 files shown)


xnyhps:
I suggest clarifying that where the text mentions flushes it means *full* flushes. zlib has sync flushes and full flushes. Sync flushes are already required to send stanzas (otherwise the data may end in the middle of a byte), but it's full flushes that are required to avoid inter-stanza correlations, which leak data.

SamWhited:
@xnyhps good call; thanks. Updated.

linuxwolf:
Merged in 751ef3ed25f700a558e396d8135fd948fd23359c, adding revision 0.3.

PR #32, opened 2015-07-27
Minor clarifications to XEP-0198

dwd:
Two things came up during community review of a patch to Openfire which are
not specified in the XEP. Both seem to be uncontentious, but will require
Council review, etc.

1) The specification makes no recommendation on what to do if the server
receives multiple <enable/> requests.

This is resolved in this instance by recommending (SHOULD, not MUST) that the
server generates a stream error, and stipulates that clients MUST NOT do this.

2) It was noted that redelivery attempts for stanzas require delay stamp
handling.

It is arguably not within this specification's remit to require this, however
it seems compelling that it be documented somewhere and this seems the best
place.

### DIFF
--- a/xep-0198.xml
+++ b/xep-0198.xml
@@ -29,6 +29,17 @@
   &dcridland;
   &mwild;
   <revision>
+    <version>1.4</version>
+    <date>2015-07-27</date>
+    <initials>dc</initials>
+    <remark>
+      <ul>
+        <li>Expressed how to handle duplicate enable requests.</li>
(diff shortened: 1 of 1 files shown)


linuxwolf:
Merged as 7dc7ba4b66a34eab500e5b15b944dbe9967aafe6; revision changed to 1.4rc1 and marked as interim.
Revision 1.4 pending vote by XMPP Council.

Rendered: http://xmpp.org/extensions/tmp/xep-0198-1.4.html
XEP Diff: http://xmpp.org/extensions/diff/api/xep/0198/diff/1.3/vs/1.4rc1

PR #39, opened 2015-08-02
Add <uri /> to XEP-0358

legastero:
Allows including a link to the advertised content or alternatives. For example, an advertisement for a  RTP video stream could include a link to a web page with a video player for the stream.

@stpeter

### DIFF
--- a/xep-0358.xml
+++ b/xep-0358.xml
@@ -63,6 +63,7 @@
   <jinglepub xmlns='urn:xmpp:jinglepub:1'
              from='sender-jid'
              id='someid'>
+    [<uri/> element for an alternate or informational URI of the content]
     [<meta/> element(s) for human-friendly information]
     [<description/> element(s) for application format(s)]
   </jinglepub>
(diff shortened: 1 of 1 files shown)


linuxwolf:
Merged as b6199650d63a1d5341dab50fee80f50d175588c3.

SamWhited:
/me twitches at the history that's being destroyed by all these commits. I really don't understand why we're losing it just to chagne the commit message. A normal merge would work just fine.

linuxwolf:
This is the model the editor team agreed to.

You are of course open to filing an complaint with the XSF Board.

SamWhited:
Being discussed in the XSF room now. This is nuts, you could merge, you could rebase and do a fast-forward merge, you could cherry-pick and push, you could ask them to squash and then merge that if they've got too many commits (or you could squash and maintain their authorship)... Git provides a lot of valid workflows, copy/pasting files around (or ammending the authors commits and destroying the hihstory) and re-pushing is not one of them. You're losing a lot of history here, which is going to make it difficult to find things in the future when you eg. want to know who made that one patch to that one XEP that one time, and instead all you can see is that @linuxwolf authored the entire file.

SamWhited:
Oops, yup, my mistake. The GitHub UI wes misleading; after observing the history, it looks like you're doing exactly what I was thinking you should be doing. Apologies.

PR #40, opened 2015-08-03
XEP-0313: MAM for MUC clarifications

mremond:
- Add a mention about MUC archives for non-anonymous rooms.
- Add lightweight initial information about checking credentials when querying MUC rooms.


intosi:
I believe you forgot the xmlns on the <message/> stanzas.

mremond:
@intosi good catch. I just fixed the example. Thanks !

zinid:
> I think this is slightly misled. The "Occupant JID" in XEP-0045 is room@service/nick, so the nickname is already accessible.
Indeed. And what about real JID included within 'http://jabber.org/protocol/muc#user' namespace in the message? Is it acceptable? I understand any server may include any tag they want, but it would be great if a client with sufficient priveleges has standard access to real JIDs. The namespace however could be 'http://jabber.org/protocol/muc#admin'.

Kev:
Yes, I think allowing the server (MAY) to insert the real JID in the existing xep-0045 format of
```
<x xmlns='http://jabber.org/protocol/muc#user'>
    <item affiliation='none'
          jid='hag66@shakespeare.lit/pda'
          role='participant'/>
```
Would be quite sensible.

zinid:
'affiliation' and 'role' requires a server to store additional metadata with every archived message. I'm not sure we need to store this and if this makes any practical sense. Also, is it acceptable if a server does not include such attributes?

Kev:
It seems like it might be useful information, but I agree that it's not going to usually be required. I think that making those two optional would be sensible (indeed, they could be made optional even when the user requesting the archive isn't allowed the full JID).

zinid:
Fine by me :)

zinid:
This is not the first time we receive "bug reports" like this one: https://github.com/processone/ejabberd/issues/703#issuecomment-135702806
Please update the XEP (MUC examples) ASAP.

linuxwolf:
This PR is still waiting on updates from the submitter.  Please see comments from Kev.

xep-0313.xml
```diff
@@ -16,6 +16,7 @@
   <dependencies>
     <spec>XMPP Core</spec>
     <spec>XEP-0030</spec>
+    <spec>XEP-0045</spec>
```
Flowdalic: Not sure if we should list XEP-45 as dependency. If at all, it's an optional one. But then you had to list XEP-60 too.

mremond: Actually, I added this dependency because I added a reference to a namespace coming from XEP-0045. However, I agree that it is not a "functional" dependency. I will remove the dependency in the PR, as it seems the way to go.

mremond:
I updated the PR based on @Kev and @Flowdalic.
I may need to think about what is implied by @zinid and @Kev discussion.

mremond:
I think I now processed all the feedback on the PR. Please, let me know if you see anything else, have more comments.
Thanks !

linuxwolf:
Will reach out to authors to review the latest changes.

linuxwolf:
Authors informed < http://mail.jabber.org/pipermail/editor/2015-October/000417.html >

Kev:
I'm clearly not good at GitHub pull requests, but this looks good to me, thanks Mickaël (and sorry for the delay).

linuxwolf:
Merged with revision 0.5 in f22a0a6

PR #41, opened 2015-08-06
XEP-0033: Multicast service MUST track directed presence broadcast

mremond:
This is to ensure entity will not stay permanently seen as online after a multicasted directed presence.

### DIFF
--- a/xep-0033.xml
+++ b/xep-0033.xml
@@ -27,6 +27,13 @@
         &stpeter;
 
         <revision>
+            <version>1.2</version>
+            <date>2015-09-22</date>
+            <initials>mre</initials>
+            <remark>Clarification: Multicast service MUST track directed presence broadcast to later ensure broadcast of subsequent unvailable presence to the same receiving parties.</remark>
+        </revision>
+
+        <revision>
             <version>1.1</version>
             <date>2004-09-15</date>
             <initials>psa</initials>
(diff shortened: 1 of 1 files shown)


Flowdalic:
I would probably state the "must track which entities have received presence" rule first, since it is required for the "must send unavailable to all entities which have received a available presence before" rule. I'd also begin with the "unavailable presence" in this rule. Also I'm not sure if we need to mention "directed presence" at all, because it appears they mutlicast service doesn't need to determine if it's an directed presence or not. What matters is that it's an "available" presence.

So I would probably write it like this in just one paragraph:

> The multicast service MUST keep track of the entities it send a presence with an absent 'type' attribute  (i.e. an 'available' presence) to and the originating entity of that presence. If the multicast service receives an unavailable presence from one of the tracked originating entities, it MUST broadcast this presence to all tracked entities which have not yet received an unavailable presence from the originating entity. Thus the multicast service ensures that all entities, which have received an (directed) 'available' presence over it, receive the related 'unavailable' presence.

But since I'm not a native speaker, I'm sure someone else will find a better wording. :)

I also think your work on this deserves your name being mentioned in the 'Acknowledgements' section. Not sure if you, some other XSF body (e.g. the editors) or the XEP authors are supposed to add that.

mremond:
Thanks @Flowdalic, this is good suggestion. I will update the patch accordingly.
I do not know about the contribution rules, so I will let official XSF editor apply the proper attribution rules of the XSF :)

mremond:
@Flowdalic I have just updated the PR, Thanks :)

linuxwolf:
Before it can be accepted, the Authors should be notified, and the XMPP Council will need to vote on it.

stpeter:
Native speaker and co-author here. :-)

I'd suggest:

This specification can be used to, in effect, send directed presence (see Section 4.6 of &rfc6121;). In order to ensure that entities that have received directed available presence through the service also are informed when the originating entity sends unavailable presence, the multicast service MUST do the following:

<ol>
<li>Keep track of the entities to which it sends available presence (i.e., presence stanzas with no 'type' attribute) along with the originating entity of that presence.</li>
<li>Upon receiving a presence stanza of type "unavailable" from an originating entity,  broadcast that unavailable presence to all entities to which it has send available presence.</li>

In this way, the multicast service ensures that all entities which have received available presence through the service also receive the associated unavailable presence.

linuxwolf:
This PR is still waiting on an update from the submitter.  Please see comments from stpeter.

mremond:
Thanks for the reminder, I will update the PR ASAP.

stpeter:
BTW, the sense of the Editor Team is that this would result in version 1.2 given that it modifies normative language. See http://logs.xmpp.org/editor/2015-09-22/#15:09:20

mremond:
Finally, updated the PR. I am very sorry about the delay. I think @stpeter wording is great. As per Peter suggestion, I attempted to prepare remark for point release changelog.

Please, let me know if there is anything else I can do to help.

stpeter:
Thanks, Mickaël! I think we're ready to bring this to the Council.

legastero:
This was approved by Council at the Oct 27 meeting (and subsequent list votes).

stpeter:
So think this can be published, no?

stpeter:
Bump. :-) Will add to Trello...

mremond:
Thanks !

SamWhited:
Merged as d63befb8a73fb6e75a7c6cd885907f8d27d89014, 65c9161227e64cae7e3624c940b1779f9d314f40, df0f08bf1dc5925b624e7176e514269ff104b9b3, ea1f090ed3707793f8a9b0d1ff26de4205190a05, be6ba8a5ae0dae0bdba7574bb48e289c15ac6137 (forgot to squash, sorry).

PR #42, opened 2015-08-08
Fix typos in XEPs 0060 and 0247

linkmauve:


linuxwolf:
Merged along with 73a62049e0b7a241ba876573b92f071f7065e641.

Combined with one other typo that prevented xep-0247 from rendering.

PR #44, opened 2015-08-11
XEP-0352: Explain how CSI is affected by in order processing

Flowdalic:
Following up on the productive discussion caused by #34, I now suggest the following change instead of #34. Hope that everyone agrees and is happy with out. If not, please raise your voice. :)

Clarify how the in order processing requirement (RFC 6120 § 10.1) affects CSI.

### DIFF
--- a/xep-0352.xml
+++ b/xep-0352.xml
@@ -116,6 +116,28 @@
   <p>This protocol is intended primarily for clients with human interaction. Due to the open-ended nature of
   the possible optimisations implemented by the server, it may not be suitable for non-IM purposes where the
   fully standard behaviour of XMPP is required.</p>
+  <section2 topic="In-order processing" anchor="in-order-processing">
+	<p>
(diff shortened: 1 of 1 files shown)


fippo:
with a LC requested for CSI I'd prefer not to have any changes made to that until the LC is done. However, a branch tracking LC feedback and proposed changes would be good :+1:

stpeter:
I will raise this issue on the list.

Flowdalic:
> I will raise this issue on the list.

I couldn't find it raised on the list, so I commented on the LC thread.

mwild1:
Just +1'd this in the LC thread on the list.

linuxwolf:
Merged via PR #105.

We're trying to work out a flow to prevent problems with the published database.  This was the first experiment.

linuxwolf:
Re-opening ... the experiment was a failure.  XEP-diff would not recognize the actual changes.

linuxwolf:
Merged in with #107 and updated revision.

PR #47, opened 2015-08-12
Advance XEPs -0293 and -0294 to Draft.

linuxwolf:
Final Council tallies completed in 2015-05, but advancement didn't actually happen!

### DIFF
--- a/xep-0293.xml
+++ b/xep-0293.xml
@@ -32,7 +32,7 @@
   </author>
   <discuss>jingle</discuss>
   <revision>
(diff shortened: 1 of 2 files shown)


stpeter:
Presumably we also need to add the schema files to the schema/ directory so the schemaloc links work...

linuxwolf:
That work (as well as updating the registrar/namespaces.xml) has already been done.

stpeter:
Well then, +1.

PR #50, opened 2015-08-18
Carbons

Kev:
This reflects discussion on list and in the XSF MUCs. There are two subsequent patches from Georg, but after discussion last night didn't lead to a conclusion about whether I can submit those directly I'm not doing so.

I'm assuming that the XEP authors will naturally be alerted as part of this submission :)


linuxwolf:
Merged as 756e38b9c2c9b7eff8edde6bfa3da789e2dd7d99.

PR #51, opened 2015-08-18
Carbons: Remove full-JID/bare-JID distinction

ge0rg:
[this is based on #50]

After some more discussion in xsf@ we figured out that at least two server implementations (prosody, ejabberd) haven't implemented section 6 of XEP-0280 anyway, and were using identical behavior for full-JID and bare-JID messages. Furthermore, section 6 is completely redundant, as the same behavior can be achieved under existing RFC 6121 §8.5.2.1.1 rules.

OTOH, message forking (as opposed to carbon-copying) introduces several "problems":

 * a carbons-enabled client with a negative priority can receive chat
   messages under the section 6 rules.
 * a carbons-enabled client can not determine if it received a bare-JID
   message due to regular message routing or due to Carbons (this is
   useful for determining if a notification should be silent or loud,
   though a "smart" client can determine that by watching presence of
   the user's other resources).
 * it makes the XEP more complicated for no benefit.

Therefore, I have gone a step further than Kev and removed section 6 (and rephrased section 7 accordingly). As this change does not break anything, I'd like to have it added to XEP-0280. However, it is based on Kev's patches, so please discuss it now, and I'll open a PR as soon as Kev's changes are (hopefully) accepted.


linuxwolf:
Merged as 756e38b9c2c9b7eff8edde6bfa3da789e2dd7d99.

PR #59, opened 2015-08-21
Initial X2X over S5066

stevekille:
This is a ProtoXEP for operating server to server over HF Radio, by mapping XEP-0361 onto STANAG 5066.


linuxwolf:
Merged as 3ec99b2c0491616e08e0fcef76597f1ce22d0457.

PR #60, opened 2015-08-21
XEP-0280: Fix bare-JID example to match RFC 6121

xnyhps:
[RFC 6121 § 8.5.2.1.1](https://tools.ietf.org/html/rfc6121#section-8.5.2.1.1) states:

> In all cases, the server MUST NOT rewrite the 'to' attribute (i.e.,
> it MUST leave it as `<localpart@domainpart>` rather than change it to
> `<localpart@domainpart/resourcepart>`).

### DIFF
--- a/xep-0280.xml
+++ b/xep-0280.xml
@@ -244,7 +244,7 @@
   <example caption='Message Forked to each of Romeo&apos;s Carbons-enabled resources'><![CDATA[
 <message xmlns='jabber:client'
          from='juliet@capulet.example/balcony'
-         to='romeo@montague.example/garden'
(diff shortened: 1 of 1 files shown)


xnyhps:
Alternatively, if the Carbons-forking is intentionally different from RFC 6121 forking, then I suggest the text in 0280 mentions that (and preferably also the reason why it is different).

linuxwolf:
These changes are no longer relevant.  Please review against 0.10, and resubmit if you believe they still apply.

linuxwolf:
This change is no longer relevant.

PR #63, opened 2015-08-25
XEP-0215: Try resolving some ambiguities

legastero:
- Adds attributes indicating that credentials are required (if they are not already provided)
- Adds attribute indicating when credentials expire
- Clarifies which attributes are needed to identify a service when requesting credentials

### DIFF
--- a/xep-0215.xml
+++ b/xep-0215.xml
@@ -123,6 +123,16 @@
       <th>Inclusion</th>
     </tr>
     <tr>
+      <td>action</td>
+      <td>When sending a push update, the action value indicates if the service is being added or deleted from the set of known services (or simply being modified). The defined values are "add", "remove", and "modify", where "add" is the default.</td>
+      <td>OPTIONAL</td>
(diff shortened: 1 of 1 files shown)


SamWhited:
"Try resolving some ambiguities" is itself a bit ambiguous. Could you please provide a description which summarizes what was changed / fixed with this PR? Just a short list of bullets should suffice; nothing onerous.

Thanks!

linuxwolf:
A note has been sent to the authors asking about these changes.

stpeter:
With my author hat on, I'd say this looks good. IMHO adding the 'action' attribute necessitates incrementing of the protocol version. Also these changes are substantive enough that I think it would be appropriate to add Lance as a co-author.

stpeter:
@legastero Would you mind making the changes outlined above? (Bump the protocol version and add yourself as an author.)

linuxwolf:
Merged as 11cba41.

PR #66, opened 2015-08-26
Fix otr-info construction section

SamWhited:
Don't work around clients that don't behave properly and stick extra XML in the <body> node.

Also remove a confusing opinion from the security section.

### DIFF
--- a/inbox/otr-info.xml
+++ b/inbox/otr-info.xml
@@ -147,14 +147,10 @@
 <section1 topic='OTR Messages'>
 	<section2 topic='Construction and Decoding'>
 		<p>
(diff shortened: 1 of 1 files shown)


linuxwolf:
Following on council@ MUC room, it looks like more changes are coming.

Do you want to delay publishing until those updates are ready?  If the changes would be ready in the next few days, then we can wait.  Otherwise, we'll be happy to revision when they are.

linuxwolf:
Merged in 36bc44599796124358c9580c966af44afe2bc606.

NOTE: This is now XEP-0364!

PR #67, opened 2015-08-26
Entity Versioning

SamWhited:
>  A method by which rosters and disco items may be versioned so that servers
> will not need to send the entire list if it has not been modified, saving
> bandwidth and time during session initialization with minimal state being 
> stored by the server and client.

Rendered version: http://hipchat.github.io/xeps/xep-entityversioning.html


linuxwolf:
IPR acknowledgement sent and received:

< http://mail.jabber.org/pipermail/editor/2015-August/000409.html >

linuxwolf:
Merged as b3725e2cf9070cd750bd025bea46da13e433a403.

PR #68, opened 2015-08-26
Update XEP-0264 to allow non-cid URIs, and expand scope

legastero:
- Changes name to "Jingle Content Thumbnails", as these can be used outside of file transfer
- Updates format of `<thumbnail />` to use a 'uri' attribute to allow HTTP URIs and not just BoB
- Updates examples to latest protocol formats
- Bumps namespace version


ashward:
I've sent Marcus a message as he is the original author.

mlundblad:
These changes seems reasonable to me.

PR #70, opened 2015-08-29
Initial draft of OTR disco

SamWhited:
An XEP to facilitate the discovery of Off-the-Record messaging support without using OTR's protocol agnostic mechanism (which does not allow querying).

### DIFF
--- a/inbox/otrdisco.xml
+++ b/inbox/otrdisco.xml
@@ -61,17 +61,23 @@
 		<p>
 			If an entity supports OTR it MUST advertise the fact by returning a
 			feature of 'urn:xmpp:otr:0' &VNOTE; in response to a &xep0030;
-			information request. It MUST also include '&lt;version&gt;' elements for
-			each version of the OTR protocol that is supported. Eg. an entity which
-			supports OTR versions 2 and 3 might include the following feature in its
-			disco info list:
-
+			information request. This indicates support for OTRv3 as defined by
+			<i><link url='https://otr.cypherpunks.ca/Protocol-v3-4.0.0.html'>
+				Off-the-Record Messaging Protocol version 3
+			</link></i>
+			<note>
+				(Accessed 2015-08-30). "Off-the-Record Messaging Protocol version 3"
+				&lt;<link url='https://otr.cypherpunks.ca/Protocol-v3-4.0.0.html'>
+					https://otr.cypherpunks.ca/Protocol-v3-4.0.0.html
+				</link>&gt;
+			</note>.
 			<example caption='Disco response'><![CDATA[
-<feature var='urn:xmpp:otr:0'>
-	<version>2</version>
-	<version>3</version>
-</feature>
+<feature var='urn:xmpp:otr:0' />
 			]]></example>
+		</p>
+		<p>
+			If older versions of OTR are required, they may be discovered out of band
+			using OTRs built in mechanism which is beyond the scope of this document.
 		</p>
 	</section1>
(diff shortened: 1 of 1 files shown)


inbox/otrdisco.xml
```diff
@@ -0,0 +1,107 @@
...
+			client for support and to know in advance that a particular version of
+			OTR is supported. This XEP provides a mechanism for discovering OTR
+			support at the XMPP level. This specification aims to solve that by
+			providing an in-band mechanism for discovering OTR support in XMPP.
+		</p>
+		<p>
+			It should be noted that newer, more secure encryption protocols exist for
+			XMPP, and that new implementations of OTR are discouraged. This protocol
+			is primarily intended to solve issues with existing implementations of
+			OTR.
```
jcbrand: Hi, I'm curious about the newer, more secure encryption protocols you mention. Which are these? Would it perhaps make sense to mention them in this spec as well?

inbox/otrdisco.xml
```diff
@@ -0,0 +1,107 @@
...
+			If an entity supports OTR it MUST advertise the fact by returning a
+			feature of 'urn:xmpp:otr:0' &VNOTE; in response to a &xep0030;
+			information request. It MUST also include '&lt;version&gt;' elements for
+			each version of the OTR protocol that is supported. Eg. an entity which
+			supports OTR versions 2 and 3 might include the following feature in its
+			disco info list:
+
+			<example caption='Disco response'><![CDATA[
+<feature var='urn:xmpp:otr:0'>
+	<version>2</version>
```
Flowdalic: `<feature>` can't have child elements as per its schema. It would also break entity caps. We could either restrict OTR to version 3 or use something like `urn:xmpp:otr:0:version:X`, where X equals the OTR version, instead.

inbox/otrdisco.xml
SamWhited: I'll add a ref once the XEP is published, but I was specifically thinking of the GSoC axolotl implementation. 

http://conversationsgsoc2015.blogspot.com/

inbox/otrdisco.xml
SamWhited: Interesting; does 0030 actually say that anywhere? I couldn't find a place that said it, so maybe that could use some clarification.

I originally used namespacint, but that felt wrong since someone coild then advertise 0:version:1 and 1:version:2 which just felt weird... I'll switch it back though.

Flowdalic: > Interesting; does 0030 actually say that anywhere?

In the schema (xep30 § 11). But I would not be opposed if xep30 would state it in human readable text too. Maybe some editor can take care of this?

Another approach would be using `#` as separator, e.g.. `urn:xmpp:otr:0#otr-version-2`.

SamWhited:
After thinking about it some more, I think @Flowdalic's suggestion of only supporting OTRv3 makes the most sense. Most clients use v3 anyways [citation needed], and if you need something older you can still fall back to OTRs built in mechanisms.

If this spec is updated in the future (eg. for OTRv4) a version bump will be necessary.

linuxwolf:
IPR acknowledgement sent: < http://mail.jabber.org/pipermail/editor/2015-August/000410.html >

linuxwolf:
IPR acknowledgement received: < http://mail.jabber.org/pipermail/editor/2015-August/000411.html >

linuxwolf:
Merged as b867a4d5682f9c3bb17a7a0fd20092e3f11d4cc3.

PR #71, opened 2015-08-30
XEP-0030: Clarify that feature elements may not have children

SamWhited:
This change to XEP-0030 clarifies that `<feature/>` elements cannot have children. Currently the text does not mention this, but the schema enforces it.

This change is purely editorial and is not a protocol change.


Flowdalic:
Not sure, but what about s/must not/MUST NOT/?

SamWhited:
Yup, you're completely right. Fixed.

stpeter:
+1 as author - this needs Council sign-off given that XEP-0030 is in the Final state, and will lead to version 2.5.

linuxwolf:
Merged as 48f8846cee76ce76bf84d95bc55523433a660aec.

Revision is 2.5rc1 until Council approval.

PR #72, opened 2015-08-30
XEP-0364: Discourage new implementations of OTR

SamWhited:
Discourage new OTR implementations in XEP-0364 as discussed in the editors MUC.

This is not a technical or protocol change, just a clarification to make an existing statement stand out more.

### DIFF
--- a/xep-0364.xml
+++ b/xep-0364.xml
@@ -74,8 +74,13 @@
 		</note>
 		and will not be redescribed here. Instead, this document aims to describe
 		OTR's usage and best practices within XMPP. It is not intended to be a
-		current standard, or technical specification, as better (albeit, newer and
-		less well tested) methods of end-to-end encryption exist for XMPP.
+		current standard, or technical specification.
+	</p>
+	<p>
+		It should be noted that newer, more secure encryption protocols exist for
+		XMPP, and that new implementations of OTR are discouraged. This document is
+		primarily intended to document and resolve issues with existing
+		implementations of OTR.
 	</p>
 </section1>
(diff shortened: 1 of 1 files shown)


rtreffer:
I think it's bad style to say "hey, this is old and bad, use one of the newer fancy alternatives!" without at least mentioning an alternative....

Unrelated to this particular PR, but might be worth adding.

SamWhited:
Yah, agreed; as soon as there's an actual XEP published for OMEMO (axolotl) this should be updated.

stpeter:
Given that OTR isn't an XSF technology, it seems strange for the XSF to discourage implementations.

SamWhited:
OTR isn't an XSF technology, but it's being used in such a way that it's tightly coupled with XSF technologies. It's this use which should be discouraged, IMO.

rtreffer:
The problem I'm seeing is that it would in effect discourage end-to-end encryption - mainly because we'd discourage the use of the only(?) widespread implementation.
It might also contradict the post-NSA mantra of encrypting as much as possible....

(To clarify: I'd vote for killing OTR, but I'm unwilling to pay the price of dropping end-to-end encryption)

I'd really prefer a neutral statement _or_ a pointer to an alternative :-S

linuxwolf:
This discussion is going beyond editorial, which means it needs to happen on standards@xmpp.org.

SamWhited:
This all makes good sense. I'll withdraw this PR temporarily (untill the OMEMO XEP is published) and we can revisit then and discuss more broadly if the XSF wants to recommend one encryption mechanism or another.

PR #75, opened 2015-09-01
Add makefile

SamWhited:
Add a Makefile for building XEPs so that one doesn't have to go fetch another repo and figure out how to get the Python scripts working.

The PDF generation is a bit hairy (and requires that texml be installed), but otherwise it's simple enough. Just run `make` (or `make help`) for info.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,32 @@
+.SILENT:
+
+OUTDIR?=build
+XMLDEPS=xep.xsl xep.xsd xep.ent xep.dtd ref.xsl
+
+.PHONY: help
+help:
+	@echo 'XEP makefile targets:'
+	@echo ' '
+	@echo '                  help  -  (this message)'
+	@echo '                   all  -  build all XEPs'
+	@echo '                 clean  -  recursively unlink the build tree'
+	@echo '              xep-xxxx  -  build xep-xxxx.html'
+	@echo ' '
+	@echo 'Output directory: "$(OUTDIR)/"'
+
+.PHONY: all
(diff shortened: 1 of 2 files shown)


linuxwolf:
I am always getting an error building the pdf variants, but I haven't tried to dive through it yet to see why:

sed: 1: "tmp/xepbuild/xep-0001.x ...": undefined label 'mp/xepbuild/xep-0001.xml.texml.tex'
make: *** [tmp/xepbuild/xep-0001.xml.texml.tex] Error 1

linuxwolf:
It would be nice to get some feedback in the terminal when each XEP (html, pdf) finishes building.

SamWhited:
> I am always getting an error building the pdf variants, but I haven't tried to dive through it yet to see why:

Not sure what that's about; it was working for me on one machine, but I've only ever gotten the PDF generation stuff to work on that one machine. The texml stuff is incredibly irritating to get working.

> It would be nice to get some feedback in the terminal when each XEP (html, pdf) finishes building.

Done, though I suspect this will be annoying when generating the entire tree.

SamWhited:
Actually, I suspect this is a difference between the GNU version of sed and whatever you're running. I'm going to take a wild guess and say you're probably on OS X?

Flowdalic:
Works for me. Can't wait till this gets merged. Thanks @SamWhited 

>> It would be nice to get some feedback in the terminal when each XEP (html, pdf) finishes building.

> Done, though I suspect this will be annoying when generating the entire tree.

I also would prefer not having `echo` in the Makefile. One could use `--trace` if you really need to know what make is doing. But that's not a blocker for me.

linuxwolf:
Merged in 931daef

PR #78, opened 2015-09-07
ProtoXEP entityversioning: Add profiles / partial sync

SamWhited:
Adds the concept of "profiles" to entity versioning (which allows the XEP to define EV in terms of the roster, but allows future XEPs to extend it to version any type of list).

Also mention that the server should be able to send down only part of a list, and the client should be able to query for changes to only part of a list.

Still missing the "auto complete" endpoint which completes the picture and allows partial sync to really work. This will come as soon as may be, but need not be part of this PR.

### DIFF
--- a/inbox/entityversioning.xml
+++ b/inbox/entityversioning.xml
@@ -8,10 +8,9 @@
 <header>
   <title>Entity Versioning</title>
   <abstract>
-    A method by which rosters and disco items may be versioned so that servers
-    will not need to send the entire list if it has not been modified, saving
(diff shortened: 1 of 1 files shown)


linuxwolf:
Marking as Needs Council, as it is currently in review.  Once the council finishing weighing in one way or the other, we'll get this PR merged.

linuxwolf:
Merged as 61dd466.

Added revision 0.0.2 -- "Add profiles / partial sync"

PR #81, opened 2015-09-11
XEP-0359: Refactored SID's client-id from attribute to element.

Flowdalic:


linuxwolf:
The shortname "sid" conflicts with:

* Stream ID in XEP-0095
* Session ID in XEP-0124

The Editor prefers "stanza-id".

linuxwolf:
Merged in 611eb46.

PR #82, opened 2015-09-11
XEP-0313: Update MAM to use stanza-id from XEP-359

Flowdalic:


Flowdalic:
Would be great if someone could double-check the values of the stanza-id's 'by' attribute.

zinid:
I don't understand the purpose of moving MAM id from <result/> to <stanza-id/>. Any rationale? How do those IDs from <stanza-id/> interact with MAM IDs?
BTW, you also need to update the namespace from 1 to 2.

linuxwolf:
Authors notified of proposed changes: < http://mail.jabber.org/pipermail/editor/2015-September/000415.html >

mwild1:
Sorry, I wasn't actually notified about this (that's now resolved).

I'm -1 on merging this right now, though it is a good start. I will provide feedback or patches soon, however I think we should wait until the new carbons reflection is out, as we will probably need to reference that. And yes, we would need to bump the namespace.

@zinid We're moving to <stanza-id/> for consistency, so that the stanza id can be used separately from MAM. It will become the standard way for the server to assign a permanent id to a stanza it receives from the client.

SamWhited:
Closing this for now since I'm working on cleaning up the backlog and @mwild1 is -1 on it. If feedback is provided or addressed, feel free to re-open and we'll re-contact the authors. Thanks!

iNPUTmice:
Including the MAM id (possibly in form of the stanza-id XEP) into live messages is important to reliably fetch the backlog of messages after being offline for a while.
I understand that it is not a perfect solution because you don't get the ID of sent messages but as one says perfect is the enemy of good or what ever. And at least I don't see a downside of injecting the archive id in form of stanza-ids into live incoming messages. Clients can still choose not to use it or use a different mechanism in the future. It's only 'additional' information and doesn't take anything away.

On the server side this is minimal effort (ejabberd already does it and a patch for prosody would have 5 lines of code or what ever)

Note that this would also require an addition to the stanza-id XEP saying that every entity that adds and cleans malicious stanza ids (stanza ids with the same by attribute as the entity that where not added by that entity) should announce the stanza id namespace in their disco thus allowing a client to know if the stanza-id in live incoming messages is in fact reliable and not faked by another entity.

iNPUTmice:
I create a PR #270 to address the required changes in the stanza-id XEP.

PR #83, opened 2015-09-13
XEP-0280: Carbons processing hints update

SamWhited:
Remove the `<private/>` element from [XEP-0280][xep0280] in favor of the `<no-copy/>` element defined in [XEP-0334][xep0334].

I wasn't sure if this would require a version bump, so I put that in a second commit. If it's decided that no version bump is needed we can drop that commit from the PR.

[xep0280]: https://xmpp.org/extensions/xep-0280.html
[xep0334]: https://xmpp.org/extensions/xep-0334.html

### DIFF
--- a/xep-0280.xml
+++ b/xep-0280.xml
@@ -41,6 +41,14 @@
     <jid>linuxwolf@outer-planes.net</jid>
   </author>
   <revision>
+    <version>0.11</version>
+    <date>2015-09-13</date>
+    <initials>ssw</initials>
+    <remark>
+      <p>Remove &lt;private&gt; in favor of XEP-0334's &lt;no-copy&gt;.</p>
+    </remark>
+  </revision>
+  <revision>
     <version>0.10</version>
     <date>2015-08-24</date>
     <initials>mm (editor)</initials>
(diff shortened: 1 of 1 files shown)


linuxwolf:
This suggestion has come up in the past, and not deemed desirable because there were already implementations.

However, we can bring it back up on the list, and see what implementations say now.

SamWhited:
Yah, I thought it was discussed again recently with the LC, but I couldn't remember what the outcome was (or find it in my archive) so I just made the change. I might just tweak it to reference the message processing hints and say that both should be supported as someone suggested since that's what most implementations appear to be doing now anyways.

Flowdalic:
> and say that both should be supported as someone suggested since that's what most implementations appear to be doing now anyways.

I suggested to change carbons so that it says that implementations must understand both flavors of the hints (legacy carbons and xep334) but use only xep334 hints for outgoing stanzas. So support both when receiving stanzas, but use only xep334 hints when sending. This way it is a non-breaking change, and we will potentially, at one day far far in the future, be able to remove the legacy carbons hints one and for all.

As far as I can tell I'm not the only one who like to see carbons to use xep334 hints, and I think that's a feasible way which all parties could agree on.

linuxwolf:
Locking this thread.  This discussion MUST happen on the mailing list, not the issue.

SamWhited:
We could equally have had a conversation via private message; this wasn't any sort of official communication, so I don't see any reason to take the exterme step of locking the thread so that no one else can leave comments. Anything important should be discussed on list first, but if you refuse to allow any discussion to happen anywhere else you're going to have to police peoples XMPP accounts and you're going to spend half your time leaving comments like that. People will talk wherever they happen to be, there's nothing that can be done about that.

SamWhited:
After list discussion, it sounds like this is not the correct behavior. Will create a new PR when I figure out what changes need to be made.

PR #85, opened 2015-09-13
XEP-0065: Fix description of <activate/> element

Flowdalic:


Flowdalic:
Follow up on http://mail.jabber.org/pipermail/standards/2015-March/029663.html

xep-0065.xml
```diff
@@ -776,7 +786,7 @@ DATA = (payload)
     <p>The "jid" attribute specifies the JID of the StreamHost. This attribute MUST be present, and MUST be a valid JID for communication over XMPP.</p>
   </section2>
   <section2 topic='&lt;activate/&gt; Element' anchor='desc-activate'>
-    <p>The &lt;activate/&gt; element is sent from the Requester to the Proxy in order to formally start the bytestream. This element is always empty and has no defined attributes.</p>
+    <p>The &lt;activate/&gt; element is sent from the Requester to the Proxy in order to formally start the bytestream. This element has no defined attributes and its XML text contains always the JID of the target.</p>
```
stpeter: I'd change "XML text" to "XML character data".

xep-0065.xml
```diff
@@ -30,6 +30,16 @@
...
+    <version>next</version>
+    <date>201?-??-??</date>
+    <initials>fs</initials>
+    <remark>
+      <ul>
+        <li>Fixed description of activate element in 9.4.</li>
+      </ul>
+    </remark>
+  </revision>
+  <revision>
```
stpeter: I don't think this requires a new revision - the correction is editorial.

xep-0065.xml
Flowdalic: Isn't that potentially causing confusion with CDATA as it's the first google result when searching for "XML character data? I'm aware the existing XEPs use this term to refer to the text wrapped around an element, but I'm not sure if it's really the best term.

stpeter: That's the term from the XML specification and we use it everywhere. If it's confusing, let's complain to the W3C rather than make up our own terminology. :-)

Flowdalic: Alright, you convinced me. :)
.

xep-0065.xml
stpeter: Actually, per discussion in the Editor Team, we decided to create "patch-point releases" for fixes like this: http://logs.xmpp.org/editor/2015-09-15/#15:52:08

xep-0065.xml
stpeter: In fact I'd say "its XML character data specifies the JID of the target"

Flowdalic:
Pushed new version of the commit.

linuxwolf:
Merged as fb6a277.

Updated revision to 1.8.1 to publish 2015-09-17.

PR #91, opened 2015-09-22
XEP-0233: Include support for Microsoft SSPI in xep-0233

mvisode:


linuxwolf:
(as co-author) these changes look fine to me.

(as editor) this update will be a new minor revision.  @mvisode is there a full name we can credit with the changes, or better yet make a co-author? (-:

linuxwolf:
Merged as bee05d6.

mvisode:
Saw the comments too late. But I see you have put my full name now.. :) Don't mind being a co-author - but maybe leave that for the next update now.

linuxwolf:
I'll keep that in mind.  Don't be surprised if you see an editorial update that adds you! (-:

Thanks again!

PR #94, opened 2015-09-26
XEP-0050: element should be attribute

sco0ter:


stpeter:
Looks like a straightforward editorial fix to me.

linuxwolf:
Merged as 4988e27.

PR #96, opened 2015-09-26
XEP-0171: Schema wrongly specifies elements/attributes as required

sco0ter:
E.g. Example 1 and 2 have no <source/> element
Many examples have no dictionary attribute.

### DIFF
--- a/xep-0171.xml
+++ b/xep-0171.xml
@@ -416,7 +416,7 @@
 
   <xs:element name='x'>
     <xs:complexType>
-      <xs:element ref='source' use='required'/>
+      <xs:element ref='source' use='optional'/>
       <xs:sequence>
         <xs:element ref='translation' use='required' minOccurs='1'/>
       </xs:sequence>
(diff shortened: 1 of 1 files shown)


xep-0171.xml
```diff
@@ -440,7 +440,7 @@
           <xs:attribute name='charset' type='xs:string' use='optional'/>
           <xs:attribute name='source_lang' type='xs:language' use='optional' />
           <xs:attribute name='destination_lang' type='xs:string' use='required'/>
```
stpeter: Actually the datatype for destination_lang needs to be corrected to 'xs:language', too.

sco0ter:
I've corrected the destination_lang to be xs:language

stpeter:
Great, thanks! Just something I noticed when reviewing your PR. IMHO this is good to go now, @linuxwolf and I will take it from here. :-)

linuxwolf:
Merged as 3020652.

PR #99, opened 2015-09-28
Add `make preview' to generate HTML on XML save

SamWhited:
Uses inotifywatch to generate an HTML file anytime an XML file is saved.

This probably won't work on OS X, but it would be easy enough to use `fswatch` if someone that actually has a machine running it wants to tweak it to fallback to fsswatch if inotifywait doesn't exist.


linuxwolf:
This looks interesting, but on hold until one of the editors can try it out for themselves.

SamWhited:
Added support for files in the inbox and rebased to remove a bad commit that I apparently merged from another branch at some point.

PR #100, opened 2015-09-28
XEP-0138: Update security recommendations

SamWhited:
This change to `XEP-0138: Stream Compression` does not invalidate existing implementations (although I was very tempted to make it REQUIRED), but merely states that implementations SHOULD perform a full flush on stanza boundaries to prevent CRIME and BREACH like attacks.

This is fairly common knowledge, but I thought the XEP should really encourage it.

### DIFF
--- a/xep-0138.xml
+++ b/xep-0138.xml
@@ -31,6 +31,12 @@
     &hildjj;
     &stpeter;
     <revision>
+      <version>2.1</version>
+      <date>2015-09-28</date>
+      <initials>ssw</initials>
+      <remark><p>Specify that implementations should perform a full flush on stanza boundaries.</p></remark>
+    </revision>
+    <revision>
       <version>2.0</version>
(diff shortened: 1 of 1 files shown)


stpeter:
Although this is a fine recommendation, I think we need to revisit the security considerations of this spec in a bit more detail, reference http://breachattack.com/ and perhaps RFC 7525 and RFC 7457, etc.

SamWhited:
I definitely agree, though I think this PR can be merged to make sure the spec addresses the common fix right now, and then we could go back and flush it out more later (I'd like to do this, but won't have the time until after summit sometime most likely).

fippo:
i remember raising some concerns on standards@ -- will look them up tomorrow. Shall we add this to council agenda?

SamWhited:
Closing after the 2015-12-09 council meeting in which we decided to find someone to do a more comprehensive review of the security of 0138.

PR #102, opened 2015-09-28
XEP-0320: clarify that holdconn is not used

fippo:
based on LC feedback from @stpeter this clarifies that holdconn is not used/defined.

Also adds reference to 0343

### DIFF
--- a/xep-0320.xml
+++ b/xep-0320.xml
@@ -303,7 +303,12 @@ a=setup:role
                 <xs:enumeration value='active'/>
                 <xs:enumeration value='passive'/>
                 <xs:enumeration value='actpass'/>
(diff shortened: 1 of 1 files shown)


xep-0320.xml
```diff
@@ -292,7 +303,7 @@ a=setup:role
                 <xs:enumeration value='active'/>
                 <xs:enumeration value='passive'/>
                 <xs:enumeration value='actpass'/>
-                <xs:enumeration value='holdconn'/>
+                <xs:enumeration value='holdconn'/><!-- not used -->
```
stpeter: It's probably more schema-ish to put this commented text into an xs:annotation element.

PR #104, opened 2015-09-29
XEP-0016: Deprecate privacy lists

SamWhited:
Instead, use the blocking command ([XEP-0191](https://xmpp.org/extensions/xep-0191.html)) and the invisibility
command ([XEP-0186]((https://xmpp.org/extensions/xep-0186.html)))

This has been discussed on the list several times and was brought up recently in another thread, but I will go ahead and create a new thread for this PR and then get it on the councils agenda for a vote (assuming the discussion is leaning in this direction).

*EDIT:* Raised on the mailing list: http://mail.jabber.org/pipermail/standards/2015-September/030388.html


SamWhited:
Closing this until the next time discussion happens. Will re-open then.

PR #106, opened 2015-10-02
XEP-0060: be more consistent with reply

sstrigler:
It's not just more consistent but also easier to implement like this in most cases probably ;)

### DIFF
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -1499,12 +1499,19 @@ And by opposing end them?
       ]]></example>
     </section3>
     <section3 topic='Success Case' anchor='subscriber-unsubscribe-success'>
-      <p>If the request can be successfully processed, the service MUST return an IQ result.</p>
+      <p>If the request can be successfully processed, the service MUST return an IQ result and SHOULD include a &PUBSUB; child element with the updated &lt;subscription/&gt; element for that node.</p>
       <example caption='Service replies with success'><![CDATA[
 <iq type='result'
     from='pubsub.shakespeare.lit'
     to='francisco@denmark.lit/barracks'
-    id='unsub1'/>
+    id='unsub1'>
+  <pubsub xmlns='http://jabber.org/protocol/pubsub'>
+    <subscription
+        node='princely_musings'
+        jid='francisco@denmark.lit'
+        subscription='none'/>
+  </pubsub>
+</iq>
       ]]></example>
     </section3>
     <section3 topic='Error Cases' anchor='subscriber-unsubscribe-error'>
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -1499,7 +1499,7 @@ And by opposing end them?
       ]]></example>
     </section3>
     <section3 topic='Success Case' anchor='subscriber-unsubscribe-success'>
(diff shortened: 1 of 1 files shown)


linuxwolf:
This change will need some list discussion, should get buy-off from authors, and definitely needs council to weigh in.

sstrigler:
Yeah, not so sure about that myself anymore. Maybe rather have a MAY there instead of a SHOULD?

linuxwolf:
Bring it to the list! (-:

SamWhited:
Do we know if this was ever discussed on list? Does anyone have a link to the discussion if so?

sstrigler:
> On 20 mars 2016, at 16:12, Sam Whited <notifications@github.com> wrote:
> 
> Do we know if this was ever discussed on list? Does anyone have a link to the discussion if so?
> 
> 

Probably not, probably I didn't know immediately how to address this and than simply forgot about it.

I raised this when I was implementing it and found that the implementation had to be odd at that point. Which let me to believe there is an inconsistency in the protocol as well.

sstrigler:
Oh well, and there it is: http://mail.jabber.org/pipermail/standards/2015-October/030422.html

stpeter:
Now that I think about it more, this behavior can be helpful - most especially if a node allows multiple subscriptions with different 'subid' values (Section 6.1.6). So mentioning that and adding a 'subid' attribute to the example seems like a good idea. I'd be +1 at that point (kind of +0 at the moment). Thanks @sstrigler for the PR!

ralphm:
Agreed with @stpeter.

SamWhited:
Ping @sstrigler I'm going through the backlog and cleaning up a bit, any interest in resuming this work with @stpeter's blessing, or should I close this? Thanks!

sstrigler:
Ah, wasn't sure how to read stpeter's response. So should I modify the PR accordingly? Will do.

sstrigler:
Also rebased on latest from upstream master and removed accidental whitespace removals from second commit.

SamWhited:
@sstrigler Many thanks! Let me know when this is in a state that you like and I'll solicit feedback from the authors again and then try to get this merged. Sorry for dropping the ball and not getting to it earlier.

sstrigler:
I think it's ready to go 😹

SamWhited:
Author comms sent. I'll get this in front of the council as soon as they both approve. Thanks!

ralphm:
These changes look good to me. :shipit:

SamWhited:
Council pinged; hopefully this will be on the agenda for next week.

sstrigler:
Coincidentally I went over the whole XEP today and found several places where it still refers to Jabber instead of XMPP. Is this by intention or should that be fixed?

SamWhited:
> Coincidentally I went over the whole XEP today and found several places where it still refers to Jabber instead of XMPP. Is this by intention or should that be fixed?

I think that would be a worthwhile fix; feel free to lump it in with these changes or submit a second PR if you prefer.

ralphm:
@sstrigler so I see. Please resolve those in a different PR, except for *Jabber ID*.

sstrigler:
I'll do another one. But not before Monday.

SamWhited:
> I'll do another one. But not before Monday.

No rush, this is editorial so we won't need to pester the council for another review before merging, I'll just wait and lump them both into a single revision. Thanks again!

stpeter:
LGTM!

PR #107, opened 2015-10-02
XEP-0352 v0.2: Explain how CSI is affected by in order processing

linuxwolf:


linuxwolf:
Merged in PR #44 with revision change.

PR #110, opened 2015-10-16
ProtoXEP: Message Attaching

SamWhited:
This proposal adds a new XEP for "attaching" one message to another. Attachments might include information such as a preview of a link in a previous message that was generated by the server, or a "sticker" or "emotion" that a user wants displayed next to a previous message in the conversation (even though that message was not the most recent message sent).

For example, HipChat currently renders messages such as the following image link (and will soon use this spec to do so:

tempest

Rendered version: https://blog.samwhited.com/xeps/xep-attachto.html


linuxwolf:
Merged to 974e0fc.

PR #112, opened 2015-10-25
protocol to negotiate omemo encrypted file tranfers over jingle

iNPUTmice:
Jingle File Transfer (XEP-0234) describes a very flexible and powerful method for peer-to-peer file transfer with interchangable transports. Unfortunatly only some of those transports can by encrypted (e.g. Jingle In-Band Bytestreams Transport Method (XEP-0261)) and none integrate into existing end-to-end encryption schemes. This specification defines an approach to encrypt the actual file before transferring it by using the OMEMO encryption.


linuxwolf:
Pending IPR response from author(s) < http://mail.jabber.org/pipermail/editor/2015-October/000419.html >.

Merged as 1b133e6.

PR #116, opened 2015-10-31
ProtoXEP: SRV records for XMPP over TLS

moparisthebest:
This proposal defines a procedure to look up _tls SRV records in addition to _tcp and mix weights/priorities.

Rendered version can be found at https://burtrum.org/xeps/tls-srv.html

I do have a POC implementation written up for Conversations [here](https://github.com/siacs/Conversations/pull/1371).

### DIFF
--- a/inbox/tls-srv.xml
+++ b/inbox/tls-srv.xml
@@ -1,0 +1,80 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE xep SYSTEM 'xep.dtd' [
+  <!ENTITY % ents SYSTEM 'xep.ent'>
+%ents;
+<!ENTITY rfc7301 "<span class='ref'><link url='https://tools.ietf.org/html/rfc7301'>RFC 7301</link></span> <note>RFC 7301: Transport Layer Security (TLS) Application-Layer Protocol Negotiation Extension &lt;<link url='https://tools.ietf.org/html/rfc7301'>https://tools.ietf.org/html/rfc7301</link>&gt;.</note>" >
+]>
+<?xml-stylesheet type='text/xsl' href='xep.xsl'?>
+<xep>
+<header>
+  <title>SRV records for XMPP over TLS</title>
+  <abstract>This specification defines a procedure to look up _tls SRV records in addition to _tcp and mix weights/priorities.</abstract>
+  <legal>
+    <copyright>This XMPP Extension Protocol is copyright (c) 1999 - 2014 by the XMPP Standards Foundation (XSF).</copyright>
+    <permissions>Permission is hereby granted, free of charge, to any person obtaining a copy of this specification (the &quot;Specification&quot;), to make use of the Specification without restriction, including without limitation the rights to implement the Specification in a software program, deploy the Specification in a network service, and copy, modify, merge, publish, translate, distribute, sublicense, or sell copies of the Specification, and to permit persons to whom the Specification is furnished to do so, subject to the condition that the foregoing copyright notice and this permission notice shall be included in all copies or substantial portions of the Specification. Unless separate permission is granted, modified works that are redistributed shall not contain misleading information regarding the authors, title, number, or publisher of the Specification, and shall not claim endorsement of the modified works by the authors, any organization or project to which the authors belong, or the XMPP Standards Foundation.</permissions>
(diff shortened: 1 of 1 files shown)


fippo:
I like the idea of going directly for TLS, this saves two RTTs. 

But the only time I have ever seen _tls used was by Microsoft's OCS/Lync. I do not think this is a good idea, even though from reading RFC 2782 this might be allowed and I can't find a registry of allowed proto values at the IANA.

In terms of approach IMO the better way is to just attempt TLS on the same port as _xmpp-client. Demuxing through. jabberd14 implemented demuxing of TLS and TCP back in 2004 already so that is doable, even though I am not sure how many servers implement this.
In case that does not work the client should attempt a reconnect using starttls.

```
as xmpp-over-tls is indistinguishable from http-over-tls
```
not when ALPN is used though? But it will make it indistinguishable for old gateways not implementing ALPN.
The ALPNs should be registered in a separate spec, either XEP or RFC. @stpeter might have an opinion on what is less hassle.

nit: startls or just with tls directly -> STARTTLS or just with TLS directly

moparisthebest:
I do see that _tls for proto is used for Microsoft Lync, and I'm almost positive I read that _sip._tls was also used for SIP, but I can't seem to find that reference now.  I did stumble upon [this thread](https://support.counterpath.com/topic/sip-tls-and-dns-srv-records-on-bria-iphone) that mentioned at least some SIP software does this:

    Looking the the code. We do "_sips._tcp", not "_sip._tls"

So I also wouldn't be opposed to doing "_xmpps-client._tcp" instead of "_xmpp-client._tls", but reading [RFC 2782](https://www.ietf.org/rfc/rfc2782.txt) which says:

```
   Proto
        The symbolic name of the desired protocol, with an underscore
        (_) prepended to prevent collisions with DNS labels that occur
        in nature.  _TCP and _UDP are at present the most useful values
        for this field, though any name defined by Assigned Numbers or
        locally may be used (as for Service).  The Proto is case
        insensitive.
```

It seems like either would be just as valid.

I'm not sure about just trying TLS first, it's doable with something like [sslh](https://github.com/yrutschle/sslh) but it's not as performant or scalable perhaps as just using separate ports, and I like that SRV records let the server operator define exactly how they prefer everything to line up.

About ALPN and such, yes a bad middle proxy could detect that ALPN is being used currently, but that's why I marked it SHOULD instead of MUST, a client could have a checkbox to disable ALPN, and the server owner could still know where to send xmpp.example.org traffic vs say www.example.org traffic via SNI (in fact that's how my server works, I wrote a patch for the above mentioned sslh to do so).

dwd:
I wasn't sure, either, about the use of _tls; so I asked Arnt Gulbrandsen (author of RFC 2782) - his opinion was that using _tls, and intermixing the _tcp proto records, was perfectly acceptable. His concerns were actually that he felt that ALPN should be made mandatory.

Given this opinion, I'm fine with this specification, though it's not clear to me if this should be a XEP or an RFC.

My own opinion is that the specification seems OK, though I suspect there'll be some details that need careful thought.

moparisthebest:
Great to know the author of RFC 2782 thinks this mixing seems acceptable, thanks!

As to ALPN as mandatory, I'm kind of torn, on one hand it'd be great for a server operator to know they can always multi-plex ports using ALPN.  On the other hand it *can* decrease privacy, and further the current state of ALPN support in various devices/languages/libraries is fairly poor, though that should be less of an issue in the future.

SamWhited:
> His concerns were actually that he felt that ALPN should be made mandatory.

Agreed. Can you bring this up on list (if it hasn't been already... I've been kind of out of the loop).

moparisthebest:
I fixed @fippo 's nit, thanks!

I also brought this up on the standards@xmpp.org mailing list here:
http://mail.jabber.org/pipermail/standards/2015-November/030560.html

linuxwolf:
Accepting XEPs will be paused temporarily until the XMPP Council and Board elections complete, which is about another week.

doctorlard:
Forgive my stupids, but is it intended that the _tcp and _tls SRV records are on different ports, or could I use 5222 (and 5269) for both (i.e. prioritising direct TLS over STARTTLS on the same port)?

linuxwolf:
IPR questionnaire sent < http://mail.jabber.org/pipermail/editor/2015-December/000428.html >

linuxwolf:
Merged as b298964

PR #119, opened 2015-11-09
Updated contact information and example JIDs

PeterWaher:
New contact information has been provided, and example JIDs are updated
to point to example.org.

### DIFF
--- a/xep-0332.xml
+++ b/xep-0332.xml
@@ -27,28 +27,14 @@
         <supersedes/>
         <supersededby/>
         <shortname>NOT_YET_ASSIGNED</shortname>
-        <author>
-            <firstname>Peter</firstname>
-            <surname>Waher</surname>
-            <email>peterwaher@hotmail.com</email>
-            <jid>peter.waher@jabber.org</jid>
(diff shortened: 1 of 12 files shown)


xep-0332.xml
```diff
@@ -30,11 +30,28 @@
         <author>
             <firstname>Peter</firstname>
             <surname>Waher</surname>
-            <email>peter.waher@clayster.com</email>
+            <email>peterwaher@hotmail.com</email>
             <jid>peter.waher@jabber.org</jid>
             <uri>http://www.linkedin.com/in/peterwaher</uri>
         </author>
 		<revision>
+			<version>0.5</version>
```
linuxwolf: These two new revisions should be collapsed into a single revision (0.4).

linuxwolf:
Given how many documents you have, I think it would be best if you used an entity reference for your author info.  See &stpeter; or &linuxwolf; in xep.ent (and various XEPs) for examples.

xep-0332.xml
PeterWaher: Removed the duplicate revision statement.

linuxwolf:
Merged collapsed as b01bdca.

PR #120, opened 2015-11-09
xep-0339 version 0.2

fippo:
remove label and mslabel from obsolete version of msid-draft. fix whitespace in code tags

cleaning up things and making them ready for LC. cc @stpeter


linuxwolf:
Merged as 92c060f.

PR #126, opened 2015-11-17
XEP-0071: Update plain version of blockquote example to escape <div/>

xnyhps:
I'm assuming `<div/>` isn't actually a valid child element for the plain `<body/>`, but that this should have been XML escaped just like it is in the xhtml version.

### DIFF
--- a/xep-0071.xml
+++ b/xep-0071.xml
@@ -699,7 +699,7 @@ You wrote:
 
    I think we have consensus on the following:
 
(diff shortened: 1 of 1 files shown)


ashward:
Emailed the author to request approval for change

stpeter:
+1

linuxwolf:
Merged as 120926b.

PR #150, opened 2016-02-10
XEP-0045: Don't leak JIDs in anonymous rooms

weiss:
Don't allow occupants of members-only rooms to retrieve the member list unless the room is non-anonymous.

### DIFF
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -3332,7 +3332,7 @@
   </query>
 </iq>
     ]]></example>
-    <p>Note: A service SHOULD also return the member list to any occupant in a members-only room; i.e., it SHOULD NOT generate a &forbidden; error when a member in the room requests the member list. This functionality can assist clients in showing all the existing members even if some of them are not in the room, e.g. to help a member determine if another user should be invited. A service SHOULD also allow any member to retrieve the member list even if not yet an occupant.</p>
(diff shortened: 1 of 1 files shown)


weiss:
Apparently, members are [permitted][1] to retrieve the members list also in anonymous rooms.  In that case, the note that I was updating seems redundant to me: "A service SHOULD also return the member list to any occupant in a members-only room [&hellip;]."  Anyway, my change would not be consistent with [table 6][1], so I'm closing this PR.

[1]: http://xmpp.org/extensions/xep-0045.html#affil-priv

xnyhps:
Why wouldn't it be possible to update table 6 to say "Yes, if the room is non-anonymous" too?

I think this is a security issue with the XEP: people who register their nick with a room with open registration will likely not assume that they reveal their real JID to all other members, so in my opinion it's fine to update a specification that is rather old and already widely implemented.

An alternative would be to allow members in semi-anonymous open rooms to retrieve the member list, but to only include the registered nicks for each member, not their real JID. But this is a larger change so has more potential to break something:

```xml
<iq from='coven@chat.shakespeare.lit'
    id='member3'
    to='crone1@shakespeare.lit/desktop'
    type='result'>
  <query xmlns='http://jabber.org/protocol/muc#admin'>
    <item affiliation='member'
          nick='thirdwitch'
          role='participant'/>
  </query>
</iq>
```

Issue #154, opened 2016-02-28
XEP-0234: Checksums are being sent using session-info, not transport-info messages


piotrjurkiewicz:
Now XEP says the following:

>Note: Computing the hash of the file before sending it can slow down the process of file transfer, because the sending application needs to process the file twice. The File Sender might prefer to send the hash after the file transfer has begun, using a **transport-info** message as described under Checksum.

Should be:

>Note: Computing the hash of the file before sending it can slow down the process of file transfer, because the sending application needs to process the file twice. The File Sender might prefer to send the hash after the file transfer has begun, using a **session-info** message as described under Checksum.

legastero:
Thanks! I've submitted a PR to fix the issue.

PR #155, opened 2016-02-28
XEP-0234: Specify session-info instead of transport-info

legastero:
Fixes https://github.com/xsf/xeps/issues/154


linuxwolf:
Mreged with 40123bf.

PR #159, opened 2016-03-12
Update references to DNA

SamWhited:
The DNA framework is now out of draft (see the [datatracker](https://datatracker.ietf.org/doc)). This updates the only reference to point to RFC 7712 and remove the old entity.

```sh
xeps $ grep -IR "dna-framework"
xeps $
```

A sanity check would be nice before I merge.


stpeter:
We also have an entity for "xmppdna" so let's remove that while we're at it.

We could also change "dna-dns" to RFC 7673, I suppose. (Although it's not cited anywhere.)

SamWhited:
Good eye, will do.

PR #160, opened 2016-03-12
XEP-0191: Allow blocking arbitrary JIDs

SamWhited:
Clarify that arbitrary JIDs (not just contacts) may be blocked with the blocking command. This matches real world usage of the spec.

I've done a minor version bump because this is backwards compatible, but does (possibly) introduce new functionality (even if most [all?] servers are already doing this).


stpeter:
+1 - yes this is the intended use.

PR #164, opened 2016-03-18
XEP-0370: Mark as experimental

SamWhited:
XEP-0370 is experimental, not a protoxep (it's already been accepted by the council and assigned a number)


SamWhited:
IPR: http://mail.jabber.org/pipermail/editor/2016-March/000467.html

PR #178, opened 2016-05-15
xep-0234: update wrong reference to ice-tcp

fippo:
ICE-TCP is RFC 6544. RFC 6455 is an important too and authored by someone familiar but not related to filetransfers ;-)


legastero:
+1. Merge it

PR #179, opened 2016-05-16
Fix the updating an XEP section of the README

SamWhited:
Update to match the current process. Submitting a PR before I merge to get @ashward's approval in case I screwed anything up.

Other sections are still wrong; I just updated the section for the thing I just did as I'm reasonably sure the steps are correct. Other sections can be updated later as we do them.


SamWhited:
Going ahead and merging. Please update and correct if you get to it. Thanks.

PR #182, opened 2016-05-17
XEP-0166: Fix draft-ietf-stox-media ref

SamWhited:
Fixes a broken reference.


SamWhited:
Change of plans, doesn't require author. I'm making the call that this is editorial and merging.

PR #187, opened 2016-05-21
Reduce legal notice in template to XML entity

SamWhited:
@ashward @stpeter Any reason we wouldn't want to do this in the template?


SamWhited:
Merging; let me know if this was a bad decision.

PR #188, opened 2016-05-21
Spam reporting

SamWhited:
Add the spam reporting XEP that was created at the XMPP Summit 20. Needs to be squashed before being merged.


SamWhited:
Initial work squashed and merged as 342d2ac56a02af67ba6c13a606d9fbd39085e8b2

PR #189, opened 2016-05-24
Add 'jingle' XMPP URI querytype to XEP-0358

Flowdalic:
In order to allow users to share files via XMPP URIs, Jingle and XEP-0358. But this is not only limited to file sharing, e.g. links Audio/Video conferences could be shared too.

If this get's the accepted, then the registry should be updated to. But http://xmpp.org/registrar/querytypes.html is currently a 404 and I've no idea where the registry is maintained.


legastero:
Author +1, merge it

PR #194, opened 2016-06-04
XEP-0156: Fix typo (s/utf-9/utf-8/)

SamWhited:
Not sure what the procedure on editorial fixes to Draft XEPs is, but I'm pretty sure the encoding here should be UTF-8, not UTF-9.

### DIFF
--- a/xep-0156.xml
+++ b/xep-0156.xml
@@ -26,6 +26,12 @@
   &stpeter;
   &lance;
   <revision>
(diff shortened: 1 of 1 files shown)


ashward:
I think this could count as an editorial change, so I don't think needs to go through council.

ashward:
btw, the procedure is that any 'material' changes to Draft XEPs need council approval.

PR #198, opened 2016-06-07
XEP-0156: The XRD host-meta file is REQUIRED and the websocket subprotocol is now RFC 7395

SamWhited:
Add a business rule clarifying that the XML version of the host-meta file in XRD format is REQUIRED, while the JSON version is OPTIONAL.

This is already specified by the RFC, but for clarity it's worth stating here too.

Also update references to the draft version of the websockets subprotocol to point to RFC 7395.

This change is editorial in nature.

### DIFF
--- a/xep-0156.xml
+++ b/xep-0156.xml
@@ -26,6 +26,12 @@
   &stpeter;
   &lance;
   <revision>
+    <version>1.1.2</version>
+    <date>2016-06-07</date>
+    <initials>XSF Editor: ssw</initials>
+    <remark><p>Clarify that the XML version of the XRD document is required by the RFC and the JSON version is optional.</p></remark>
+  </revision>
+  <revision>
     <version>1.1.1</version>
     <date>2016-06-06</date>
     <initials>XSF Editor: ssw</initials>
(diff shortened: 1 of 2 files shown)


xep-0156.xml
```diff
@@ -168,6 +174,7 @@ _xmppconnect IN TXT "_xmpp-client-websocket=wss://web.example.com:443/ws"
       <li>HTTP queries for host-meta information MUST be used only as a fallback after the methods specified in <cite>RFC 6120</cite> have been exhausted.</li>
       <li>A domain SHOULD NOT present information in host-meta link records that is available via the DNS SRV records defined in <cite>RFC 6120</cite>.</li>
       <li>The order of XMPP related link entries in the host-meta file SHOULD NOT be interpreted as significant by the presenting domain or the receiving entity.</li>
+      <li>Support for the XML version of the host-meta file is REQUIRED as specified by <cite>RFC 6120</cite> &sect;3 while the JSON version (and other representations) are OPTIONAL.</li>
```
sonnyp: Minor but it sounds like it is the XEP that makes alternative representations optional.

I'd go with something like
```
As specified in RFC 6120, support for the XML/XRD encoding of the host-meta resource is REQUIRED while alternative representations such as JSON are OPTIONAL.
```

sonnyp:
For reference

> The host-meta document uses the XRD 1.0 document format as defined by [OASIS.XRD-1.0]

> The server MAY offer alternative representations of any XRD document it serves

> If no "Accept" request header field
   is included with the request, or if the client requests an
   "application/xrd+xml" representation, the server MUST respond using
   the REQUIRED XRD 1.0 XML representation described in Section 3.1.

https://tools.ietf.org/html/rfc6415#section-3

xep-0156.xml
SamWhited: Done. Moved it out of the business logic too since I'm not sure it made sense there the way I'd worded it.

sonnyp:
:+1:

PR #199, opened 2016-06-08
Fix some copypasta problems.

bjc:
Items returned from a pubsub items request are encapsulated in an
<items/> element, not <publish/>, which is used for publishing items.

Example for requesting the secret key was labelled as requesting the
public key.


SamWhited:
Merged. @bjc sorry for the delay, and thank you for your submission!

PR #202, opened 2016-07-08
XEP-0106: Update to use PRECIS and newer XMPP RFCs

SamWhited:
Updates XEP-0106 to be compatible with [PRECIS](https://tools.ietf.org/html/rfc7564) and [RFC 7622](https://tools.ietf.org/html/rfc7622).

Possible known issue that will need feedback: I think Unicode spaces in the Zs category were mapped to ASCII space in Nodeprep, but they won't be with the UsernameCaseMapped profile of PRECIS. Should we add an extra step for this as a concern of the XEP, or am I wrong about how Nodeprep behaves?


SamWhited:
Author emails sent.

stpeter:
Hey @SamWhited thanks for working on this. In NodePrep, both ASCII space and the various non-ASCII space characters were disallowed (see Appendix A.5 of RFC 3920, which specifies that the characters in C.1.1 and C.1.2 of RFC 3454 were prohibited output). We retained that functionality in RFC 6122 and RFC 7622.

While we're at it, we might as well update the email format reference from RFC 2822 to RFC 5322 (note that the section numbers changed!).

Everything else looks fine (those Wireless Village examples are awfully dated by now, but I see no big reason to change them).

SamWhited:
> In NodePrep, both ASCII space and the various non-ASCII space characters were disallowed (see Appendix A.5 of RFC 3920, which specifies that the characters in C.1.1 and C.1.2 of RFC 3454 were prohibited output). We retained that functionality in RFC 6122 and RFC 7622.

Ah, for some reason I thought I checked and noticed this was different; oops. Oh well, nothing to do here :)

> While we're at it, we might as well update the email format reference from RFC 2822 to RFC 5322 (note that the section numbers changed!).

Updated.

stpeter:
Hey @SamWhited one more slight fix to those sections of RFC 5322:

s/3.2.4/3.2.3/ (for the Atom section)
s/3.2.5/3.2.4/ (for the Quoted Strings section)

SamWhited:
@stpeter I fixed the atom section I missed (thanks!), but I don't see anymore references to 2.2.5 or that section name

stpeter:
👍

PR #206, opened 2016-07-12
XEP-0375: Refactor tables to be feature-focused

SamWhited:
Based on feedback from the mailing list (/cc @dwd) and @iNPUTmice.


SamWhited:
Update: Also added "Avatars" as an IM compliance suite item (listing XEP-0084 as the only provider since IQ and vCard based avatars should be deprecated)

PR #213, opened 2016-07-15
XEP-0375: Webclients don't need BOSH and websockets

SamWhited:
Adds a footnote to indicate that web clients don't need to implement BOSH and Websockets. It should be an OR.

### DIFF
--- a/xep-0375.xml
+++ b/xep-0375.xml
@@ -43,6 +43,14 @@
     &stpeter;
     &sam;
     <revision>
+      <version>0.3</version>
+      <date>2016-07-20</date>
+      <initials>ssw</initials>
+      <remark>
+        <p>Don't require both BOSH and Websockets.</p>
+      </remark>
+    </revision>
+    <revision>
       <version>0.2</version>
       <date>2016-07-11</date>
       <initials>ssw</initials>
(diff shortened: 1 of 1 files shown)


dwd:
I'm not sure this is right - unless Websockets have improved greatly recently, there are deployment considerations that would mean a webapp might be better supporting both for a while yet.

SamWhited:
Do any webapps support both? I assumed there was no reason to support both since it was just a connection mechanism and which they used was deployment specific (although you get something vaguely similar to 0198 with BOSH, so maybe it should be "Websockets and 0198").

sonnyp:
We supported both at redbooth.com

WebSocket is not available on IE < 10 and in some (enterprise) network configurations it's not usable.

SamWhited:
I'm still unsure about this commit now; I feel like individual deployments would have no need to support both (it would just be a matter of deployment policy which one the client supported if &lt;company-or-service-developing-the-client&gt; just used BOSH, they'd only support BOSH) and therefore we shouldn't require it, but general clients being deployed by many people should probably support both for the reasons others have outliend.

jcbrand:
+1 to what @SamWhited is saying, since BOSH alone is sufficient for all modern and not-so-modern browsers. I've personally worked on multiple deployments of XMPP web-clients that only support BOSH.

SamWhited:
Decided that I think this is definitely correct, so I'm merging. Further discussion can happen on list and I'll revert if we get this to a vote eventually and the council or community really disagrees with me.

xep-0375.xml
```diff
@@ -314,6 +322,10 @@
       &#x2021; Support for the Entity Use Cases and Occupant Use Cases is
       REQUIRED; support for the remaining use cases is RECOMMENDED.
     </p>
+    <p>
+      &sect; Only one of the recommended providers must be implemented for
+      compliance.
+    </p>
```
sonnyp: @SamWhited  s/providers/mechanisms ?

PR #214, opened 2016-07-15
Remove old submission mechanism

SamWhited:
There are probably other files in here that could go too.


Kev:
I'm not sure this is right, is it? People still have the option of mailing the Editors instead of using Github, if they want to.

ashward:
Totally. There are a number of contributors who don't have github accounts and are actively against using github. We therefore need to make sure there is a secondary method of contributing.

SamWhited:
This HTML page isn't being used anywhere though, is it? This information should live in the README for this repo, or on the new website.

Kev:
On 18 Jul 2016, at 14:25, Sam Whited <notifications@github.com> wrote:
> This HTML page isn't being used anywhere though, is it? This information should live in the README for this repo, or on the new website.

Ah, I’d misunderstood. Apologies.

/K

PR #218, opened 2016-07-27
XEP-0045: participant should be member

sco0ter:
I think this sentence is confusing.

A moderator cannot be a participant at the same time, because it can only have one role.
I think the context is referring to an affiliation here, which should be "member" or "member or admin".


tfar:
Looks like a sensible change.

SamWhited:
@sco0ter any chance you could add a new revision block with a minor version bump at the top? Thanks.

PR #219, opened 2016-07-29
XEP-0045: Business Rule for client-requested full resync

ge0rg:
As [discussed on the ML][0]: When a client re-joins a MUC where the MUC service did not notice the client leaving, the client will end up with an incomplete list of participants and no room subject.

This patch adds a sentence to the business rules section that explicitly requires a MUC service to send a full sync whenever it receives a MUC join presence (as opposed to the old groupchat protocol).

In my eyes, this is only a clarification of the existing standard definition, so we do not need to bump the version.

Georg

[0]: http://mail.jabber.org/pipermail/standards/2016-June/031180.html


tfar:
Seems on this is mostly a clarification of behaviour already described and implementers on the mailing list all say this reflects current implementation  behaviour. I don't find a mention in http://xmpp.org/extensions/xep-0001.html that this would require any action from the council. The change looks okay to me. Would be nice if someone with merge powers could merge this.

SamWhited:
@ge0rg Any chance you could add a revision block at the top with a minor version bump? Thanks.

Flowdalic:
Council trello card: https://trello.com/c/wBfhDIMV

stpeter:
Although this seems like a fine change, I'm still curious how this is currently handled in running code.

stpeter:
I'm +1 on this, not having seen any concerns from implementers.

Flowdalic:
Council voted all +1: https://trello.com/c/wBfhDIMV

PR #221, opened 2016-08-02
Fix typo in dependencies

Zash:
s/RCC/RFC/


SamWhited:
Thanks; I'm going to just merge this and it will get folded in when / if another iteration of this XEP is done or the next time all XEPs are generated (I don't think it's worth doing a version bump for a single typo fix).

PR #226, opened 2016-08-14
XEP-0143: Fix new XEP creation instructions

linkmauve:
The xeps repository has been split from the xmpp repository, this
reflects that change.


SamWhited:
@linkmauve Thanks for your submission; could you please add a revision block at the top? Thanks!

linkmauve:
Hi, I added another commit with the revision, I chose to only increase the patch level because it doesn’t even count as a minor revision imo.

PR #228, opened 2016-08-16
Mix

Kev:
Address summit discussions

### DIFF
--- a/xep-0369.xml
+++ b/xep-0369.xml
@@ -37,6 +37,12 @@
 
   &stpeter;
   <revision>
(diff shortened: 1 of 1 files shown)


Kev:
@stevekille please confirm "Send a reply to the author(s) in order to verify that, if the proposal is accepted, they will cede rights to the protocol in accordance with the XSF's IPR policy https://xmpp.org/about/xsf/ipr-policy."

stevekille:
I can conform compliance to IPR policy

PR #229, opened 2016-08-17
XEP-0369 misc editor fixes

linkmauve:
Sorry for having done a PR before the patch got merged, this is a rebase on top of current master.  Each commit is reviewable separately as a single unit.

### DIFF
--- a/xep-0369.xml
+++ b/xep-0369.xml
@@ -192,10 +192,10 @@
     <section3 topic="Subject Node" anchor="subject-node">
       <p>The subject node publishes the current subject of channel.  Subject history is stored in MAM.    The id of each node is the current value of the channel subject.   A single item is stored in this node at a time.  Users subscribe to this node to receive subject updates. The following example shows the format of a item in the Subject node.</p>
       <example caption="Subject Node"><![CDATA[
-    <items node='urn:xmpp:mix:nodes:subject'>
-      <item id='How to use Toads' />
-    </items>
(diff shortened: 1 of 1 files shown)


xep-0369.xml
```diff
@@ -637,9 +637,6 @@
     to='mix.shakespeare.example'
     from='hag66@shakespeare.example/pda'
     id='7nve413p'>
-  <register xmlns='urn:xmpp:mix:0'>
```
Kev: > rfc6120 §8.2.3.6: “An IQ stanza of type "result" MUST include zero or one child elements.”

Well, yes, but this isn't a result stanza, it's an error.

SamWhited: I feel that including the original payload is rather useless and we shouldn't encourage this behavior, so either way I think it's worth removing this.

stevekille:
I have been over all of these, and from my perspective these are good to roll in,   Steve

xep-0369.xml
linkmauve: I will change the commit message, but the change should indeed stay.

Also github still randomly fails to send email notifications… /rant

Kev: It's useful for eyeballing/debugging, but it's just a diagnostic thing, yes.

Kev:
Let's merge at the same time as Sam's once Sam's have been reviewed, and we can do one version update with both.

PR #232, opened 2016-08-19
XEP-0070: Mention a plaintext fallback

mathieui:
In the previous version, an entity requesting XEP-0070 validation could
add a `<body/>` element to the `<message/>` sent, which would be displayed
in any client (which may not support this XEP). If the client does not
support the XEP, and the entity does not support a plaintext fallback,
the client would be prompted to act, without any result being possible
(as the wording says that the `<confirm/>` element must be sent back as
well).

This change suggests that an entity adding a body should implement this
kind of fallback (which may be as dumb as "ok/nok", as long as the
`<body/>` element is clear enough).


guusdk:
+1

SamWhited:
I forgot to add this to the council agenda. Done now (https://trello.com/c/Cvtj9LTf/132-review-plaintext-fallback-xep-0070-pr); it should be discussed next week. Sorry for the delay.

stpeter:
I assume that the <body/> would provide instructions about what to send back to the server in order to complete the workflow? It might be good to mention that.

SamWhited:
@mathieui this is now ready for merge; sorry for the delay. What initials should I put in the revision notes for you?

SamWhited:
Merged as 50c04ec (will add authors initials if/when they reply)

mathieui:
@SamWhited Oops, sorry, I missed it. (I think I answered that in a previous PR, but probably on the website repo; "mp" is fine)

SamWhited:
Thanks!

PR #234, opened 2016-08-25
Version bump for Link Mauve Edits

stevekille:


xep-0369.xml
```diff
@@ -36,6 +36,13 @@
   </author>
 
   &stpeter;
+  
+  <revision>
+    <version>0.2.1</version>
+    <date>2016-08-25</date>
+    <initials>sek</initials>
```
Kev: These are Link Mauve's edits, rather than yours.

stevekille:
Fixed in update

PR #235, opened 2016-08-25
fix typo (joing -> join)

marktran:


SamWhited:
Thanks for the report; this was actually just fixed in some other editorial changes that I merged in and will be in the next published version.

PR #236, opened 2016-08-26
XEP-0369: Sort out the broken XML; Tidy Refs; Updates Acks

stevekille:


SamWhited:
Turns out that `<cite>` is HTML (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/cite). It is also referenced in `xep.dtd` and is used heavily throughout the XEP series, so I don't think we should stop using it; instead maybe we can make it link to the original citation somehow?

stevekille:
Please, please, please can we drop <cite> use in 369 and get this update (and other pending ones) pushed.

For the most part oXygen (a market leading XML editor) has been serving me very well in editing 369.   However, it chokes on <cite>

I don't have the skills to fix xep,dtd so that oXygen is happy.    

I don't have time or inclination to switch editors.

I am very keen to get on and address a stack of email comments and update 369.

intosi:
It shouldn't choke on `<cite/>` anymore.  It was missing as allowed child of a few elements, but that should've been fixed now with my DTD changes in #242.

Kev:
Abandoning after discussion with @stevekille out of band. A replacement PR will follow.

PR #237, opened 2016-08-26
XEP-0033: Clarify Reply Handling, add 'ofrom' type and fix examples

sco0ter:
- A stanza can contain multiple replyto and replyroom items (as per §
  4.6.4), make it clear in the reply handling section as well.
- Fix examples. Include the BCC address in the addresses element, when the
  stanza is sent to the BCC addressee.
  This is as per the sentence ' Each 'bcc' recipient MUST receive only the
  <address type='bcc'/> associated with that addressee.'
- Include 'ofrom' attribute as defined in XEP-0045.


Flowdalic:
@stpeter could you have a look and ack/nack this?

SamWhited:
Going to take this to the council since the author is busy.

SamWhited:
Sorry for the delay; will merge this soon.

PR #238, opened 2016-08-28
XEP-0369: Better i18n and Unicode safety

SamWhited:
* Allows for MUC rooms with titles in multiple languages. This can be useful eg. for a service that attempts to automatically translate rooms into the users preferred language and locale.
* Use the PRECIS nickname profile defined in RFC 7700 for nickname safety

/cc @stevekille @Kev @stpeter


xep-0369.xml
```diff
@@ -204,20 +204,20 @@
       <p>Private Messages are not stored in the messages node.</p>
     </section3>
     <section3 topic="Subject Node" anchor="subject-node">
-      <p>The subject node publishes the current subject of channel.  Subject history is stored in MAM.    The id of each node is the current value of the channel subject.   A single item is stored in this node at a time.  Users subscribe to this node to receive subject updates. The following example shows the format of a item in the Subject node.</p>
-      <example caption="Subject Node"><![CDATA[
-<items node='urn:xmpp:mix:nodes:subject'>
-  <item id='How to use Toads' />
-</items>
-]]></example>
+      <p>The subject node publishes the current subject of channel. Subject history is stored in MAM. A single item is stored in this node at a time which may contain one or more "text" elements that SHOULD posess an xml:lang attribute that describes the natural language of the subject. Users subscribe to this node to receive subject updates. The following example shows the format of a item in the subject node.</p>
```
Kev: I don't think that having a requirement to specify the language is practical. How about specifying it when more than one is present?

SamWhited: That's why it's a SHOULD; maybe it needs to be clarified

Kev: I think a SHOULD for something that generally impractical is wrong. I'd rather MUST possess xml:lang attributes if more than one text element is present.

SamWhited: I think most places that have similar elements just list this as a SHOULD, but that also makes sense to me. I'm happy either way, so I'll change it to a MUST if there is more than one `<text>` element.

stevekille:
I am happy with the conclusion here.

PR #239, opened 2016-08-28
XEP-0369: Remove circular definition

SamWhited:
Nit pick: The definition of the map node is somewhat circular. This tweaks the language to remove the circular definition using a synonym.

/cc @stevekille @Kev @stpeter


Kev:
I don't care one way or the other about this one.

stevekille:
I think this is an improvement

PR #240, opened 2016-08-28
XEP-0369: Minor formatting and grammar fixes

SamWhited:
* Fix examples that do not follow the proxy JID format as specified in section 3.5 and other small typo fixes.
* Change instances of "presence message" to "presence stanza" (a message is a thing in XMPP land so "presence message" doesn't make a lot of sense and is confusing)
* "UUID unique ID" is redundant; remove the "unique ID"

/cc @stevekille @Kev @stpeter


stpeter:
+1

linkmauve:
Why are changes like 7ecea1417b201591253853c6611a494b0e6a68ee required?  I already fixed all examples and it renders correctly on Firefox, does it fail by adding too many whitespaces on any other browser?

SamWhited:
Most of the examples still have extra whitespace around them for me on Firefox, but it might not be necessary. Oops, looks like I mixed up a few patches.

Kev:
+1

stevekille:
Most of this is good.  

I'm not keen on your "	Let's not be redundant and repeat ourselves"

Not sure all readers will know what UUID means.    Perhaps use the text to give an expansion of UUID.

PR #241, opened 2016-08-28
Explicit Message Encryption ProtoXEP: Add a 0.0.2 version

linkmauve:
- Made the 'name' attribute optional for existing mechanisms.
- Added a remark about the possibility to hide encrypted messages
  following user input.
- Made explicit that this protocol affects any encryption mechanism,
  present or future, not only those listed here.
- Display the namespace of the encryption mechanism in the default
  messages.

### DIFF
--- a/inbox/eme.xml
+++ b/inbox/eme.xml
@@ -30,6 +30,20 @@
     <jid>linkmauve@linkmauve.fr</jid>
   </author>
   <revision>
+    <version>0.0.2</version>
+    <date>2016-08-28</date>
+    <initials>egp</initials>
+    <remark><ul>
+      <li>Made the 'name' attribute optional for existing mechanisms.</li>
+      <li>Added a remark about the possibility to hide encrypted messages
+        following user input.</li>
+      <li>Made explicit that this protocol affects any encryption mechanism,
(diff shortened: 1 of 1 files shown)


inbox/eme.xml
```diff
@@ -104,25 +118,31 @@
...
+    <p>🔒 This message was encrypted with OpenPGP for XMPP
+      (urn:xmpp:openpgp:0), <link url="#">click here</link> to enable this
+      plugin.</p>
   </div>
   <div class='example'>
-    <p>🔒 This message was encrypted with OpenPGP for XMPP, downloading and
-      installing an appropriate plugin, your message will be displayed
-      shortly.</p>
+    <p>🔒 This message was encrypted with OpenPGP for XMPP
+      (urn:xmpp:openpgp:0), downloading and installing an appropriate plugin,
```
Flowdalic: I think the wording can be improved. How about something like

> This message was encrypted with OpenPGP for XMPP (urn:xmpp:openpgp:0). Download the <link url='#'>OpenPGP plugin</link> to enable this encryption method.

linkmauve: I actually just dropped this message, the first one was enough.

PR #248, opened 2016-09-22
Update OMEMO to use Olm instead of Axolotl

SamWhited:
Made a quick pass over the OMEMO XEP and swaped out references to Axolotl for Olm (leaving Axolotl where appropriate). Also did some minor cleanup and used XEP entities (already merged) to reference Olm and Curve25519 for consistent foot notes.

I did a read through after and didn't see anything that would break or needs to be changed except maybe the implementation notes, for which I left a TODO comment.

/cc @strb @iNPUTmice

### DIFF
--- a/inbox/omemo.xml
+++ b/inbox/omemo.xml
@@ -108,7 +108,7 @@
       <di><dt>OMEMO element</dt><dd>An &lt;encrypted&gt; element in the urn:xmpp:omemo:0 namespace. Can be either MessageElement or a KeyTransportElement</dd></di>
       <di><dt>MessageElement</dt><dd>An OMEMO element that contains a chat message. Its &lt;payload&gt;, when decrypted, corresponds to a &lt;message&gt;'s &lt;body&gt;.</dd></di>
       <di><dt>KeyTransportElement</dt><dd>An OMEMO element that does not have a &lt;payload&gt;. It contains a fresh encryption key, which can be used for purposes external to this XEP.</dd></di>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Hi,

two things
* Olm doesn't have the concept of Signed PreKeys. Thus any mention of Signed PreKeys should be removed from the XEP
* I would really like to incorporate the comment "The (ciphertext of the) payload should be authenticated in each SignalMessage, preferably as AAD." from the audit. That shouldn't be too difficult from a technical standpoint (when patching / adopting the library) but makes us unable to just 'refer' to the olm spec. Instead we might have to copy paste it?

SamWhited:
> Olm doesn't have the concept of Signed PreKeys. Thus any mention of Signed PreKeys should be removed from the XEP

Thanks, I need to change everything over to "root key" and "chain key" or whatever Olm calls them too. Anything else you notice like this please let me know.

> "The (ciphertext of the) payload should be authenticated in each SignalMessage, preferably as AAD."

I vaguely remember a discussion between you and Andy about deciding if we wanted to copy paste the spec or not; it makes senes to me to describe the crypto as well as the wire format, especially if we're going to make tweaks, but I think it might be best to get a draft using Olm out first and then start work on another one that includes the rest of the spec.

iNPUTmice:
Ok. I thought a little bit about it and we should be able to incorporate the Audit feedback without changing Olm.
I think I'm going to write that down very quickly before we let this become an official XEP. Otherwise we would technically have to do a name space bump afterwards.

SamWhited:
Sounds good; feel free to work on this branch if you want it, or we can throw it away and use yours.

SamWhited:
Superceded by #251

PR #249, opened 2016-09-28
OpenPGP IM: Explicitly require the support of all stanza types

fbeutel:
Explicitly mention that all stanza types are supported.
Also, be stricter with the wording on transparent handling of stanzas.

### DIFF
--- a/xep-0374.xml
+++ b/xep-0374.xml
@@ -50,6 +50,12 @@ Standards Foundation.</permissions>
     <jid>valodim@stratum0.org</jid>
   </author>
   <revision>
+    <version>0.1.2</version>
+    <date>2016-09-28</date>
+    <initials>fb</initials>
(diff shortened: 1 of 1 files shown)


SamWhited:
I got ahold of the author and discussed this today; they have rejected the change and will get back to you soon with discussion or a rational.

Flowdalic:
@fbeutel feel free to bring this to the standards@ mailing list if you want to discuss the rationale. The gist is that I don't think that just because an entity announces feature X together with announcing support for OX, it should automatically supports also OX encrypted stanzas of X. This would require heavy implementation work and refactoring on e.g. client libraries when they add OX, vs simply adding the OX extension elements.

Note that this is a generic design pattern when it comes to end-to-end encryption schemes and XMPP. For example OMEMO would have to come up with an answer for the same question when support for arbitrary extension elements would be added.

PR #251, opened 2016-10-01
Update OMEMO XEP to use Olm and some minor changes

iNPUTmice:
Update the OMEMO XEP to reference Olm instead of axolotl (work by @SamWhited )
mark the use of PreKeyMessages with an extra optional attribute (as suggested by @chrisballinger) 
put the gcm authentication tag into the encrypted key element.

### DIFF
--- a/inbox/omemo.xml
+++ b/inbox/omemo.xml
@@ -1,6 +1,6 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <!DOCTYPE xep SYSTEM 'xep.dtd' [
-  <!ENTITY % ents SYSTEM 'xep.ent'>
+  <!ENTITY % ents SYSTEM "xep.ent">
 %ents;
 ]>
 <?xml-stylesheet type='text/xsl' href='xep.xsl'?>
@@ -8,13 +8,7 @@
 <header>
   <title>OMEMO Encryption</title>
   <abstract>This specification defines a protocol for end-to-end encryption in one-on-one chats that may have multiple clients per account.</abstract>
-  <legal>
-    <copyright>This XMPP Extension Protocol is copyright (c) 1999 - 2014 by the XMPP Standards Foundation (XSF).</copyright>
-    <permissions>Permission is hereby granted, free of charge, to any person obtaining a copy of this specification (the &quot;Specification&quot;), to make use of the Specification without restriction, including without limitation the rights to implement the Specification in a software program, deploy the Specification in a network service, and copy, modify, merge, publish, translate, distribute, sublicense, or sell copies of the Specification, and to permit persons to whom the Specification is furnished to do so, subject to the condition that the foregoing copyright notice and this permission notice shall be included in all copies or substantial portions of the Specification. Unless separate permission is granted, modified works that are redistributed shall not contain misleading information regarding the authors, title, number, or publisher of the Specification, and shall not claim endorsement of the modified works by the authors, any organization or project to which the authors belong, or the XMPP Standards Foundation.</permissions>
-    <warranty>## NOTE WELL: This Specification is provided on an &quot;AS IS&quot; BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied, including, without limitation, any warranties or conditions of TITLE, NON-INFRINGEMENT, MERCHANTABILITY, or FITNESS FOR A PARTICULAR PURPOSE. In no event shall the XMPP Standards Foundation or the authors of this Specification be liable for any claim, damages, or other liability, whether in an action of contract, tort, or otherwise, arising from, out of, or in connection with the Specification or the implementation, deployment, or other use of the Specification. ##</warranty>
-    <liability>In no event and under no legal theory, whether in tort (including negligence), contract, or otherwise, unless required by applicable law (such as deliberate and grossly negligent acts) or agreed to in writing, shall the XMPP Standards Foundation or any author of this Specification be liable for damages, including any direct, indirect, special, incidental, or consequential damages of any character arising out of the use or inability to use the Specification (including but not limited to damages for loss of goodwill, work stoppage, computer failure or malfunction, or any and all other commercial damages or losses), even if the XMPP Standards Foundation or such author has been advised of the possibility of such damages.</liability>
-    <conformance>This XMPP Extension Protocol has been contributed in full conformance with the XSF's Intellectual Property Rights Policy (a copy of which may be found at &lt;<link url='http://xmpp.org/extensions/ipr-policy.shtml'>http://xmpp.org/extensions/ipr-policy.shtml</link>&gt; or obtained by writing to XSF, P.O. Box 1641, Denver, CO 80201 USA).</conformance>
-  </legal>
+  &LEGALNOTICE;
   <number>xxxx</number>
   <status>ProtoXEP</status>
   <type>Standards Track</type>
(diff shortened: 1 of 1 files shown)


ara4n:
Hey folks - this looks very cool. One proviso: we're currently having Olm audited by NCC Group, and something that has been raised is whether our reasons for not signing prekeys are valid.  We'll have more info during the audit talk-through on Tuesday (Oct 4), but it's possible that it'll be signing pre-keys after all...

iNPUTmice:
Hi @ara4n 

what are your reasons not to sign prekeys by the way? As I understand without signed prekeys you risk loosing forward secrecy on all prekey messages because the server can replace them.

ara4n:
The original rationale was to improve deniability, but it is looking like this is increasingly misguided. Will update once as soon as we've got a proper answer.

SamWhited:
@iNPUTmice FYI, we might want to hold off for a little bit and at least see what these are like and how they're licensed: https://moderncrypto.org/mail-archive/messaging/2016/002249.html

iNPUTmice:
@SamWhited if I understand them correctly they just want to publish the algorithm (How the ratchet is generate from the prekeys and How you create message keys from the ratchet). Which is great but this won't include a wire format or a library. So OMEMO would still need a library/wire format underneath which could be Olm. I imagine that once the spec is out (and maybe an RFC or standard of some sort) Olm could make use of that standard and still keep it's wire format. (Which our XEP would then inherit basically) 

So holding of for that reason doesn't seem right to me.

I would however wait a couple of days until @ara4n comes back with the signed prekey / not signed prekey decision and eventually revert that 'remove signed prekeys' commit.

chrisballinger:
There is also an independent SignalProtocol RFC in development that will be accompanied by permissively licensed source: https://twitter.com/matthew_d_green/status/782251561762381824

Due to the terms of our current contract we will initially be supporting OMEMO with SignalProtocol/AxolotlV3 that will be compatible with the current version of Conversations. However, we are designing the code to easily support both namespaces and allow us to add support Olm without much additional work.

Another interesting development: people are also now actively working on [OTRv4](https://github.com/twstrike/otrv4), which will include similar features like async crypto, prekeys, etc.

ara4n:
@SamWhited: as @iNPUTmice says, i think OWS are just publishing their equivalent of the Olm spec but for the libsignalprotocol implementation of the double ratchet; there's no new code there.  I'm intrigued to see how it will compare to the terse-but-complete Olm docs!

@iNPUTmice: the situation on signed prekeys is that @richvdh and myself have been led to believe that the original rationale for not signing them (i.e. deniability) is illfounded given that users publish their signed prekeys to the general public, you still get deniability given you can't prove who of the population actually picked up and used the prekeys.  (And separately from that, Matrix is currently a trainwreck for deniability given everyone's metadata is cryptographically baked into the message DAG for a room).  So, the likelihood is high that we will add signed prekeys back in.  However, as a final check we want to run this past @NegativeMjark who wrote Olm in the first place and made the original call to not sign them, and he's on holiday until Monday.

@chrisballinger: the nascent RFC work isn't for SignalProtocol or Signal, afaik, but one of the other double ratchet implementors looking at trying to turn their impl into a standard.  In terms of contractual obligations on the ChatSecure side, it may be worth bearing in mind that the initial NCC audit of Olm is complete, although it will be ~4 weeks until they publish the public report.

chrisballinger:
@iNPUTmice Here are some questions/comments I had about the spec while we were implementing it:

1. Based on the audit recommendations, I think a `date` attribute should be added to each `<device>`, so clients can stop sending messages to outdated devices. Not sure what the suggested logic should be for this, and if it should be quantized to prevent info leak?
2. I think an optional `name` attribute for `<device>` would be helpful to help distinguish between devices. Cryptocat seems to have support for this in their implementation.
3. What is the suggested behavior when there are `rid` collisions? For example, when sending a message to your own multiple devices and contact's multiple devices, what happens when there is a collision in `rid`?  There is a small chance of collision for randomly generated `rid`s but, if for whatever reason a maliciously added `rid` is added for an implicitly trusted identityKey, what should happen?
4. It's mentioned that the GCM auth tag SHOULD be >= 128 bit, but how would we know if it's larger? We are using libgcrypt which requires us to know the tag size and location in advance and are currently assuming a 128-bit tag appended to the end.

dwd:
Let's just get it numbered for now. Giving a number to a moving target is
tricky - once it's a XEP it removes all the IPR concerns and we can discuss
in detail on the mailing list (and in the chatrooms).

iNPUTmice:
> It's mentioned that the GCM auth tag SHOULD be >= 128 bit, but how would we know if it's larger? We are using libgcrypt which requires us to know the tag size and location in advance and are currently assuming a 128-bit tag appended to the end of the payload.

It's under specified for the current implementation with the ```conversations``` namespace. In there you just have to trust that it is the last 16 bytes.

The current PR clarifies that though by putting it in the ```<key>```. The key length is specified to be 16 bytes and the auth tag is just the rest.

> What is the suggested behavior when there are rid collisions? For example, when sending a message to your own multiple devices and contact's multiple devices, what happens when there is a collision in rid? There is a small chance of collision for randomly generated rids but, if for whatever reason a maliciously added rid is added for an implicitly trusted identityKey, what should happen?

For now I'd say if decryption fails try the next matching rid.

> I think an optional name attribute for `<device>` would be helpful to help distinguish between devices. Cryptocat seems to have support for this in their implementation.

We deliberately decided against names because names can be used to trick people into accepting something they shouldn't. Trust should be established on fingerprint alone.

> Based on the audit recommendations, I think a date attribute should be added to each <device>, so clients can stop sending messages to outdated devices. Not sure what the suggested logic should be for this, and if it should be quantized to prevent info leak?

I think we should just store the last time we received a message from a device and if that age is above a certain threshold we should stop sending messages to that device. A date in PEP can be manipulated by the server admin.

> Let's just get it numbered for now. 

Let's just wait for the signed Pre Key decision. I would feel more comfortable (security wise) to have it in there.

But in general I agree that all further discussion that are not directly related to the changes in this PR should happen on the mailing list.

SamWhited:
(Comment posted with the disclaimer that I haven't really thought about the implications of any of this)

> 1. Based on the audit recommendations, I think a `date` attribute should be added to each `<device>`, so clients can stop sending messages to outdated devices. Not sure what the suggested logic should be for this, and if it should be quantized to prevent info leak?

Using a date is tricky; many operating systems don't have a monotonically increasing clock which could cause weird issues around eg. leap seconds.

If the risk of a user happening to break something by using OMEMO during a leap second is considered low enough (it probably is), then I'd recommend including an expiration date and an issued date (valid_before and valid_after); this would allow schemes where we go ahead and register devices that will be available in the future, but aren't right now (and shouldn't be trusted right now). Not sure if/when this would be used, but it's nice to have the option.

> 2. I think an optional `name` attribute for `<device>` would be helpful to help distinguish between devices. Cryptocat seems to have support for this in their implementation.

In XMPP at least this is already covered by doing a disco#info on the device; I'm not sure it's necessary to add it to `<device>` (although I'm not sure that it hurts either, and it saves a round trip; I'm not necessarily disagreeing, just pointing out that an alternative already exists).

chrisballinger:
Moved discussion to the mailing list 📬

ara4n:
@iNPUTmice just to confirm, Olm is indeed going to sign its ephemeral keys; we're making the changes over the next week or so.

iNPUTmice:
I brought the signed prekeys back (even though the olm library hasn't been updated yet). Also squashed this should be ready for a merge now.
A rendered version can be found here: https://gultsch.de/omemo.html

SamWhited:
Inbox updated: https://xmpp.org/extensions/inbox/omemo.html

Flowdalic:
Council card: https://trello.com/c/7CAEi8qL

ara4n:
@iNPUTmice - sorry for lack of update here; have been on holiday. In Matrix we ended up doing the prekey signing at the application layer rather than in libolm, choosing to keep both modes of operation in libolm depending on the application's requirements - see http://matrix.org/git/olm/tree/docs/signing.rst for the justification.

https://github.com/matrix-org/matrix-js-sdk/pull/243 is an example of what got changed at the application layer, fwiw.

iNPUTmice:
@ara4n Thank you for pointing that out. I will see if we can mention this in the XEP somewhere.

ara4n:
cool. my point is mainly that the libolm library didn't end up sprouting support for the prekey signing; instead it'd be OMEMO's problem.  It's not exactly a hard thing to implement though :)

PR #252, opened 2016-10-03
Automatically generate the diff URL from the previous and current version for announce

linkmauve:
This prevents the editor from having to think about the correct URL.


Kev:
Thanks. I'm not sure whether there might be cases that this doesn't work, but let's find out :)

PR #257, opened 2016-10-06
XEP-0115: Added registrar consideration about stream features

linkmauve:


SamWhited:
Merged as c0407412362b6687ab92805fcb76f2523a04273e

PR #262, opened 2016-10-16
XEP-0225 fix stream feature example

sonnyp:


Flowdalic:
I'm not sure if this change is correct, as I do believe that it's indeed the server sending the

```
<stream:stream
       from='example.com'
       id='gPybzaOzBmaADgxKXu9UClbprp0='
…
```

stream open element and not the component.

sonnyp:
@Flowdalic what makes you think that? wouldn't make much sense IMHO

SamWhited:
@sonnyp Both clients and servers send an opening stream tag. However, since this opening stream is followed up by stream features, and only the server advertises features, it is the server stream. It's just there to illustrate that a stream is started, and features are sent that may include this particular feature. The example is correct as is.

sonnyp:
@SamWhited thanks

@Flowdalic my bad, read too fast.

Flowdalic:
> @Flowdalic my bad, read too fast.
No problem. :)

> However, since this opening stream is followed up by stream features, and only the server advertises features, it is the server stream.

Also note the `from='example.com'` in the open stream tag.

PR #263, opened 2016-10-17
XEP-0198: Send 'a' element before stream closure

sco0ter:
Clients should add a delay element when resending stanzas, so that the
original send date is preserved.


dwd:
These both look like sensible changes, but I'd like to run them past Council and the standards list first.

sco0ter:
Ok, thanks! It's just something which I've noticed during client development.
Related mail: https://mail.jabber.org/pipermail/standards/2015-July/030088.html

Flowdalic:
Created card for council: https://trello.com/c/fXsAbLzM

stpeter:
+1 - this is a nice change. :-)

PR #265, opened 2016-10-19
Change title to Message Retraction

legastero:


intosi:
Forgot to bump the version? Seems like there are two 0.0.2 entries inthe revisions section.

legastero:
ha, fixed. thanks @intosi.

PR #266, opened 2016-10-20
Fix Bylaws link

horazont:
The ``&BYLAWS;`` URL was broken (404). Fixing this is in the sprit expressed in xsf/xmpp.org#184.

The other URLs in xep.ent work (except for XMPPICA), but they lead to redirects. This could be fixed, but I do not know if the URLs which are currently used in these cases are well-defined in any manner and are intended to be used that way.


bear:
thanks!

PR #268, opened 2016-10-26
Check for dead links when building XEP

SamWhited:
Error if building an XEP contains dead links. Affects CI and normal users building XEPs.

EDIT: CI is failing because the step is working. Existing dead links should be fixed before this is merged.


SamWhited:
Closing this because it should really *only* check new XEPs, which will require more thought. Old XEPs may have dead links that should be left alone for historical reasons. It would also slow CI down a lot to check all XEPs.

PR #271, opened 2016-10-28
Fix tls reference in http upload xep

iNPUTmice:
The reference to the TLS rfc is wrong this commit fixes that.


iNPUTmice:
pinging @Flowdalic for author approval.

SamWhited:
Version bumped and published. Not sending announcement since this is purely an editorial fix.

Thanks!

PR #274, opened 2016-10-29
XEP-0001: Remove dead links

SamWhited:
Removed several dead links from XEP-0001. If the links are fixed or there are new pages for them, I'm happy to change to those pages, but they didn't appear to be anything we have anymore.


stpeter:
+1

PR #270, opened 2016-10-28
add abilty to discover entites that add stanza-ids and follow the rules

iNPUTmice:
stanza ids are not very useful in itself if you a receiving entity can not verify that the originating entity (the one in the by attribute) did in fact add that stanza id itself. A disco features solves this problem.

PR #275, opened 2016-10-30
XEP-0313: server MAY add the id to incoming messages

iNPUTmice:
Note that this PR depends on PR #270 being merged before.

This PR brings the old `<archive-id/>` back but as a `<stanza-id/>` defined in XEP-0359. This PR only influence the live incoming messages and doesn't change the syntax of regular interaction with MAM. It also doesn't include a namespace bump and is purely optional (MAY keyword). All security aspects are handled by the stanza-id XEP.

I know that the original archive-id was dropped because it doesn't have a way to communicate the id for outgoing messages but I figure it is better to have the ID half the time than never.
Ejabberd already adds the id (but with a wrong by attribute IIRC) and a patch in prosody would be extremely simple and will help clients a lot. This feature doesn't stand in the way of future refactors of the entire MAM spec which @mwild1 has been planning and is also entirely optional. I hope we can merge this to kinda bridge the gap until we have something better.

### DIFF
--- a/xep-0313.xml
+++ b/xep-0313.xml
@@ -28,6 +28,14 @@
   &mwild;
   &ksmith;
   <revision>
+    <version>0.5.2</version>
+    <date>2016-10-30</date>
+    <initials>dg</initials>
+    <remark>
+      <p>Added method for server to communicate the archive id</p>
+    </remark>
+  </revision>
+  <revision>
     <version>0.5.1</version>
     <date>2016-03-07</date>
     <initials>XEP Editor (ssw)</initials>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Rendered version can be found here: https://gultsch.de/xep-0313.html#archives_id

iNPUTmice:
I again changed the wording a little on what the by-attribute should contain. I think in the context of the entire XEP this is now as clear as it gets.

SamWhited:
Author coms sent/

xep-0313.xml
```diff
@@ -169,6 +177,17 @@
...
 	</section2>
+	<section2 topic='Communicating the archive ID' anchor='archives_id'>
+		<p>When an incoming message is archived, the server MAY add an &lt;stanza-id/&gt; element as defined in &xep0359; to the message, which informs the client of where and under what ID the message is stored. When doing this the server MUST follow the business rules defined in XEP-0359. The 'by'-attribute MUST be set to the address of the archive. For regular users that’s the bare JID of the account and for MUC that’s the bare JID of the room.</p>
+		<p>Servers MUST NOT include the &lt;stanza-id/&gt; element in messages addressed to JIDs that do not have permissions to access the archive, such as a users’s outgoing messages to their contacts. However servers SHOULD include the element as a child of the forwarded message when using &xep0280;</p>
```
Kev: I think having SHOULD here is in conflict with the MAY in the previous paragraph, isn't it?

I'd have thought that if we bring the MAM IDs back, we'd want to make them compulsory and bump namespace, wouldn't we?

mwild1: @Kev What do you think about s/MAY/SHOULD/ and not bumping namespace? I've been working on a new revision with a collection of changes, and have persuaded myself for now that keeping the current namespace is best. I don't think this change will change my mind on that, unless you have a convincing argument (which I'd be willing to hear).

iNPUTmice: In the context of my original PR (with the MAY) this is meant as
> However **if servers decide to add stanza-ids they** SHOULD include the element as a child of the forwarded message when using &xep0280;

But I'm of course also fine with SHOULD instead of MAY.

@Kev a namespace bump isn't necessary because the security aspects are handled in the stanza-id XEP.

SamWhited:
I followed up with Matt who says that he's spoken with Daniel and Kev about this and that there are still changes that need to go in and that he has a local copy that needs a bit more work before it can be submitted. Leaving this open for now.

SamWhited:
Closing this to clean up the editors queue. Please reopen or submit a new PR when the authors come to an agreement; thanks.

iNPUTmice:
Hi,

I want to urge everybody to get this merged. I mean no disrespect to the original author but the fact is that nothing has happened on this front for months.

Here are my arguments on why this should be merged ASAP
* There is no argument that injecting the MAM id into live messages is a good thing. The only open question is on whether to do this in the MAM namespace (which would require a NS bump) or with stanza-ids (where the general consensus is that we can stick with the `urn:xmpp:mam:1` namespace
* The suggested changes do not stand in the way of potential future changes from the author. Even if the original author decides to bump the NS later for other reason those stanza-ids don't hurt.
* The changes described here actually reflect the reality of the implementations. Both ejabberd and prosody (the internal MAM in 0.10) do inject stanza-ids

I'm by the way fine with changing the MAY to SHOULD or MUST. The MAY was only born out of the idea to be as conservative as possible on not hinder what ever @mwild1 plans to do with the XEP. However from a chat I had with him a while back it seems that he'd actually prefer something stronger.

SamWhited:
I don't appear to be able to reopen this for some reason (maybe GitHub's copy of the branch was deleted or something); but a new PR including a summary of the original discussion and the comment you just posted might make it more visible. I also would like to see this done.

PR #287, opened 2016-11-14
XEP-0124: Fix typo and DTD

SamWhited:
Editorial change, but it's to a draft document. As usual, I cannot remember what that means despite the fact that I do this every time.


SamWhited:
Closed by e14fc07

PR #291, opened 2016-11-23
Fix address

stpeter:


SamWhited:
Cherry picked 505a7a9 as 08cfcb9

PR #301, opened 2016-12-01
XEP-0071: Inline external image and fix DTD

SamWhited:
The important thing in this PR is the addition of several elements in the DTD. I'm not sure if we want to add these, or want to fix the XEP to not include them. I could go either way. Thoughts welcome.

/cc @stpeter @Flowdalic


SamWhited:
Going ahead and merging, we can revert later if the DTD fixes aren't desirable.

PR #305, opened 2016-12-03
0060 multimod

stpeter:


xep-0060.xml
```diff
@@ -4412,7 +4418,7 @@ And by opposing end them?
     id='subman3'>
   <pubsub xmlns='http://jabber.org/protocol/pubsub#owner'>
     <subscriptions node='princely_musings'>
-      <subscription jid='bard@shakespeare.lit' subscription='subscribed'/>
+      <subscription jid='polonius@denmark.lit' subscription='subscribed'/>
```
SamWhited: NIT: I briefly remember discussion about using `example.net` instead of `*.lit` since technically .lit could become a TLD in the future. I'm not super concerned either way, especially since this is just modifying it to be consistent with the text which already used `.lit`.

stpeter:
We can also use foo.example. I suggest that we submit a separate issue for this and fix all the examples in this XEP (or all the XEPs) at the same time.

SamWhited:
Cherry-picked as 1.13.4 / 16194d4

PR #312, opened 2016-12-06
XEP-0367 and XEP-0363: Content Edits.

Shriank:
Content edits for making "Security Considerations" easier to read.

### DIFF
--- a/xep-0363.xml
+++ b/xep-0363.xml
@@ -90,8 +90,8 @@
 </section1>
 <section1 topic='Requirements' anchor='reqs'>
   <ul>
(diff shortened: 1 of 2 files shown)


Shriank:
Thanks!

SamWhited:
Hi @Shriank; I thought about this a bit more today and decided not to merge these edits; the language is fine, but it's also fine as is so there's no real value in changing it (it just takes editor time to publish the new version). Thanks for your willingness to submit them though and let me know if you have anymore things like this in the future; don't let this particular PR discourage you. Thanks!

PR #318, opened 2016-12-08
XEP-0053: Require a registry assignment policy

SamWhited:
More (potential) changes to XEP-0053. This defines a handful of registry assignment policies and requires that new registries define a policy (or reference one of the policies defined in 0053).


stpeter:
If we're going to pursue this, I think it is best discussed on the standards@ list.

However, I have my doubts about pursuing it. The IETF has been moving toward "First Come, First Served" in most cases, and I don't see a compelling reason for the XSF to move in the opposite direction by adding more formality to our processes.

SamWhited:
Makes good sense to me; thread started here: https://mail.jabber.org/pipermail/standards/2016-December/031771.html

SamWhited:
This didn't really go anywhere; closing and it can be addressed on a case by case basis if another issue arrises or if the 0045 issue is brought back up.

PR #335, opened 2016-12-29
XEP-0375: Add e2e encryption to compliance suites and remove MIX

SamWhited:
Adds End-to-End encryption to the compliance suites for instant messaging with OMEMO as the only provider (in my mind not all use cases for XMPP need e2e encryption, so I didn't add it to the core suites).

Also removes MIX; I'm still torn about this, but it seems like most of the community wanted it removed for now.

This also bumps the year to 2017.

### DIFF
--- a/xep-0375.xml
+++ b/xep-0375.xml
@@ -6,9 +6,9 @@
 <?xml-stylesheet type='text/xsl' href='xep.xsl'?>
 <xep>
   <header>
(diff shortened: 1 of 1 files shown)


SamWhited:
After discussion with the council it was decided that bumping the year should be handled differently and that OMEMO wasn't ready yet. Closing for now.

PR #346, opened 2017-01-03
Add PDF generation back to Makefile

SamWhited:
Adds a make target for `xep-xxxx.pdf` and gitignores pdf files in the main directory.

Something similar to this was removed a while back, but I'm not sure why. Would anyone complain if I added it back?


SamWhited:
Merged as 30bb05ba941e1b801529f361b1ad409c4aec5531

PR #354, opened 2017-01-06
XEP-0383: Add an XML Schema

SamWhited:
No idea if this makes sense. An online validator thing appears to validate everything properly.

### DIFF
--- a/xep-0383.xml
+++ b/xep-0383.xml
@@ -241,7 +241,39 @@
   </section2>
 </section1>
 <section1 topic='XML Schema' anchor='schema'>
-  <p>TODO.</p>
+  <code><![CDATA[
+<?xml version='1.0' encoding='UTF-8'?>
+
+<xs:schema
+    xmlns:xs='http://www.w3.org/2001/XMLSchema'
+    targetNamespace='urn:xmpp:burner:0'
+    xmlns='urn:xmpp:burner:0'
+    elementFormDefault='qualified'>
+
+  <xs:annotation>
+    <xs:documentation>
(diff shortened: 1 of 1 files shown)


xep-0383.xml
```diff
@@ -241,7 +241,39 @@
...
+    <xs:complexType>
+      <xs:sequence minOccurs="0" maxOccurs="1">
+        <xs:element name='jid' type='bareJIDType'/>
+      </xs:sequence>
+    </xs:complexType>
+  </xs:element>
+
+  <xs:simpleType name='bareJIDType'>
+    <xs:restriction base='xs:string'>
+      <xs:minLength value='5'/>
```
linkmauve: Why is there this restriction?  A valid bare JID could be “a@a” which is three characters, or even “a” which is one character (even though it doesn’t really make sense in this context, a specific deployment could hand out domain JIDs).

SamWhited: Probably because I can't count. Thanks!

xep-0383.xml
```diff
@@ -241,7 +241,39 @@
...
+      <xs:maxLength value='2047'/>
```
linkmauve: This restriction is slightly wrong, as per the XML Schema specification, “the unit of length is measured in number of list items”, here a character and not a byte.

SamWhited: Oh good catch; I'm not really sure what to do about this one. I suspect many of our schemas have this issue (since I think I copy/pasted this from somewhere else and tweaked the numbers).

PR #355, opened 2017-01-06
Use LaTeX makeglossary tool during PDF generation

SamWhited:
Experimenting with using actual LaTeX glossaries for PDF generation. I'm not 100% sure that there's a point to this, but it was good for figuring out how the build process worked. It might be nice for eventually being able to generate a global glossary PDF from all XEPs…

This adds a optional &lt;glossary&gt; node which looks exactly like a &lt;section1&gt; node in the HTML, but is converted to a LaTeX glossary for the PDF. The only thing that is valid in the glossary is a &lt;dl&gt;.


SamWhited:
Fun experiment; probably not worth merging. Closing.

PR #356, opened 2017-01-06
Remove scripts

SamWhited:
I don't think these scripts are in use anymore. If someone could confirm, it would be appreciated.


SamWhited:
I don't have the confidence to merge this; it doesn't really matter anyways. Will ask the iteam what can be cleaned up at some point.

Kev:
I don't feel confident about this either. I don't believe these are used any more, but maybe @stpeter remembers their original use.

Flowdalic:
Remove, and if something breaks simply revert?

PR #357, opened 2017-01-08
XEP-0153 and XEP-0008 are superseded by XEP-0084

SamWhited:
Changes the status of XEP-0153 to "Obsolete" so that it will stop showing up in the list and people will stop implementing it (XEP-0084 was already marked as superseding it).

Also fixes the 'supersededby' element in both 0153 and 0008.

Trello: https://trello.com/c/Ka3zzdG1
Mailing list: https://mail.jabber.org/pipermail/standards/2017-January/031832.html

### DIFF
--- a/xep-0008.xml
+++ b/xep-0008.xml
@@ -19,7 +19,7 @@
     <spec>XMPP IM</spec>
   </dependencies>
   <supersedes/>
-  <supersededby/>
+  <supersededby><spec>XEP-0084</spec></supersededby>
   <shortname>None</shortname>
   &temas;
   &xvirge;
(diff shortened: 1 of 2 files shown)


linkmauve:
A big regression from XEP-0153 to XEP-0084 is that avatars can’t work in MUC or with contacts you have no from subscription with, it is possibly the reason XEP-0084 has had such a low adoption rate.

Adding a note about this would be sensible.

iNPUTmice:
> A big regression from XEP-0153 to XEP-0084 is that avatars can’t work in MUC or with contacts you have no from subscription with, it is possibly the reason XEP-0084 has had such a low adoption rate.

they do work in MIX though. And with MIX having a number now and everything I don't see a reason not to obsolete XEP-0153.

SamWhited:
-1's from Tobias and Dave. Closing.

PR #360, opened 2017-01-09
Don't duplicate footnotes in HTML

SamWhited:
I was just copy/pasting and tweaking XSLT from StackOverflow, so someone who has any idea how this generate id stuff works should probably review this.

This also needs to be done for the PDFs probably, but this can be a separate PR.
We probably also want a way to cite a footnote from something that has different text than the footnote (eg. `&xmppcore;` and `<cite>XMPP Core</cite>` and `<cite>RFC 6120</cite>` should point to the same thing); this can also come later.

### DIFF
--- a/xep.xsl
+++ b/xep.xsl
@@ -4,28 +4,28 @@
 
 Copyright (c) 1999 - 2015 XMPP Standards Foundation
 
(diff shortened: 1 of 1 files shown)


tfar:
Looks good to me. Can be merged.

PR #364, opened 2017-01-10
Mix

stevekille:
XEP -0369 - MIX

Update to 0.6.5

     Modify MIX Proxy so that client sends to bare JID


SamWhited:
Merged as 16318dc and b24567e

PR #370, opened 2017-01-17
Bind2.0

Kev:
Will comply with XSF's IPR rules, etc.


SamWhited:
Merged as c396e02; thanks! Will be reviewed by the council within two weeks, etc…

PR #372, opened 2017-01-17
Defers many old XEPs

### LINKED ISSUE #369: Defer experimental XEPs that have not been updated in 12 months
I suppose at some point we had XEPs automatically being deferred after 12 months of inactivity, but I was glancing through some XEPs and discovered quite a few that should be deferred but aren't.

It would be good to do a survey of existing experimental XEPs and defer the ones that haven't been touched in a while.

SamWhited:
Lots of old XEPs that haven't been touched in 12+ months and need deferring. Some of these are widely used and should probably be advanced; I'll ask the various people involved if they want to continue pushing them forward or try to advance them and we can undefer them if anyone wants to take over and champion an XEP on a case by case basis.

Fixes #369

### DIFF
--- a/xep-0215.xml
+++ b/xep-0215.xml
@@ -10,7 +10,7 @@
   <abstract>This document specifies an XMPP protocol extension for discovering services external to the XMPP network.</abstract>
   &LEGALNOTICE;
   <number>0215</number>
-  <status>Experimental</status>
+  <status>Deferred</status>
   <type>Standards Track</type>
   <sig>Standards</sig>
   <approver>Council</approver>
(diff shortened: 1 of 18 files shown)


SamWhited:
- Pinged Lance about XEP-0264
- Pinged Emil et al about  XEP-0298
- Pinged Spencer about XEP-0333
- Pinged Matt about XEP-0334 and XEP-0335

Zash:
How hard would it be to automate this?

SamWhited:
I'm sure it would be fairly easy for someone on the infrastructure team who cares enough to do it, although I'm always a tiny bit hesitant to have a bot comitting directly to the repository. Then again, if it saves us the trouble, I say let's do it.

Zash:
Maybe not committing directly, but if it could open PRs then the manual work is reduced to clicking 'merge'

PR #373, opened 2017-01-18
XEP-0286 clarifications

SamWhited:
Some simple clarifications to XEP-0286. Also a retab / rewrap, which makes the diff a bit nasty; probably best to look at one commit at a time.

### DIFF
--- a/xep-0286.xml
+++ b/xep-0286.xml
@@ -8,12 +8,12 @@
 <header>
   <title>Mobile Considerations</title>
 	<abstract>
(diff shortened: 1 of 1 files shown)


ge0rg:
Hey, good work on this. I saw the XEP in the deferred flood, reread it and found it quite up-to-date actually.

Could you please also add a CSI mention to section 5? It's referenced above, but having all the relevant things in one place would be great!

SamWhited:
Good catch; will do, thanks!

PR #375, opened 2017-01-18
Add 2017 TLS Recommendations to the inbox

SamWhited:
In preparation for the 2017 compliance suites, add an XEP making TLS recommendations.


stpeter:
We have RFC 7590 on this topic. It seems best to specify only things that have changed since then (i.e., don't repeat anything that's in RFC 7590).

SamWhited:
> We have RFC 7590 on this topic. It seems best to specify only things that have changed since then (i.e., don't repeat anything that's in RFC 7590).

I thought about that, but since RFCs aren't very agile it seemed like it would be nice to have something that we could update every year or two easier, and I didn't want people to have to look in two places.

That being said, you're probably right: I should at least mention RFC 7590, and if I do that I might as well just elid anything that's already in there. I'll think about it; thanks!

stpeter:
Yes, I know that RFCs aren't very agile, but we wrote that spec for a reason. I would strongly prefer to use a XEP only to supplement RFC 7590, basically as a way of documenting anything above and beyond RFC 7590. If the list of additions gets long enough, then it makes sense to update RFC 7590 (and RFC 7525, too).

SamWhited:
@stpeter I mentioned RFC 7590 in the intro, and removed one bit of duplicate information (the part about STARTTLS support being required); is there anything else you noticed was duplicated?

stpeter:
The "Versions" section is not consistent with RFC 7525 and it would be best to point to RFC 7525 rather than trying to repeat that information here (with or without errors). We also need to add RFC 7525 to the dependencies.

The "Authentication Mechanisms" section is about SASL, not TLS. Perhaps this document would be better renamed "Stream Security 2017".

The "Cipher Suites" section is, again, trying to repeat information from RFC 7525. Lots of hard work by real security experts went into RFC 7525 so I would strongly suggest that we just reference it here.

RFC 2119 does not define "NOT REQUIRED" as a compliance term so it's best not to use that phrase here.

We don't need the paragraph about "Legacy SSL 3.0 and earlier..." in the security considerations because that is already covered earlier in the document (preferably by pointing to RFC 7525).

SamWhited:
> The "Versions" section is not consistent with RFC 7525 and it would be best to point to RFC 7525 rather than trying to repeat that information here (with or without errors). We also need to add RFC 7525 to the dependencies.

Ah I see, thanks, I hadn't found 7525, I'll reference that instead.

> The "Authentication Mechanisms" section is about SASL, not TLS. Perhaps this document would be better renamed "Stream Security 2017".

I was specifically trying to discuss it's use of channel binding with TLS, but that could have been clearer.

> The "Cipher Suites" section is, again, trying to repeat information from RFC 7525. Lots of hard work by real security experts went into RFC 7525 so I would strongly suggest that we just reference it here.

I had a TODO (in a comment) to find a source to use for the ciphers since I hadn't put together a list yet; this would be perfect, thanks.

> RFC 2119 does not define "NOT REQUIRED" as a compliance term so it's best not to use that phrase here.

Oops.

Since most of this does appear to duplicate an existing RFC, it may be best if I withdraw it for now and just reference the RFC directly from the compliance suite. Thanks!

stpeter:
Ah yeah, that's another option. I mean, if we do have stuff that's not addressed by 7525/7590 then by all means let's document it somewhere. But your course of action seems best for the moment. Thanks for caring about security. :-)

PR #382, opened 2017-01-26
Carbons: add <no-copy/> hint in addition to <private/>

ge0rg:
This adds both <private> and <no-copy> into the XEP for maximum interoperability.

While Carbons is doing its rounds in Last Call, and people can't decide whether <no-copy> is less strict but more useful than <private>, this one adds a bit of overhead and removes a large burden from implementations only covering one of the two.

Edit: first discussion on standards@:
https://mail.jabber.org/pipermail/standards/2015-September/030288.html

### DIFF
--- a/xep-0280.xml
+++ b/xep-0280.xml
@@ -329,7 +329,7 @@
 
 </section1>
 <section1 topic='Avoiding Carbons for a single message' anchor='avoiding'>
-  <p>Some clients might want to avoid Carbons on a single message, while still keeping all of the other semantics of Carbon support. This might be useful for clients sending end-to-end encrypted messages, for example. To exclude a &MESSAGE; from being forwarded to other Carbons-enabled resources, the sending client add a &lt;private/&gt; element qualified by the namespace "urn:xmpp:carbons:2" as a child content element to the &MESSAGE; stanza.</p>
+  <p>Some clients might want to avoid Carbons on a single message, while still keeping all of the other semantics of Carbon support. This might be useful for clients sending end-to-end encrypted messages, for example. To exclude a &MESSAGE; from being forwarded to other Carbons-enabled resources, the sending client must add a &lt;private/&gt; element qualified by the namespace "urn:xmpp:carbons:2" and a &lt;no-copy/&gt; hint according to &xep0334; as child content elements to the &MESSAGE; stanza.</p>
 
   <p><strong>Note:</strong> use of the private mechanism might lead to partial conversations on other devices.  This is the intended effect.</p>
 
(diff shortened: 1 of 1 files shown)


xep-0280.xml
```diff
@@ -329,7 +329,7 @@
 
 </section1>
 <section1 topic='Avoiding Carbons for a single message' anchor='avoiding'>
-  <p>Some clients might want to avoid Carbons on a single message, while still keeping all of the other semantics of Carbon support. This might be useful for clients sending end-to-end encrypted messages, for example. To exclude a &MESSAGE; from being forwarded to other Carbons-enabled resources, the sending client add a &lt;private/&gt; element qualified by the namespace "urn:xmpp:carbons:2" as a child content element to the &MESSAGE; stanza.</p>
+  <p>Some clients might want to avoid Carbons on a single message, while still keeping all of the other semantics of Carbon support. This might be useful for clients sending end-to-end encrypted messages, for example. To exclude a &MESSAGE; from being forwarded to other Carbons-enabled resources, the sending client must add a &lt;private/&gt; element qualified by the namespace "urn:xmpp:carbons:2" and a &lt;no-copy/&gt; hint according to &xep0334; as child content elements to the &MESSAGE; stanza.</p>
```
linuxwolf: ought this "must" be **"MUST"**?

linuxwolf:
FWIW, other than the quip above, I'm +1.

xep-0280.xml
ge0rg: I have pondered about this as well, and came to the conclusion that a "must" better reflects the situation. This is not a strict protocol requirement on the one hand, and "MUST" might indeed require that we bump the version, which I don't like very much.

SamWhited: If this isn't a "MUST" in the RFC2119 sense, maybe this should be rephrased to prevent implementors from assuming it was a typo in the future then?

xep-0280.xml
```diff
@@ -341,6 +341,7 @@
   <body>Neither, fair saint, if either thee dislike.</body>
   <thread>0e3141cd80894871a68e6fe6b1ec56fa</thread>
   <private xmlns='urn:xmpp:carbons:2'/>
+  <no-copy xmlns="urn:xmpp:hints"/>
```
SamWhited: Not that it really matters, but we should probably use single vs. double quotes consistently.

xep-0280.xml
```diff
@@ -351,6 +352,7 @@
...
```
SamWhited: Ditto on the single vs. double quotes.

xep-0280.xml
SamWhited: If this isn't a "MUST" in the RFC2119 sense (and I agree, I don't think it is), maybe this should be rephrased to prevent implementors from assuming it was a typo in the future then? "the client might add…"?

ge0rg: The client **SHOULD** add? Sounds wrong too. When I thought about it, "must" was the best I came up with.

xep-0280.xml
ge0rg: Oops, I just copied that from the hints xep verbatim. Will fix soon

SamWhited:
I went ahead and made the suggested tweaks and merged as b3dee57aaf0996f03a7b31ed9f8527360483146e

PR #384, opened 2017-01-26
XEP-0340: Editorial changes, fixing syntax highlighting in examples

linkmauve:


SamWhited:
Recently someone told me they didn't like the idea of using comments in examples, even for illustration; I'm not really sure what we should od about this now.

I also don't think this needs a version bump, it doesn't actually affect the output text in any meaningful way (it's more like a formatting change).

linkmauve:
What were the arguments against?  In this case the SEND and RECV tags don’t add much information so I wouldn’t be against removing them altogether.  Another option would be to put them in the caption, encoding that in the XEP itself.

Btw I played with the idea of generating the examples directly from the XML, avoiding any potential not-well-formed issue, but my XSLT-fu is still way too weak.

SamWhited:
> What were the arguments against?

I'm not sure; might be worth a mailing list thread.

SamWhited:
Closing this for now since there's a conflict; thanks for the change though. I'll ask around and come up with a definitive way to format examples like this.

PR #385, opened 2017-01-27
XEP-0186: Defer after years of inactivity

SamWhited:
Also possibly a candidate to advance


stpeter:
We had some discussion about this on the standards@ list. There was agreement to get it done if someone can find the time and energy to check it over, incorporate feedback, and push it forward. I volunteered, but haven't made the time for it yet. :-)

SamWhited:
I added it to the council agenda since it seems like it might be a good candidate to move forward or find someone to hand hold it for a bit; thanks for the background info!

PR #386, opened 2017-01-27
XEP-0369 MIX changes to 0.7

stevekille:
Correct multi-language subject setting;
      Remove MIX Proxy terminology and replace with Participant Server Behaviour;


SamWhited:
Merged as e7eb898, 68f128e, 30cd35b, and fa5fac8

PR #390, opened 2017-01-28
List the supersededby in the abstract

linkmauve:
This makes it way more explicit when a XEP has been superseded by some other specification(s), while leaving the same current list in Appendix A.

This will require all of the HTML files to be rebuilt once it gets merged.

### DIFF
--- a/xep-0119.xml
+++ b/xep-0119.xml
@@ -7,7 +7,7 @@
 <xep>
 <header>
   <title>Extended Presence Protocol Suite</title>
-  <abstract>This document specifies a set of XMPP extensions that provide support for extended presence information. Note: This document has been retracted since its functionality is handled by XEP-0163.</abstract>
+  <abstract>This document specifies a set of XMPP extensions that provide support for extended presence information.</abstract>
   &LEGALNOTICE;
   <number>0119</number>
   <status>Retracted</status>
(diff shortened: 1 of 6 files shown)


stpeter:
👍

linkmauve:
The first version was only updating XEP-0270, this updated PR also removes duplicated text in the abstract of XEPs 0161, 0119 and 0281, as well as making 0281 and 0289 explicitly supersed{ed,ing}.

I haven’t bumped the revision of any of these XEPs because the change is purely cosmetic, please tell me if I should do so instead.

SamWhited:
Great idea! This is much better. Maybe instead of a line in the abstract we could add this as another metadata table entry? Eg. just under `Status: Obsolete` it could have a `Superseded by: …`

linkmauve:
That was my reasoning as well, I originally didn’t even notice the `supersededby` field in Appendix A, and adding yet another field in the metadata didn’t look any better.  My other option was to have it just follow the red retracted/obsolete status explanation.

SamWhited:
>  My other option was to have it just follow the red retracted/obsolete status explanation.

Great minds… linkmauve/xeps#1

linkmauve:
ITYM https://github.com/linkmauve/xeps/pull/2 :)

Anyway, this is indeed a more semantic solution, I merged it, thanks.

PR #392, opened 2017-01-28
XEP-0270, XEP-0302, XEP-0375: Supersede each other in that order

linkmauve:


SamWhited:
I'm not sure this is true; since 0302 was never accepted, it never superseded anything.

SamWhited:
This will be fixed shortly; I forgot about this PR and redid it while accepting the 2017 compliance suites today. Thanks!

PR #393, opened 2017-01-28
XEP-0302: Move to status Obsolete

linkmauve:
Council decided last week to obsolete XEP-0270 (2010 compliance suites) in favour of XEP-0375 (2016 compliance suites), it makes a lot of sense to also obsolete the 2012 version.

### DIFF
--- a/xep-0302.xml
+++ b/xep-0302.xml
@@ -10,7 +10,7 @@
   <abstract>This document defines XMPP protocol compliance levels for 2012.</abstract>
   &LEGALNOTICE;
   <number>0302</number>
-  <status>Deferred</status>
+  <status>Obsolete</status>
   <type>Standards Track</type>
   <sig>Standards</sig>
   <dependencies>


SamWhited:
Added to council agenda.

SamWhited:
Actually, these appear to be deferred; is there a point to obsoleting them? I don't think they were ever accepted in the first place, so obsolete might not make much sense.

linkmauve:
The deferred status gives mixed feelings, at times we have to tell people that it’s actually fine for them to implement a deferred XEP despite the scary red banner, at other times it acts as a mortuary for XEPs that didn’t advance in time (like here). Generally I’d like to have a much more explicit difference between these two, there are a lot of other XEPs I’d like to “remove” that way.

I wonder if a process change would be useful here, to have an explicit path from deferred to retracted, deprecated and obsolete.

stpeter:
The original idea behind Deprecated and Obsolete was that a spec can't even get to those states unless it's already been Draft or Final. However, looking again at the definitions in Sections 9.9 and 9.10 of XEP-0001, I see no reason why the Council could not change a Deferred XEP to Deprecated or Obsolete if it so chooses. I suggest that the Council discuss this.

SamWhited:
Discussed today; will be obsoleted shortly.

PR #394, opened 2017-01-28
Update http: URLs to https: in xep.ent

linkmauve:
The only functional difference is the removal of the XMPPICA entity, which was apparently a CA project in the XSF, but isn’t referenced anywhere.


linkmauve:
Err, just noticed that even though https://mail.jabber.org/mailman/listinfo/jsf-ietf/ redirects to https://mail.jabber.org/mailman/listinfo/xsf-ietf/ neither list exists.

stpeter:
Yes, we got rid of those lists at one point. We can remove any references to them.

PR #403, opened 2017-02-07
Implement initial last call comments for XEP-0368, editorial changes, SNI is now MUST

moparisthebest:
The only thing I'm not sure of is where the comments re: RFC 2595 belong @dwd I put them in IANA Considerations, but that somehow feels wrong, any thoughts?


stpeter:
Good fixes. :-)

PR #404, opened 2017-02-07
SASL2 - New SASL Profile

dwd:
As discussed at the Summit.

I declare that:

* I am legally able to assign copyright to the XMPP Standards Foundation
* I shall do so on acceptance by the Council
* I shall perfect as required.

Anything else I need to say?


Flowdalic:
I'd probably also wrap the initial response in next-authenticate in an extra element. While I can't think of a use case it seems worth to leave that door open.

PR #411, opened 2017-02-10
XEP-0387: several changes

ge0rg:
This PR contains three diffs to XEP-0387, in increasing order of controversy:

1. a minor typo fix
2. Addition of [XEP-0368: SRV records for XMPP over TLS](https://xmpp.org/extensions/xep-0368.html) as an Advanced Core feature
3. Removal of 'User Avatars' from IM-Core, making it IM-Advanced only.

Please feel free to apply any subset of these. I think that (3) is the most controversial one and needs the strongest rationale, so here it is:

* avatars require a graphical display to make sense (this is slightly mitigated by footnote 20)
* avatars consume network bandwidth
* avatars are more of a cosmetic element than a functional one (not having them will not cause your messages to get lost in transmission)
* there are interop problems with file formats (some clients insist on using webp :-P )
* there are interop problems with MUC

I'd rather see 0198 and [XEP-0184: Message Delivery Receipts](https://xmpp.org/extensions/xep-0184.html) as core features, before addressing avatars.

### DIFF
--- a/xep-0387.xml
+++ b/xep-0387.xml
@@ -39,6 +39,7 @@
       <spec>XEP-0280</spec>
       <spec>XEP-0313</spec>
       <spec>XEP-0352</spec>
+      <spec>XEP-0368</spec>
     </dependencies>
     <supersedes>
       <spec>XEP-0375</spec>
@@ -95,6 +96,14 @@
           <td align='center'>&#10003;</td>
           <td align='center'>&#10003;</td>
           <td>&rfc7590;</td>
+        </tr>
+        <tr>
+          <td><strong>Direct TLS</strong></td>
+          <td align='center'>&#10005;</td>
+          <td align='center'>&#10005;</td>
+          <td align='center'>&#10003;</td>
+          <td align='center'>&#10003;</td>
+          <td>&xep0368;</td>
         </tr>
         <tr>
           <td><strong>Feature discovery</strong></td>
--- a/xep-0387.xml
+++ b/xep-0387.xml
@@ -188,7 +188,7 @@
         <tr>
           <td><strong>User Avatars</strong></td>
           <td align='center'>N/A</td>
-          <td align='center'>&#10003;&nocli;</td>
+          <td align='center'>&#10005;</td>
           <td align='center'>N/A</td>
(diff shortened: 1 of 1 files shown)


SamWhited:
> a minor typo fix

Thanks! I'll cherry pick this one over.

> Addition of XEP-0368: SRV records for XMPP over TLS as an Advanced Core feature

Oops, I thought I added that already; I think I must have lost a few commits somewhere, because I was missing several others I was pretty sure I had too. I disagree with the levels a bit here though, so I'm just going to add it alongside RFC7590 in the core "TLS" section; see: 254df708483a04223b64b62385bb76b0b253c83f (maybe later with a note saying that it's about supporting direct TLS itself, not necessarily about having SRV records if they're not needed).

> Removal of 'User Avatars' from IM-Core, making it IM-Advanced only.

This one I disagree with entirely; to have a good, consistent chat experience, avatars are necessary (except where they don't make sense, as already noted).

ge0rg:
> This one I disagree with entirely; to have a good, consistent chat experience, avatars are necessary 

[I know I'm not a typical user, but I'm living without avatars since day 1]

As we are not going to agree and this point is important to me, what is the correct escalation process? Do I need to submit a forked copy of the XEP and let the board vote? ;-)

SamWhited:
I think there are three possible courses of action, and I'm not really sure which one is correct, so you might ask in the XSF room:

1. Submit a competing compliance suites XEP and wait for the council to vote to advance one of them and obsolete the other (although I'd note that I suspect this would just be confusing for anyone looking for BCP's or compliance suites)
2. Bring it up on the list and try to get enough consensus to convince me to change the existing XEP
3. Wait until we vote to advance this one, and then bring it up during LC and see what the council decides

ge0rg:
> Submit a competing compliance suites XEP

It would also mean that I have to keep track of all your changes plus my diff and update the alt-XEP, and this doesn't really look like a nice and friendly way of handling such things...

> Bring it up on the list and try to get enough consensus to convince me to change the existing XEP

I've brought it up on the list and you were the only one to respond. I would rather not speculate about the reasons, though... ;-)

> Wait until we vote to advance this one

That actually sounds like a reasonable approach, thanks.

PR #415, opened 2017-02-11
XEP-0387: Relax some requirements, and add important dependencies from 2017

linkmauve:
This PR does these changes to the latest compliance suites:
- Add XEP-0156: Discovering Alternative XMPP Connection Methods for web clients.
- Add XEP-0071: XHTML-IM for message formatting.
- Allow web clients to use a server component connecting directly with TCP.
- Make XEP-0357 optional on platforms where it is unnecessary.

I can split some commits in a different PR if you decide some items should have some more discussion, but I believe these are the right way to go at least for this year.

### DIFF
--- a/xep-0387.xml
+++ b/xep-0387.xml
@@ -6,6 +6,7 @@
 <!ENTITY usecases "<note>Support for the Entity Use Cases and Occupant Use Cases is REQUIRED; support for the remaining use cases is RECOMMENDED.</note>">
 <!ENTITY onlyone "<note>Only one of the recommended providers must be implemented for compliance.</note>">
 <!ENTITY nocli "<note>Not required for command line or terminal based interfaces.</note>">
+<!ENTITY pushplatform "<note>Not required on platforms which don’t tend to kill user programs.</note>">
 ]>
 <?xml-stylesheet type='text/xsl' href='xep.xsl'?>
 <xep>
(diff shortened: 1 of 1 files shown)


SamWhited:
> Add XEP-0156: Discovering Alternative XMPP Connection Methods for web clients.

Thanks; I'll definitely use this one. Although, it means that web clients which are designed to work with a specific service would not be "web compliant" (since they probably already know where there server is located and don't need to look it up). I'm not sure that it matters (as long sa they support BOSH or the XMPP websocket subprotocol instead of some custom protocol that's the important thing), but it may be worth thinking about.

> Add XEP-0071: XHTML-IM for message formatting.

No. We can argue this one on the list again if you want, but as long as I'm maintaining these I will not add this unless the whole community is against me.

> Allow web clients to use a server component connecting directly with TCP.

How does the web client connect to the server component? This isn't something I've seen before, but it does sound like it might be useful to make this optional if other web clients are doing something different (that's not just a custom protocol to connect to the server component).

> Make XEP-0357 optional on platforms where it is unnecessary

I suppose it makes sense to add a footnote here, but I think we should probably be more specific than that. Or maybe we want to just drop push alltogether? I'm not sure what the best way to handle this is. I'll give it some thought. I hate that half the suites seem to have qualifiers these days, but it's a messy business I suppose.

Thanks again for the PR, I'll get some of these cherry picked over soon.

SamWhited:
Closing this for now since I *think* I cherry picked everything I need out of it. If I missed anything, I'll get to them.

PR #418, opened 2017-02-14
XEP-0280: Add 'Usability Considerations' section

tfar:
Describe automatically marking messages as read when receiving a sent carbon copy for a conversation from another device.


tfar:
This might be more complex than initially thought. Just because another client sent a message doesn't mean that client also received the messages your client has received and is showing. There is no knowing they've read the same messages you're seeing, if they don't have carbons enabled or came online later.

ge0rg:
Thanks for adding this section. I'd suggest to write "sent" as "<sent>" or 'sent' instead, to indicate that it is a stanza element.

I'd also like to add some more text regarding notifications (if you are interacting on your "other" device, you might not want to hear this device beeping all the time). My attempt at solving this was https://github.com/pfleidi/yaxim/issues/107#issuecomment-31456140 but I could imagine further improvement by delaying notifications of carbons:

1. your client receives a carbon notification
2. it waits for e.g. 30 seconds to see if there is any incoming activity from your other client
3. if no activity is received, it notifies the user of the message

PR #420, opened 2017-02-15
XEP-0313: server MAY add the id to incoming messages

iNPUTmice:
Hi,

I want to urge everybody to get this merged. I mean no disrespect to the original author but the fact is that nothing has happened on this front for months.

Here are my arguments on why this should be merged ASAP
* There is no argument that injecting the MAM id into live messages is a good thing. The only open question is on whether to do this in the MAM namespace (which would require a NS bump) or with stanza-ids (where the general consensus is that we can stick with the `urn:xmpp:mam:1` namespace
* The suggested changes do not stand in the way of potential future changes from the author. Even if the original author decides to bump the NS later for other reason those stanza-ids don't hurt.
* The changes described here actually reflect the reality of the implementations. Both ejabberd and prosody (the internal MAM in 0.10) do inject stanza-ids
* The author doesn't seem to have general objects to the idea; If I understood him correctly these changes were more or less part of some bigger changes he had already done (But never committed!)

I also cleaned up the 'should add to carbons' phrasing and changed that to MAY as well.

### DIFF
--- a/xep-0313.xml
+++ b/xep-0313.xml
@@ -28,6 +28,14 @@
   &mwild;
   &ksmith;
   <revision>
+    <version>0.5.2</version>
+    <date>2016-10-30</date>
+    <initials>dg</initials>
+    <remark>
+      <p>Added method for server to communicate the archive id</p>
+    </remark>
+  </revision>
+  <revision>
     <version>0.5.1</version>
     <date>2016-03-07</date>
     <initials>XEP Editor (ssw)</initials>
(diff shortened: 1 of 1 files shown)


Kev:
This should be covered by Matt's impending mega-patch. I think merging this and causing conflicts with Matt's patch wouldn't be sensible.

SamWhited:
> This should be covered by Matt's impending mega-patch.

I'd like to echo what Daniel said already: It's been months, and I haven't even seen a draft of the elusive mega patch.

I understand people are busy, and I'm not blaming them for not submitting the patch, that's fine, but we shouldn't keep deferring real updates because of the theoretical possibility that someone might do some work someday on this that would conflict.

Flowdalic:
> It's been months

It has actually been more than a year since such patches where said to be delivered soon: ["I will provide feedback or patches soon"](https://github.com/xsf/xeps/pull/82#issuecomment-147111692). No disrespect towards Matthew, I know that he did for XMPP in the past and I'm grateful for that, but I believe that this hurts the XMPP ecosystem a lot.

Kev:
If Matt's able to provide his patch to me, I think I can find time to do whatever tidying up is needed (if any) and get it submitted.

iNPUTmice:
> This should be covered by Matt's impending mega-patch. I think merging this and causing conflicts with Matt's patch wouldn't be sensible.

With the information I have I'm assuming that this merge conflict will just be on a file level and not on a logic level. Based on my talks with Matt I'm assuming that his patch will have the exact same syntax on the protocol level. I think it can be expected that Matt is able to resolve a simple merge conflict on file level if the alternative is that we wait even longer for this.

I'm willing to change my mind on this if someone steps forward to say that Matt's patches are in conflict semantically or syntactically on a protocol level.

SamWhited:
Closing since I think this is covered by Matt's recent changes and IIRC @iNPUTmice said that this could be closed. Feel free to reopen if I'm mistaken.

PR #402, opened 2017-02-05
Clarify CSI and Carbons state after SM resumption

Flowdalic:

PR #422, opened 2017-02-16
Clarify CSI and Carbons state after SM resumption

Flowdalic:
And since this needs a namespace bump of carbons anyway, also remove
the <private/> element in favor of <no-copy/>.

Same as #402, but without the change to XEP-0198.

Pinging @linuxwolf and @mwild1


Flowdalic:
Split into #427 and #428

PR #425, opened 2017-02-16
Remove all (?) trailing whitespaces in XEPs and in examples

linkmauve:
I grew a bit tired of having to make one PR per XEP to fix their examples, so thanks to sed I fixed it everywhere at once.

In the last commit I also made sure CI will bark whenever someone tries to add some trailing whitespaces, but not in examples as this is quite more difficult.

Since this is an editorial change, no version has been bumped.


SamWhited:
For the purpose of having a second set of eyes on it (since the PR itself is way too huge to verify manually), can you paste the exact sed scripts you used to do this? I'm sure it's fine, simple enough change, but given how big it is we should probably try to verify that nothing got broken somehow.

Thanks! I know it doesn't _really_ matter, but this always drives me nuts… great idea to add a check for it in CI too!

linkmauve:
I used the exact sed commands I put in the commit messages, except I probably added a few manual fixes in the third commit. :x

SamWhited:
oops, didn't see them under the fold; thanks!

SamWhited:
Looks good as far as I can see; thanks again!

PR #426, opened 2017-02-16
XEP-0368: Version 0.1.2 Implement more last call comments, mixing SRV records is now SHOULD

moparisthebest:
Title says about all.


stpeter:
Looks good. :-)

PR #427, opened 2017-02-18
XEP-0352: Clarify the CSI state after SM resumption.

Flowdalic:


xep-0352.xml
```diff
@@ -24,6 +24,12 @@
   <shortname>csi</shortname>
   &mwild;
   <revision>
+	<version>0.2.1</version>
```
linkmauve: You are using tabs here and everywhere in this file, while the common indent style used by XEPs is two spaces.

Flowdalic: Thanks, fixed.

mwild1:
+1 to clarifying this.

I'm fine with the current wording, but feel some very slight change might make it clearer. It might be better to say that stream resumption "does not affect the current CSI state, which always defaults to 'active' for new streams. Clients wishing to immediately go to the inactive state should do so after stream resumption."

What do you think?

Flowdalic:
@mwild1 I've changed the wording. If you give a +1 to this, I'll merge and publish a new CSI version (with my editor hat on).

mwild1:
Go for it! :)

tfar:
+1, sounds sensible.

SamWhited:
We discussed this issue in the council meeting a bit today, and while clarification is good, I think this statement is wrong:

> After a previous stream was resumed using mechanisms like &xep0198;, the CSI state is <em>not</em> restored

The CSI state is restored, the problem is that you can't know if the CSI state was ever set or not because you don't actually know if the server saw your CSI element. If a stanza was sent after your CSI element, you can see if it was acked (in which case the server also saw your CSI element), but if the CSI element was the last thing to be sent, then you went offline and did a stream restore, you can't be sure.

iNPUTmice:
Apparently people misunderstood my veto on #429 so I'm going to be extra careful about what I say in here now. I'm putting off my council hat now because this is not something for the Council to decide but for the XEP author. I'm just going to voice my own private opinion.

I agree with @SamWhited and I'm just going to repeat in other words what @SamWhited already said. The CSI state is a property of the stream that should be restored when that stream is being resumed. This can be compared to other properties of the stream that are also being restored like subscription to Carbons, the subscription to the block list (if you retrieved the block list once you will receive subsequent updates.)
However it is in fact a problem that the client doesn't necessarily know about the CSI state because CSI is not an IQ and doesn't get acknowledged.
There are two possible solutions to that problem. a) add wording to the XEP that explains the problem and recommends clients to (re)send the desired state after resumption b) change CSI to IQ.

Which of those two solutions it is going to be it's not for me to decide. That's something the XEP author has to do.

The reason that I - and I suspect @SamWhited as well - are so keen on pointing out the fact that the CSI state is - while possibly unclear to the client - in fact clear to the server, is that we don't want the problem we have with CSI to be generalized to other stream properties like Carbons, Blocklist commands, Roster and so forth.

mwild1:
> There are two possible solutions to that problem. a) add wording to the XEP that explains the problem and recommends clients to (re)send the desired state after resumption b) change CSI to IQ.

I thought the text I proposed, and Flow incorporated, clarified that my intention is (and always was) option 'a'.

There are other properties of a stream that are not restored after XEP-0198 resumption, e.g. compression (though admittedly whether this is standardized I'm not sure). My reasoning I also laid out in a post to standards@ here: https://mail.jabber.org/pipermail/standards/2017-February/032127.html

SamWhited:
> There are other properties of a stream that are not restored after XEP-0198 resumption, e.g. compression

It's my understanding that only the inner stream is restored, and compression is a property of another stream. Maybe this is where the confusion lies or what needs to be clarified?

Flowdalic:
As far as I can tell we reached mostly consensus but have to decide if the CSI state is restored or not. If it is restored, I would change the wording to something like

> After a previous stream was resumed using mechanism like &xep0198; the CSI is state is also restored. Clients either need to track if the last CSI nonza was acknowledged (using the in-order processing guarantee § 5.1), or simply reestablish the desired CSI state after resumption.

I have no strong feelings towards one of those options, as long as we clarify this part before CSI goes to draft. Therefore I leave it up to @mwild1 as author to decide which version to choose.

Side note: As discussed with other members of the community, we may want to define what an "XMPP stream" and an "XMPP session" is. We also may want to discuss if it's sensible to introduce the notation of an "outer" and "inner" stream, and define those.

SamWhited:
@mwild1 Ping. Any opinions on that last comment?

horazont:
Switching to Needs Author and pinging @mwild1 since there appears to be missing clarity.

mwild1:
@Flowdalic @iNPUTmice @SamWhited this is still marked as 'needs author'. My opinion has been asked multiple times, and I already answered with my intention multiple times.

To clarify, here's how I view things:

**Q1: Should CSI change to <iq>?**

I feel that it shouldn't. The intention all along has been for CSI commands to be as lightweight as possible, the client should be able to fire-and-forget.

**Q2: Should the CSI state be restored after resumption?**

I feel that it shouldn't. I provided some rationale for this in my post to the mailing list (linked above). Sam provided some good arguments in the other direction. However ultimately I feel there is benefit to the state being known after resumption, as opposed to it being unknown.

---

Now, if someone does feel strongly about either of these, please say so. But from what I gather the strong feeling is that *something* should be documented, but there aren't very strong feelings about what that is.

Therefore I agree with the current proposed text, and am happy for it to be merged.

SamWhited:
> this is still marked as 'needs author'.

My apologies, I forgot about this. Updated the label.

PR #428, opened 2017-02-18
XEP-0280: Clarify the Carbons state after SM resumption.

Flowdalic:
And since this needs a namespace bump of carbons anyway, also remove
the &lt;private/&gt; element in favor of &lt;no-copy/&gt;.

### DIFF
--- a/xep-0280.xml
+++ b/xep-0280.xml
@@ -23,6 +23,7 @@
     <spec>XEP-0030</spec>
     <spec>XEP-0085</spec>
     <spec>XEP-0296</spec>
+    <spec>XEP-0334</spec>
   </dependencies>
   <supersedes>
     <spec>XEP-0259</spec>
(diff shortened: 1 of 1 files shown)


ge0rg:
Just a small note that this all-in-one commit will make #434 fail, at least because the other PR also removes the <private/> element.

Ppjet6:
What's the state of this?

ge0rg:
Looks like back in 2017, council [informally decided](https://logs.xmpp.org/council/2017-03-08#2017-03-08-70a456d72a6dc76f) that the clarification doesn't need a namespace bump.

Ppjet6:
Ok so, we can remove the namespace bump and push this if it's supposed to be in there already?

@Flowdalic could you have a look?

Kev:
I don't think the assertion that this doesn't need a bump is right. Removing <private/> is an interop issue.

Ppjet6:
Ok. If this needs another go through council I'm also fine with it. I just want to cleanup abandoned issues/PRs.

ge0rg:
Please let me clarify. The **clarification** about the state of Carbons post-resumption does not need a bump. The replacement of `private` with `no-copy` does need a bump indeed, but it was only added into the PR because of the clarification's perceived need for the bump. Therefore, this PR must not be accepted as is, but if it is reduced to just the clarification, I'm sure it can be accepted as purely editorial.

lnjX:
@Flowdalic Is the clarification on this still needed? If so could you rebase the changes, so we can get the pull-request done. :)

Kev:
No activity in four years.

PR #429, opened 2017-02-18
XEP-0198: Clarify state after SM resumption.

Flowdalic:


xep-0198.xml
```diff
@@ -29,6 +29,12 @@
   &dcridland;
   &mwild;
   <revision>
+	<version>1.5.3</version>
+	<date>2017-02-18</date>
+	<initials>fs</initials>
+	<remark><p>Add note about wheter or not state is restored on a resumed stream.</p></remark>
```
SamWhited: Nit: "wheter"

Flowdalic: Thanks, fixed.

Flowdalic:
Council Card: https://trello.com/c/IR0c5nfq

SamWhited:
This has been downvoted by the council, so I'm closing this PR. It was generally agreed in the meeting today that adding a note to 0198 is not necessary, and might require changes to other protocols or encourage new protocols to change the default behavior of stream management (which works as is). Changes to other protocols that break stream management or don't behave well with stream management might still be good though.

Thanks for the PR though!

PR #430, opened 2017-02-18
XEP-0313: Revision 6.0 update

mwild1:
Human-readable diff available here: https://matthewwild.co.uk/uploads/xep-0313-r6.diff.html


SamWhited:
@mwild1 Hi Matt, thanks for the submission. This appears to have header that doesn't match the DTD and a few other small issues. See the CI output for more info. Could you take a look? Thanks!

PR #434, opened 2017-02-22
XEP-0280: Rewrite of the 'Messages Eligible for Carbons Delivery' section

ge0rg:
Hi,

this PR combines multiple changes, again ordered by level of controversy.

* https://github.com/xsf/xeps/commit/aac8f94b0df64b250f7d5320d9f2e5e76b38d2ff Removal of the `private` element, as +1ed by @linuxwolf [on standards@](https://mail.jabber.org/pipermail/standards/2017-February/032271.html) - **this requires a version bump IMO**, as it changes the server behavior when parsing messages with only the `private` element present.
* https://github.com/xsf/xeps/commit/f0e432c15ee3b811a4065abc4baa158ba5fae0c9 Improvement of the "Messages Eligible for Carbons Delivery" section that conveys the same set of rules in less words
* https://github.com/xsf/xeps/commit/7f529e1fb8841cdc16b122bb62539f8e727650b0 addition of clear MUC / MUC-PM instructions, as [suggested on standards@](https://mail.jabber.org/pipermail/standards/2017-January/032048.html)
* https://github.com/xsf/xeps/commit/c60ec805134b6059ab46a1f75a8802b77fca5a1b minor wording correction to accomodiate the above changes
* https://github.com/xsf/xeps/commit/9c388a51c61541507c599832038b6562f3d01841 **controversial** change of the "Messages Eligible for Carbons Delivery" languge from "serves MAY do this" to "servers SHOULD do this" (based on field experience with running Carbons in the wild). The language in the old XEP is very vague and non-binding, and this patch attempts to make clearly defined rules for clear use cases, so that clients can expect consistent behavior (even though they may not rely on it).

If we are doing a version bump, we should make the most out of it (without nonzas, at least).

### DIFF
--- a/xep-0280.xml
+++ b/xep-0280.xml
@@ -268,15 +268,15 @@
 <section1 topic='Messages Eligible for Carbons Delivery' anchor='which-messages'>
   <p>The focus of this specification is instant messaging applications and so those (and only those) &MESSAGE; stanzas used for instant messaging SHOULD be delivered as Carbons.</p>
 
-  <p>Defining precisely which messages are used for instant messaging and which are not is difficult, as future specifications may add additional payloads used for, or not used for, instant messaging; as such, the rules for which messages are eligible for carbons delivery is left as an implementation detail for servers.</p>
-  <p>The following is a suggested set of rules a server MAY use, or it MAY use its own; in either case it SHOULD follow the general intent of these rules. A &MESSAGE; is eligible for carbons delivery if and only if at least one the following is true:</p>
+  <p>The following is the set of rules that a server implementation SHOULD use to determine which messages should be carbon-copied. Future specifications MAY add or override rules, though they are generally advised to use &xep0334;.</p>
+  <p>A &MESSAGE; is only eligible for carbons delivery if it does not contain a &lt;no-copy/&gt; child according to <cite>XEP-0334</cite> and if at least one the following is true:</p>
     <ul>
       <li>it is of type "chat".</li>
       <li>it is of type "normal" and contains a &lt;body&gt; element.</li>
       <li>it is of type "error" and it was sent in response to a &MESSAGE; that was eligible for carbons delivery (Note that as this would require message tracking and correlation on the server, it is unlikely to be generally appropriate for most implementations).</li>
       <li>it matches the MUC-related rules outlined below.</li>
     </ul>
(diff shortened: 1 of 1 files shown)


ge0rg:
Rendered version: https://op-co.de/tmp/xep-0280.html#which-messages

ge0rg:
This is not the final set of changes yet to make everything perfect.

I'm pondering about adding these two rules:
* carbon-copy bodyless messages containing a 0184 ack
* carbon-copy bodyless messages containing a 0333 chat marker

But maybe those two XEPs need to be fixed up to mandate type=chat?

ge0rg:
Another open issue that just came up on xsf@/gajim@ MUC is messages-to-self. I've sent a [problem statement to standards@](https://mail.jabber.org/pipermail/standards/2017-March/032421.html).

tfar:
@ge0rg Are you proposing these changes to be merged? Could you open a thread on standards ML to have them discussed please? Ta.

ge0rg:
The relevant standards@ threads are linked above, and there even was some discussion on them.

However, in the meantime it looks like the council -1ed XEP-0334 (Message Processing Hints), which was the excuse to bump the Carbons namespace. Council discussion:

https://mail.jabber.org/pipermail/standards/2017-April/032583.html and followup discussion

I'm really not sure how to move forward from here. While I don't care much about which hint is to be added into the messages, I really want to make the Carbons rule properly specified, and I'm not sure this would be possible without a bump.

SamWhited:
The council today decided to start working on a document (details to TBD) that will explore the routing rules of XMPP, where the problems are, and how they can be solved. In light of this, are you okay with closing this PR @ge0rg and waiting on the results of that document? Hopefully we'll have a better idea of what needs to be done afterwards and this discussion will be a bit easier. Naturally this PR can be reopened if we decide that its rules are the correct ones.

ge0rg:
I'd prefer to keep this open as a reminder of the unsolved problem, but whatever works for the editor team is OK with me.

horazont:
I prefer to have this closed, to be honest. I don’t like having a baseline of open PRs.

I also don’t think that the PRs in the xeps repository are necessarily a good place to keep reminders about open problems, unless a concrete solution is proposed. That solution is then either merged or rejected. I think in this case, when we agree that this is part of a broader scope and must be solved e.g. in coherence with MAM and MR2.0, this should be closed.

Feel free to reopen if you disagree about the last part.

PR #436, opened 2017-02-22
XEP-0045: Add <x/> tag to MUC-PMs

ge0rg:
To faciliate better interop with Carbons, MUC-PMs need to contain an &lt;x/&gt; tag ([rationale on standards](https://mail.jabber.org/pipermail/standards/2017-January/032048.html)).

This one is a complement to #434.

### DIFF
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -1903,6 +1903,7 @@
 
   <section2 topic='Sending a Private Message' anchor='privatemessage'>
     <p>Since each occupant has its own occupant JID, an occupant can send a "private message" to a selected occupant via the service by sending a message to the intended recipient's occupant JID. The message type SHOULD be "chat" and MUST NOT be "groupchat", but MAY be left unspecified (i.e., a normal message). This privilege is controlled by the "muc#roomconfig_allowpm" room configuration option.</p>
+    <p>To allow for proper synchronization of these messages to the user's other clients by &xep0280;, the sending client SHOULD add an &lt;x/&gt; element qualified by the 'http://jabber.org/protocol/muc#user' namespace to the message.</p>
     <example caption='Occupant Sends Private Message'><![CDATA[
 <message
     from='wiccarocks@shakespeare.lit/laptop'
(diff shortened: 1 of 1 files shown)


intosi:
The examples have unclosed x elements.

ge0rg:
@intosi good catch, thanks. Fixed the PR.

SamWhited:
Added to the council agenda. Thanks. https://trello.com/c/wF37u9DJ

iNPUTmice:
From the council chat room:

```
[06:59:09 PM] ‎Kev‎: It renders existing implementations non-compliant, and poses a potential interop problem.
‎[06:59:43 PM] ‎SamWhited‎: It is a SHOULD, not a MUST (but I'm leaning that way too)
‎[07:00:00 PM] ‎Kev‎: SamWhited: A SHOULD means you can rely on it, pretty much, because anyone not doing it made a deliberate decision.
‎[07:00:16 PM] ‎SamWhited‎: Kev: What I heard there is "you can't rely on it" :)
‎[07:00:28 PM] ‎Kev‎: SamWhited: Then it should be MAY.
‎[07:00:31 PM] ‎Kev‎: SHOULD means you can rely on it.
‎[07:00:50 PM] ‎Kev‎: At the least it needs text saying that while a service/client SHOULD attach it, receiving entities can't assume that it'll be present, because $REASONS.
‎[07:03:04 PM] ‎Kev‎: I think that an additional sentence saying something approximately "Note that this requirement was added late in MUC's life, when implementations were already stable, and thus clients MUST NOT rely on receiving this element".
‎[07:03:19 PM] ‎Kev‎: would probably be sufficient to get this text in there, while not breaking anything.
‎[07:03:20 PM] ‎SamWhited‎: yah, that seems sane
‎[07:03:27 PM] ‎daniel‎: i think everyone (including the author of the PR) will probably be fine with MAY
‎[07:03:33 PM] ‎SamWhited‎: or that
‎[07:03:54 PM] ‎Kev‎: MAY would not introduce an interop concern, so I'm far less opposed to it.
```

(shortened to be more to the point)

TLDR either change the wording to MAY or add a sentence that implementations MUST NOT rely on it.

ge0rg:
Thanks very much for the heads up. I'll prepare an update, following the second route (because otherwise, we'll never get this sorted out).

ge0rg:
I've added the following note to the XEP:

> Note: because this requirement was only added in revision 1.28 of this XEP, receiving clients MUST NOT rely on the existence of the <x/> element on private messages for proper processing.

SamWhited:
@ge0rg can I get you to add a revision block to this and then we'll merge? Thanks

ge0rg:
@SamWhited at your service :)

xep-0045.xml
```diff
@@ -46,6 +46,14 @@
   <registry/>
   &stpeter;
   <revision>
+    <version>1.28</version>
+    <date>2017-05-31</date>
+    <initials>gl</initials>
+    <remark>
+      <p>Introduce &lt;x/> tag in MUC-PMs to support better Carbon delivery.</p>
```
SamWhited: Nit: `&gt;` for consistency with the other side of the tag and other places tags are used in the XML

ge0rg:
@samwhited feel free to change, I'm away from computers for the next two weeks.

ge0rg:
Force-pushed a minor update of the revision block, to fix comment https://github.com/xsf/xeps/pull/436#pullrequestreview-41815570

horazont:
Removing "Needs Council", apparently it got accepted [(source)](https://trello.com/c/wF37u9DJ/).

PR #453, opened 2017-03-20
make: Rerun xelatex to build ToC in PDFs

SamWhited:
LaTeX sometimes requires multiple passes to build external files; make sure the ToC compiles by guessing if we need another pass based on log output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ HTMLDEPS=xep.xsl $(CSSTARGETS) $(JSTARGETS)
 CSSTARGETS=$(OUTDIR)/xmpp.css $(OUTDIR)/prettify.css
 JSTARGETS=$(OUTDIR)/prettify.js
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -79,7 +81,10 @@ $(OUTDIR)/%.pdf: %.xml $(XMLDEPS) $(TEXMLDEPS)
 	sed -i -e 's|\([\s"]\)\([^"]http://[^ "]*\)|\1\\path{\2}|g' \
 		-e 's|\\hyperref\[#\([^}]*\)\]|\\hyperref\[\1\]|g' \
 		-e 's|\\pageref{#\([^}]*\)}|\\pageref{\1}|g' "$(@:.pdf=.tex)"
-	cd $(OUTDIR); xelatex -interaction=batchmode -no-shell-escape "$(notdir $(basename $@)).tex" && echo "Finished building $@"
+	while ($(DO_XELATEX) ; \
+		grep -q "Rerun to get" $(<:.xml=.log) ) do true; \
```
Flowdalic: Couldn't this turn into an endless loop in case xelatex always emits the pattern?

SamWhited: yah, but I've never seen it do that

PR #454, opened 2017-03-20
Add ISR-SASL2 ProtoXEP to inbox/

Flowdalic:


xep.dtd
```diff
@@ -121,7 +121,7 @@ THE SOFTWARE.
   alt CDATA ''
   height CDATA ''
   width CDATA ''>
-<!ELEMENT link (#PCDATA)* >
+<!ELEMENT link (#PCDATA | cite)* >
```
SamWhited: I don't fully understand what the consequences of this are; we probably shouldn't start adding random things to link though. Remember it gets used in PDFs, HTML, etc.

Flowdalic: It's not adding random things. It allows for things like  

```
<link url='http://xmpp.org/extensions/xep-0198.html#resumption'>
  <cite>XEP-0198</cite>§ 5
</link>
```

I've successfully generated HTML and PDF versions using this.

PR #458, opened 2017-03-24
Publish the OMEMO bundles in several items in one unique PEP node

edhelas:
Based on discussions made during the FOSDEM Summit

### DIFF
--- a/xep-0384.xml
+++ b/xep-0384.xml
@@ -139,66 +139,50 @@
     </p>
   </section2>
   <section2 topic='Discovering peer support' anchor='usecases-discovering'>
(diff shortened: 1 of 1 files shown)


SamWhited:
author comms sent.

remko:
This issue can be closed per https://mail.jabber.org/pipermail/standards/2017-April/032576.html

The summary of the discussion of that thread is that, although this sounds like a valid idea, PEP does not guarantee support for multiple items on a single node.

SamWhited:
Closing per the thread that @remko pointed out. Thanks for the change though!

PR #460, opened 2017-03-29
XEP-0384: Introduce ODR, improve specification

strb:
This new version of the specification introduces the OMEMO Double
Ratchet. We now use our own, new wire format, making the protocol
independent from third parties.

This revision also improves the general quality of the specification
by describing behaviors in more detail and using terminology more
consistently.

PR #463, opened 2017-05-06
OMEMO (XEP-0384) v0.2: Misc. Improvements

remko:
Changes to make the XEP consistent with Olm.
These mostly reflect the outcome of the discussion in thread https://mail.jabber.org/pipermail/standards/2017-March/032522.html

There are also extra things in here, raised by other people, or by myself.

- Add fingerprint key
- Sign identity keys
- Use Olm authenticated encryption for encrypting payloads
- Made bundle elements kebab-case
- Removed pre-keys from bundle that are not used by Olm
- Removed references to the signal protocol and library
- Removed iv element
- Don't mandate device ID format, suggest GUID
- Add JID tag to keys ensure uniqueness ID
- Add recipient and sender JID to encrypted keys to ensure message authenticity
- Add max_items attribute to bundle query
- Add use case for cross-signing identity keys

### DIFF
--- a/xep-0384.xml
+++ b/xep-0384.xml
@@ -28,6 +28,27 @@
     <jid>andy@strb.org</jid>
   </author>
   <revision>
+    <version>0.2</version>
+    <date>2017-05-05</date>
+    <initials>rt</initials>
+    <remark>
+      <ul>
+        <li>Add fingerprint key</li>
+        <li>Sign identity keys</li>
+        <li>Use Olm authenticated encryption for encrypting body payload</li>
+        <li>Made bundle elements kebab-case</li>
+        <li>Removed pre-keys from bundle that are not used by Olm</li>
+        <li>Removed references to the signal protocol and library</li>
+        <li>Removed IV element from header</li>
+        <li>Don't mandate device ID format, suggest GUID</li>
+        <li>Add JID tag to keys ensure uniqueness ID</li>
+        <li>Add recipient and sender JID to encrypted keys to ensure message authenticity</li>
+        <li>Add max_items attribute to bundle query</li>
+        <li>Add use case for cross-signing identity keys</li>
+      </ul>
+    </remark>
+  </revision>
+  <revision>
     <version>0.1</version>
     <date>2016-12-07</date>
     <initials>XEP Editor: ssw</initials>
@@ -66,8 +87,9 @@
       This XEP defines a protocol that leverages &olm; encryption to provide
       multi-end to multi-end encryption, allowing messages to be synchronized
       securely across multiple clients, even if some of them are offline. Olm
-      is a cryptographic double ratched protocol based on work by Trevor Perrin
-      and Moxie Marlinspike first published as the Axolotl protocol.
+      is an implementation of the cryptographic 
+      <span class='ref'><link url='https://whispersystems.org/docs/specifications/doubleratchet/'>Double Ratchet Algorithm</link></span> <note>The Double Ratchet Algorithm &lt;<link url='https://whispersystems.org/docs/specifications/doubleratchet/'>https://whispersystems.org/docs/specifications/doubleratchet/</link>&gt;.</note>.
+      Olm has been publicly reviewed by the NCC Group &olm-review;.
     </p>
   </section2>
   <section2 topic='Overview' anchor='intro-overview'>
@@ -133,9 +155,19 @@
   <section2 topic='Setup' anchor='usecases-setup'>
     <p>
       The first thing that needs to happen if a client wants to start using
-      OMEMO is they need to generate an IdentityKey and a Device ID. The
-      IdentityKey is a &curve25519; public/private Key pair. The Device ID is a
-      randomly generated integer between 1 and 2^31 - 1.
+      OMEMO is they need to generate a <em>Device ID</em>, 
+      a <em>Fingerprint Key Pair</em>, and  an <em>Identity Key Pair</em>.
+    </p>
+    <p>
+      The <em>Device ID</em> is a randomly generated identifier, and MUST be
+      unique for all clients for the user's JID. The identifier SHOULD be
+      a GUID to ensure global uniqueness.
+    </p>
+    <p>
+      The <em>Fingerprint Key Pair</em> is an &ed25519; public/private key pair for signing. The public part of the signing key is used as a fingerprint of the device.
+    </p>
+    <p>
+      The <em>Identity Key Pair</em> consists of a &curve25519; public/private key pair. This key should be refreshed periodically.
     </p>
   </section2>
   <section2 topic='Discovering peer support' anchor='usecases-discovering'>
@@ -149,8 +181,8 @@
     <items node='urn:xmpp:omemo:0:devicelist'>
       <item>
         <list xmlns='urn:xmpp:omemo:0'>
-          <device id='12345' />
-          <device id='4223' />
+          <device id='a74e10af-939d-4d90-b486-9a43ad39119b' />
+          <device id='75d22a2d-4eb4-47c7-b5f9-c454941bd36d' />
         </list>
       </item>
     </items>
@@ -165,43 +197,40 @@
     <publish node='urn:xmpp:omemo:0:devicelist'>
       <item>
         <list xmlns='urn:xmpp:omemo:0'>
-          <device id='12345' />
-          <device id='4223' />
-          <device id='31415' />
+          <device id='a74e10af-939d-4d90-b486-9a43ad39119b' />
+          <device id='75d22a2d-4eb4-47c7-b5f9-c454941bd36d' />
+          <device id='22f242ab-73ad-4071-aa9b-a0f8e8e5b598' />
         </list>
       </item>
     </publish>
   </pubsub>
 </iq>]]></example>
     <p>This step presents the risk of introducing a race condition: Two devices might simultaneously try to announce themselves, unaware of the other's existence. The second device would overwrite the first one. To mitigate this, devices MUST check that their own device ID is contained in the list whenever they receive a PEP update from their own account. If they have been removed, they MUST reannounce themselves.</p>
-    <p>Furthermore, a device MUST announce it's IdentityKey, a signed PreKey, and a list of PreKeys in a separate, per-device PEP node. The list SHOULD contain 100 PreKeys, but MUST contain no less than 20.</p>
+
+    <p>Furthermore, a device MUST announce the public parts of its fingerprint and identity key pair, and a list of PreKeys in a separate, per-device PEP node. The public identity key MUST be signed with the fingerprint key. 
+      The list of PreKeys SHOULD contain 100 PreKeys signed with the fingerprint key. The list of PreKeys MUST contain no less than 20 PreKeys.
+      The client SHOULD generate a new signed identity key periodically (e.g. once a month).
+    </p>
+
     <example caption='Announcing bundle information'><![CDATA[
 <iq from='juliet@capulet.lit' type='set' id='announce2'>
   <pubsub xmlns='http://jabber.org/protocol/pubsub'>
-    <publish node='urn:xmpp:omemo:0:bundles:31415'>
+    <publish node='urn:xmpp:omemo:0:bundles:22f242ab-73ad-4071-aa9b-a0f8e8e5b598'>
       <item>
         <bundle xmlns='urn:xmpp:omemo:0'>
-          <signedPreKeyPublic signedPreKeyId='1'>
-            BASE64ENCODED...
-          </signedPreKeyPublic>
-          <signedPreKeySignature>
-            BASE64ENCODED...
-          </signedPreKeySignature>
-          <identityKey>
-            BASE64ENCODED...
-          </identityKey>
-          <prekeys>
-            <preKeyPublic preKeyId='1'>
-              BASE64ENCODED...
-            </preKeyPublic>
-            <preKeyPublic preKeyId='2'>
-              BASE64ENCODED...
-            </preKeyPublic>
-            <preKeyPublic preKeyId='3'>
-              BASE64ENCODED...
-            </preKeyPublic>
-            <!-- ... -->
-          </prekeys>
+          <fingerprint-key key="9EA1q2XYopPW9SyzKCkLYJ6XVT3MaF8sk75C4efi+e0=" />
+          <identity-key key="S8NwiEt6yh8KKHzLYPxUhol1yN8RdfynqH2CsvO4fmk=">
+            <signature>XhZtkP0STMZ1fXEVYz1aUkdt04zIVsihFkawLE+aHZWFCn
+XeGn2ZThTU5TQm0pJo3oFWd7JED0aCEge0ry3hCg==</signature>
+          </identity-key>
+          <pre-key key="6xY6aSIG+AvW+jfQGgf+bUD0nPKY8NNUqIm4e2oWQ0E=">
+            <signature>7uBP6mIXBdKzkxKOQcIpDNOQe+SX8dWl74h1KexH6bQ3UIQN7TN37I+SgxeEChXecjoBsPa62p5W4I2S9GXKJQ==</signature>
+</signature>
+          </pre-key>
+          <pre-key key="KRMGUwvmL4ci5ySfPSXO4B3ZQQgtdxU6YoAzYXBsqAk=">
+            <signature>s702IcFuxdiQwfiUn+x8b2iu/OaZDYFeIAlUdtmcmo7plwLm
+p/1QwaO7sta8LHocbtoU/b0VEJDkaapUZ65eDw==</signature>
+          </pre-key>
         </bundle>
       </item>
     </publish>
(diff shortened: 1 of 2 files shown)


SamWhited:
Hi @remko et al. I just wanted to apologize for the delay on processing PRs. We've had no editor infrastructure, but it should be up and running again "soon" and then we'll confer with the author and merge this change if they agree.

Thanks for your patience.

iNPUTmice:
Not the author here but

> These mostly reflect the outcome of the discussion

seems a bit like an overstatement.

We probably want to go with #460 instead. It's not completely done yet but is the output of several discussions within the OMEMO community. (To be clear it is completely the work of @strb but it has been reviewed by several OMEMO developers who also helped to improve wordings. See the extended acknowledgment section in that PR.)

To summarize the arguments for #460 and against this PR.
1) It's gonna be an easier upgrade path for all the existing implementations.
2) What's described in that PR is entirely public domain (No dependency on GPL only code).
3) The one missing implementation of the key convert method is not a strong enough argument to ignore point 1 and 2.

remko:
Thanks, it's nice to finally get some feedback on this.

- The unavailability of crypto primitives in crypto libraries is IMO not a concern to be taken so lightly
- This protocol is still in experimental stage. Easy upgrade paths from previous versions of this XEP (and especially protocols predating the XEP) are nice to have, but not nice enough to make concessions already.
- Moving away from Olm to a custom ratchet means dropping all audited code
- There were other concerns by the Olm team (who are more knowledgeable than me), such as the overengineered key-sharing of the protocol (AFAIK, introduced as a backwards-compatible measure for Signal)
- This proposal enables deniability, which X3DH doesn't. It can be discussed whether this is wanted or not.

vanitasvitae:
I have a question regarding the use of AES-CBC. As far as I understand (I'm not an expert), AES CBC is (contrary to AES GCM) not authenticated. Wouldn't the use of AES-CBC open up the vulnerability mentioned in the [OMEMO audit](https://conversations.im/omemo/audit.pdf)  (possibility of MitM by replacing encrypted payload via a compromised device) again? Currently the authentication tag of the payload is appended to the messageKey to prevent that. Is that also possible with AES-CBC?

remko:
@vanitasvitae It's a good question. You need to combine AES-CBC with HMAC-SHA to get authentication (which is what Olm does for the message encryption). However, given that Olm authenticates the message (i.e. the CBC key), I'm assuming (no crypto expert here) that authentication of the CBC key doesn't bring anything.

iNPUTmice:
@remko You'd have to authenticate the message text. Not not the key. Otherwise anyone with the key can replace the message.

vanitasvitae:
Thats because the messageKey is the same for every recipient.

remko:
I see, point taken.

iNPUTmice:
FWIW #460 also moves from AES-GCM to AES-CBC-HMAC-SHA.

remko:
Updated the PR, addressed authentication attacks (as specified in the Olm spec), and added some extra things that were raised elsewhere.

remko:
Updated the PR. Now using Olm authenticated encryption for encrypting body payloads.

remko:
Updated the PR: Added a cross-signing use case, which can improve usability in a TOFU model when a new client is used to talk to existing contacts.

vanitasvitae:
Nice to see the idea of identityKey cross signing getting adopted :)

gkdr:
> * Don't mandate device ID format, suggest GUID
> * Add JID tag to keys ensure uniqueness ID

A device ID should be unique per device, not unique globally. Adding the JID should be more than enough on its own. I think this is unnecessary altogether because a collision is very unlikely even as it is and it is easy enough to deal with it.
Another point to consider is that the Signal implementation uses ints internally, so there would be a lot of patching necessary.

> * Split identity key into encryption and signature key
> * Sign identity encryption keys

The identity key is a (EC) Diffie-Hellman key, __not__ an encryption key. This is not PGP, and the incorrect naming should be changed.
If you don't want to call it DH key and encryption key, I think it's worth considering concatenating the two and still call it identity key.

> * Use Olm authenticated encryption for encrypting payloads
> * Moving away from Olm to a custom ratchet means dropping all audited code


Maybe someone can tell me I'm absolutely wrong, but I don't agree. Changing a library to expose internal values while not touching the sensitive crypto parts should not matter, or else any client using a crypto lib is automatically unusable.
I believe that using an implementation-independent algorithm is a good step, and depending on Olm instead of Signal does not do that.
I'm not sure where the improvement is to use "Olm authenticated encryption" over "AES-256-CBC with HMAC-SHA-256" which is essentially is according to the docs.

> * This proposal enables deniability, which X3DH doesn't. It can be discussed whether this is wanted or not.
> * Allow optional signatures on pre-keys

I think you got me wrong on the mailing list.
X3DH only does not provide "full deniability", i.e. you cannot deny having ever written the message.
It is "peer-and-time deniable" though -- you can still deny the intended recipient and the timeframe of sending.
You can read up on this in: Cas Cremers and Michele Feltz. One-Round Strongly Secure Key Exchange with Perfect Forward Secrecy and Deniability. https://eprint.iacr.org/2011/300


In general, deniability is a rather philosophical attribute - either you claim it was someone else in your household or even malware writing the message instead of yourself to get insta-deniability, or the judge will accept the plaintext message surrounding the ciphertext as proof anyway. It would surprise me if you can find a case where "deniability" saved someone.


That leads to the next point - the weaknesses introduced by not using a signed pre key AND a prekey at the same time.
The first thing is: it enables a replay attack until the encryption key changes. Maybe inconsequential for the instant messaging use case, but rather ugly to be allowed by design.
But the more important thing is the following attack, as I already explained on the mailing list, and I'll just quote myself here: "If Eve replaces Alice's ephemeral public keys with her own when they are requested by Bob, she can decrypt messages sent from Bob to Alice in case the learns Alice's long-term private key at any later point in time."
It breaks forward secrecy, which I strongly believe is __not__ acceptable design.
I'm sure OWS did a great deal of thinking and weighing the tradeoffs against each other and I think they arrived at the best conclusion.

To avoid confusion, since the X3DH specification does mention using XEdDSA, I propose a "cached, signed 3DH" (__CS3DH__) for OMEMO which works the same except that the signature on the signed pre key is from a separate Ed25519 signing key (which is already included in your proposal) and not derived from the Curve25519 key using the XEdDSA scheme.

> * The unavailability of crypto primitives in crypto libraries is IMO not a concern to be taken so lightly

It should also not be taken lightly that a handful of applications already implemented it the way it was specified. Generalization will also take some work but is a lasting step (and if done right might not require a lot of changes to the already existing code, just additional calls to expose more values), switching implementations on the other hand is not.

vanitasvitae:
> A device ID should be unique per device, not unique globally. 
Unique per Jid, right?

gkdr:
@vanitasvitae Yes, thanks. I'll correct it.

remko:
@gkdr Thanks for the comments.

> Adding the JID should be more than enough on its own.

My suggestion to move to GUIDs wasn't that much to avoid collisions. Requiring IDs to be 32-bit integers feels like an arbitrary restriction that doesn't belong in a protocol IMO, unless there's a good reason. (and that this is because the underlying library requires this is a bad smell). I don't think XMPP puts any restriction on IDs anywhere else.

> Another point to consider is that the Signal implementation uses ints internally, so there would be a lot of patching necessary.

This PR moves the XEP back into a direction where using the Signal implementation wouldn't make much sense I believe, so I don't think that's a concern. If the integers are the only thing keeping protocol implementations to use Signal, then this sounds like a tradeoff that can be made, but I doubt this will be the case.

> The identity key is a (EC) Diffie-Hellman key, not an encryption key.

I don't think this PR says that. Ed25519 keys are used as identity keys (as is done in Olm), and are used to sign the EC keys (as is done in Olm). 

> I think you got me wrong on the mailing list.

The bit on reducing deniability was actually taken from https://matrix.org/git/olm/about/docs/signing.rst

> I'm not sure where the improvement is to use "Olm authenticated encryption" over "AES-256-CBC with HMAC-SHA-256" 

There isn't. The current version of the spec says "AES-128-GCM", which is what this change addresses. Also, saying it's "Olm authenticated encryption" avoids the need to re-specify how exactly the keys are derived from the generated key.

> It should also not be taken lightly that a handful of applications already implemented it the way it was specified. 

Per XEP-0001, this shouldn't be a concern for Experimental XEPs.
Also, I don't think we have much choice here. If the decision is made to not break current clients, moving to Olm would not be possible, in which case (as far as i can tell) you'll have a hard time convincing the XSF to go forward with a XEP (as I understand, Olm was a precondition of accepting the initial version). Of course, that's just my interpretation of things, don't take my word for that.

gkdr:
> My suggestion to move to GUIDs wasn't that much to avoid collisions. Requiring IDs to be 32-bit integers feels like an arbitrary restriction that doesn't belong in a protocol IMO, unless there's a good reason.

I see. Maybe this isn't a valid concern but it seems to me like the message doesn't need to be so big. A GUID is long enough by itself, and then the JID is also included... Is there a reason not to start mandating an ID format?
I agree that the internals of a specific implementation are not a good argument, but the point I was trying to make is that all current implementations already use it and aside from patching the lib it is probable that it makes database migration and a lot of code and interface changes necessary (in statically typed languages at least). It is just a lot of work to fix something that is not broken.



>I don't think this PR says that.

The point is that a Curve25519 key is not an encryption key. But this is said in several places, starting from the summary of changes you posted in the OP.
I'll quote all the mentions of the Curve25119 key as encryption key I could find:

>> * [Split identity key into encryption and signature key](https://github.com/xsf/xeps/pull/463/files#diff-04135463724e725e57f2981e3bb7c544R36)
>> * [The <em>Identity Key Pair</em> consists of a &curve25519; public/private key pair for encryption](https://github.com/xsf/xeps/pull/463/files#diff-04135463724e725e57f2981e3bb7c544R168)
>> * [The public encryption identity key MUST be signed with the signing identity key.](https://github.com/xsf/xeps/pull/463/files#diff-04135463724e725e57f2981e3bb7c544R210)
>> * [The signature of the encryption identity key is verified, a random public pre-key is selected (and its signature verified if it is signed). The selected pre-key and encryption identity key is used to build an Olm session.](https://github.com/xsf/xeps/pull/463/files#diff-04135463724e725e57f2981e3bb7c544R249)


>The bit on reducing deniability was actually taken from https://matrix.org/git/olm/about/docs/signing.rst

I probably got you wrong then, since you said "enabling deniability" (which to me implied that it did not exist before).
I want to stress here that the main point of my comment was the part about the broken forward secrecy, which is why I think neither signed prekeys nor one-time prekeys should be optional.


>Per XEP-0001, this shouldn't be a concern for Experimental XEPs.

Can you point out to me where it says that?

>Also, I don't think we have much choice here. If the decision is made to not break current clients, >moving to Olm would not be possible, in which case (as far as i can tell) you'll have a hard time >convincing the XSF to go forward with a XEP (as I understand, Olm was a precondition of accepting >the initial version). Of course, that's just my interpretation of things, don't take my word for that.

Don't get me wrong, my concern is not so much about not breaking clients. The initial proposal with ODR would break them too. There just has to be a good reason to break things.
I can tell you I won't change my work just to be XEP-compliant, i.e. just because the XSF decided they like one implementation of an open standard better than the other.
What I do see as a good reason though is adopting a more general double ratchet format to allow use of any double ratchet implementation. I can keep using mine, others can use a different one, new implementations can be added (e.g. the one provided by Wire, because the client should be written in Rust). That was the whole intention behind the ODR and I think it captures the spirit of the third objective mentioned in section 2 of XEP-0001 the best: "To ensure interoperability among the disparate technologies used on XMPP networks."

To summarize, and emphasize again, ODR was meant to enable implementers to move away from a specific double ratchet implementation.
I suspect that the mention of OWS' reference implementation in the XEP made the wrong impression about the intentions, however it is simply a fact that all current implementations use it which is a good reason to use it as an example.

remko:
> The point is that a Curve25519 key is not an encryption key. 

I see what you mean. My terminology is indeed wrong. Will go back and address that.

> I think neither signed prekeys nor one-time prekeys should be optional.

That's a clear statement, thanks. I'll go back to read your links.

> What I do see as a good reason though is adopting a more general double ratchet format 

I don't really get that reasoning. By introducing a custom ratchet, you remove the option from clients to use a verified library for their double ratchet. If the XEP mandates Olm, you can still use your own custom ratchet implementation if you want to, you just use different parameters than with ODR?

remko:
@gkdr Updated PR to remove the optionality of signatures on prekeys, and changed the ed/ec key terminology.

gkdr:
>I don't really get that reasoning. By introducing a custom ratchet, you remove the option from clients to use a verified library for their double ratchet. 

ODR is not supposed to be a custom ratchet. It is for moving the "wire format" from some implementation-specific serialized binary data to XMPP.
For example this concerns N, the number of the message in the current sending chain, and PN, the number of messages in the previous chain, needed for the case that messages arrive out of order (see [section 2.6 of the double ratchet specification](https://whispersystems.org/docs/specifications/doubleratchet/) for an example).
For pre key messages, it is other simple data like the IDs of the used pre keys.
I do think things like these do not need to be transported as serialized binary data.

IIRC, the only thing that actually changes is the "info string" that goes into the derivation functions. 
Other than that, the ODR just defines which of the suggestions the double ratchet specification makes are to be used.
As far as I can tell from the [Olm spec](https://matrix.org/git/olm/about/docs/olm.rst), both libsignal-protocol-* and libolm use the same crypto primitives and constants (aside from the info strings), so they would be compatible to each other.
Therefore, the only changes needed do not touch the crypto code, but simply expose these internal IDs, keys and counters from the opaque structs so they can be sent over XMPP (which in the easiest case is just making the struct non-opaque by copypasting its definition to a location that is visible, but I didn't look to much into it yet).
I do not believe this step un-audits a lib.

>If the XEP mandates Olm, you can still use your own custom ratchet implementation if you want to, you just use different parameters than with ODR?

Sorry, I don't understand this question, but I suspect I might have answered it already. If not, let me know.

Thanks for considering the changes I suggested, I'll look at the details some other time.

remko:
> I do think things like these do not need to be transported as serialized binary data.

AFAIK, these need to be authenticated, so they need to be serialized first. This is why it doesn't make sense to 'move this to XMPP', and this is why you need more than just the specification of the ratchet.

> Sorry, I don't understand this question

What I meant was this:

> IIRC, the only thing that actually changes is the "info string" that goes into the derivation functions.

If that's the only thing that changes, then why not just use the Olm info strings? That way, there's a reference implementation, together with a fully documented protocol (including choice of primitives, message format, key agreement protocol, implementation notes, ...). 

> Thanks for considering the changes I suggested, I'll look at the details some other time.

Great, thanks!

vanitasvitae:
>If that's the only thing that changes, then why not just use the Olm info strings? That way, there's a reference implementation, together with a fully documented protocol (including choice of primitives, message format, key agreement protocol, implementation notes, ...).

Same is true for libsignal at the moment. The only difference is the license.

remko:
> Same is true for libsignal at the moment. The only difference is the license.

Well, the license is a serious issue (much bigger than just replacing strings). But, as I said before, that wouldn't be an issue if it would be easy to create independent implementations, which it isn't because of X3DH. Other people are concerned about the legal issues surrounding uses of Signal. And there's the message format that is undocumented AFAIK. So there's more than 'just the license' going on here.

vanitasvitae:
Sorry, what I wanted to say is that I'd prefere a library independent specification, so that one could patch both libsignal and olm the same way.

gkdr:
>AFAIK, these need to be authenticated, so they need to be serialized first. This is why it doesn't make sense to 'move this to XMPP', and this is why you need more than just the specification of the ratchet.

You are right, at the moment authentication is missing, but that is a different issue. My criticism is more about the "binary" than the "serialized".
A simple solution to serialization is string concatenation, over which a MAC can be computed, but I'll have to think some more about that. (Actually putting the data into an element's attributes probably counts as serialization as well.)

>If that's the only thing that changes, then why not just use the Olm info strings?

The library still needs to be changed to expose internal values, so doing this little change as well seems cleaner.

remko:
> A simple solution to serialization is string concatenation

If it can represent the data unambiguously, I guess it's a little bit simpler (less flexible, not that it matters that much). But why reinvent the wheel, especially if it means you can no longer use the libraries that are already there?

> (Actually putting the data into an element's attributes probably counts as serialization as well.)

Using XML as serialization format for authentication tends to bring in annoying things like canonicalization AFAIK.

> The library still needs to be changed to expose internal values

Does it? Which internal values? I haven't looked into the details, but I was under the impression that libolm's goal was to be used as-is, so I'll bring this up with the authors.

manchito:
> Other people are concerned about the legal issues surrounding uses of Signal.
Who or where are they? Which clients are they implementing OMEMO for? AFAIK those who have actually written working OMEMO implementations (myself included) haven't found any such issues. Yes, libsignal is GPL licensed, but the algorithms it uses have public specifications and the lib itself has also been audited/reviewed [1](https://conversations.im/omemo/audit.pdf) [2](https://eprint.iacr.org/2014/904.pdf) [3](https://eprint.iacr.org/2016/1013.pdf). As I understand it, the purpose of ODR is to allow us to keep those well tested libraries, be it Olm or Signal and only touch their higher ends without severely modifying the essentials.

gkdr:
>Using XML as serialization format for authentication tends to bring in annoying things like canonicalization AFAIK.

I was just commenting on the word "serialization" and meant that as it is, it is already serialized. You can ignore that sentence :D

> But why reinvent the wheel, especially if it means you can no longer use the libraries that are already there?

But it's not "libraries", it's only libolm. You could ask just as well: Why make every existing implementation switch the library if the specification to all parts is open?

>Does it? Which internal values? I haven't looked into the details, but I was under the impression that libolm's goal was to be used as-is, so I'll bring this up with the authors.

Sorry, I meant "for use with ODR".

>Again, more to explicitly re-specify, just because we don't want to use something existing. That feels like NIH to me.

I don't quite understand how that is different from saying that we want to just keep using libsignal, after all it not only exists, but on top of that is used by all of the 5 or so implementations of OMEMO.

We're trying to do it the "clean" and right way, not the easy way, so existing implementations shouldn't be an argument, I think.
And again, the point of ODR is to make every existing implementation usable with a little change. Maybe it will turn out that it's too much effort after all (which is what experimental XEPs are for), but in my opinion it's a good goal.

remko:
@gkdr So, it seems our biggest disagreement right now (pending your further review) is the wire format and info strings. Maybe we can try to get to an agreement on the rest of this PR, and commit it as an intermediate version so someone can suggest the ODR changes on top of it. This would make the current XEP version consistent in its current philosophy, and move the discussion away from dealbreakers for me. It would make further discussions easier for everyone I think.

strb:
I'm just going to leave the philosophical debate for the mailing list, and focus on giving some concrete feedback. 

I can still see a few other issues here:
* The naming of "fingerprint key" and "identity key" is very confusing to me. I don't understand what a "fingerprint key" is supposed to be. Any public key has a fingerprint, so without knowing how exactly this key is used, it's impossible to tell it's purpose from this name. More importantly though, calling a Diffie-Hellman key an "identity key" is incredibly misleading, as it is nothing of the sort. You can't use a key exchange to prove your identity, the opposite is true. You need to separately authenticate such a key exchange because it does **not** provide authentication. Your "identity key" is essentially just a long-term signed pre-key. If anything, what you call "fingerprint key" should be called "identity key", as it is actually what you are using to prove your identity.
* This is more of a nitpick, but the wording in the "sending a message" use-case is a bit confusing in that it suggests that an authenticated encryption mechanism is used to generate keys. You probably mean that you're using the same key derivation mechanism as the olm authenticated encryption, but in that case it might be clearer to phrase it as "applying HKDF as in olm authenticated encryption" or something like that.
* I suppose going with GUIDs makes sense if you're abandoning signal-protocol, but then why would you tag the envelopes with the JID as well? Are GUIDs really not unique enough by themselves? This feels like unnecessary bloat.
* Cross-signing as a concept is a UX nightmare, and if you want this to work at all, you need to go into significantly more detail.
    * Are you going to apply this transitively? Can I sign keys of another account? How are you going to handle revocation? There is a reason why nobody understands how to use PGP, and you're basically going to be stuck reinventing it here at this point. 
    * Furthermore, how do these signatures even get published? Does the signing device somehow communicate the signature to the signee device, which then publishes it? If so, how does this work? Or does the signing device directly manipulate the signee's bundle? This would simply get overwritten by the signee device routinely re-publishing their bundle.

Mostly though, I am a bit confused by the inclusion of these random tweaks that don't feel thought all the way through. If this revision is simply about fixing up some OLM-related inconsistencies, this stuff doesn't belong in here. If it's intended as a more general update to the specification, then there are much more important pain points to address than inane details like "int vs GUID".

remko:
> The naming of "fingerprint key" and "identity key" is very confusing to me.

"Identity key" is the same as it has always been, and comes from the Double Ratchet spec. The fingerprint key is [terminology coming from Matrix/Olm](https://matrix.org/docs/guides/e2e_implementation.html#keys-used-in-end-to-end-encryption
), and is used to authenticate the identity key and pre-keys. Libsignal doesn't have the concept of a fingerprint key: they use the identity key for both authentication and key agreement (which introduces the need for XEdDSA).

> the wording in the "sending a message" use-case is a bit confusing

If it's confusing, it can certainly be clarified.

> Are GUIDs really not unique enough by themselves

GUIDs are only a suggestion; clients are free to use their own ID generation mechanism (as is the case everywhere else), but they need to make sure it's unique for a specific JID. That's why you need the JID.

> Cross-signing as a concept is a UX nightmare

It's a nightmare if you do it transitively, yes. This is not what I suggest, I suggest signing of keys within the same JID. This also avoids the questions on how to communicate signatures, as you can just publish the signature to PEP of your JID. This cross-signing of *own devices* is actually there to make UX still acceptable without sacrificing security when a new device is used. But it's something that can still be debated.

> Mostly though, I am a bit confused by the inclusion of these random tweaks

This is a fair point. If it makes agreeing on the other parts possible, I can take these tweaks out.

manchito:
> This is a fair point. If it makes agreeing on the other parts possible, I can take these tweaks out.

Cross-signing is a very interesting topic as well as complex. I personally think your idea goes in the right direction, but I agree it is not the right place/timing for that. It would be better IMO to treat that specific "tweak" separately, as well as other big ones which are surely still to come like standardising multi-party behaviour or HTTP File Upload.

vanitasvitae:
I also have the feeling cross signing should not be part of the main spec, but rather go into a seperate addon xep?

strb:
> "Identity key" is the same as it has always been, and comes from the Double Ratchet spec.

This does not follow. DR uses their key for two things (signing and key exchange), and names them after one of the two functions (namely signing, *not* key exchange). You're now using the same name for a separate key exchange key. It is most definitely not the same as it has always been, you can't just focus on the format of the key while completely ignoring how it's used.

> The fingerprint key is terminology coming from Matrix/Olm, and is used to authenticate the identity key and pre-keys. 

While it makes sense to reuse their terminology, theirs is terribly misleading and should probably be changed as well. Key exchange is completely orthogonal to authentication.

> Libsignal doesn't have the concept of a fingerprint key: they use the identity key for both authentication and key agreement (which introduces the need for XEdDSA).

Yes, and they call it IdentityKey because of the former function, not the latter.

remko:
@vanitasvitae 

> I also have the feeling cross signing should not be part of the main spec, but rather go into a seperate addon xep?

Since it's purely optional on both sides, I don't think it does. But again, this cross-signing is controversial enough to warrant more discussion, so I can take that out.

remko:
@strb 

> While it makes sense to reuse their terminology, theirs is terribly misleading and should probably be changed as well. 

Changing names is not a problem. Suggestions welcome.

remko:
FYI, I'm waiting for decisions from council on how to proceed with OMEMO to see whether I want to revive this patch on top of the current spec or close it.

iNPUTmice:
Council is currently not involved with this. As far as council is concerned Andy wants to remain author of the XEP.

remko:
@iNPUTmice Good to know, I'll resubmit my proposals then.

horazont:
@remko Can this PR be closed then until you resubmit?

horazont:
I am closing this for inactivity. Feel free to reopen or create a new PR with the updated changes.

PR #464, opened 2017-05-11
Add Dockerfile serving HTML extensions

Kev:
Test-Information:
Runs, http://localhost:3080/extensions/xep-0001.html serves styled
content.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM debian:8
+MAINTAINER XSF Editors <editor@xmpp.org>
+
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN apt-get update && apt-get -y dist-upgrade
+RUN apt-get install -y xsltproc libxml2-utils libxml2 texlive fonts-inconsolata make nginx
+
+RUN mkdir /src
+COPY *.xml xep.* *.css *.xsl /src/
+COPY Makefile *.js  /src/
+WORKDIR /src
+RUN mkdir /var/www/html/extensions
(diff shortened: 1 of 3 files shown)


guusdk:
Excellent! Maybe you should add a short reference to this in the README.md file of this repository. That'll help new people find this.

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN apt-get install -y xsltproc libxml2-utils libxml2 texlive fonts-inconsolata make nginx
+
+RUN mkdir /src
+COPY *.xml xep.* *.css *.xsl *.js *.xsl Makefile /src/
+RUN mkdir /src/resources
+COPY resources/*.pdf /src/resources/
+WORKDIR /src
+RUN apt-get install -y curl python python-pip texlive-xetex 
+RUN curl https://pilotfiber.dl.sourceforge.net/project/getfo/texml/texml-2.0.2/texml-2.0.2.tar.gz -o texml-2.0.2.tar.gz && tar -xf texml-2.0.2.tar.gz && pip install texml-2.0.2/ && rm -rf texml-2.0.2
+RUN apt-get install -y texlive-fonts-recommended texlive-fonts-extra
```
horazont: Why three separate ``apt-get install`` calls? As far as I understand Docker, wouldn’t it make sense to consolidate them to build an intermediate image, reducing build time?

Same holds for the curl thing.

Kev: Because I was doing this incrementally and didn't want to blow my cache away while I was tweaking. We can easily consolidate once everything's stable, but while still fiddling with this the cache is tremendously helpful.

PR #469, opened 2017-05-30
XEP-0319: Add a precision about the XEP-0082 profile used.

linkmauve:


xep-0319.xml
```diff
@@ -25,6 +25,12 @@
   <shortname>idle</shortname>
   &tobias;
   <revision>
+    <version>1.0.1</version>
+    <date>2017-05-30</date>
+    <initials>egp</initials>
+    <remark><p>Add a precision about the &xep0082; profile used.</p></remark>
```
SamWhited: Nit: "Be precise about the XEP-0082 profile used" sounds a bit better; precision is not plural/singular, so you can't have "a precision" (I'm not sure what that rule is actually called). I wouldn't add a link in the remarks either, although I'm not actually sure if it's against the DTD or not.

Kev: Double-nit. You can have "A precision", e.g. "This is accurate to a precision of one Cochrane." ;)

SamWhited: Ah yes; that's not using it in a plural / singular sense, so it works, but I can't actually figure out what the difference is… English is hard.

horazont:
[Voting started at 2017-08-30](https://mail.jabber.org/pipermail/members/2017-August/008636.html).

jcbrand:
> Voting started on 2017-08-30, voting period ends on 2017-09-13.

Everyone except the author (@linkmauve), who wasn't present, has already voted on this. Can we assume @linkmauve's consent or should we wait for Wednesday's meeting?

jcbrand:
@linkmauve Has now also voted +1. That means everyone has voted.

https://mail.jabber.org/pipermail/standards/2017-September/033237.html

horazont:
@jcbrand Did you intentionally omit sending the email for this change?

jcbrand:
@horazont I followed your instructions in the README and did the `--dry-run` command, but got no output. So it looked to me like no email has to be sent and I therefore assumed that emails are only sent out for XEPs whose statuses change (e.g. from ProtoXEP to experimental).

jcbrand:
@horazont The email has now been sent.

PR #470, opened 2017-06-01
Omemo signal protocol

iNPUTmice:
As suggested on the standards mailing list here is an updated version of XEP-0384 that actually reflects what has been implemented by several clients.
As one can see the changes are very minimal. It also adds a section on how the SignalProtocol has not been documented and only exists in GPLv3 implementations.
It also changes the track from Standards to Historical.
A rendered version can be found [here](https://gultsch.de/files/xep-0384.html).
Note: I haven't run this by the author yet. That's just my own personal idea. I can't say if @strb would agree to this.

### DIFF
--- a/xep-0384.xml
+++ b/xep-0384.xml
@@ -28,6 +28,15 @@
     <jid>andy@strb.org</jid>
   </author>
   <revision>
+    <version>0.2</version>
+    <date>2017-06-02</date>
+    <initials>dg</initials>
+    <remark>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
I updated this PR to not change the track.

SamWhited:
Just FYI, I did ask @strb about this PR; I think he was going to reply to the editor@ email or on list.

PR #471, opened 2017-06-07
XEP-0065: Remove duplicate example

vanitasvitae:
Example 10 was a duplicate (the stanza was a copy of ex. 9, the description a copy of example 11) so I removed it.


horazont:
Kevin suggested to play it safe and let this pass through council.

horazont:
[Voting started at 2017-08-30](https://mail.jabber.org/pipermail/members/2017-August/008636.html).

PR #474, opened 2017-06-13
Update MIX XEP-0369 to 0.9.3

stevekille:
Remove Legacy MIX Namespace;
      Add mix element in message to hold MIX additional information;
      Roster Update Clarifications;
      Clarify when messages are delivered to clients;


stevekille:
On hold, pending discussion

PR #476, opened 2017-06-15
XEP-0321: Use neutral pronouns instead of his/he

louiz:


horazont:
@louiz Can you please [sign the CLA using the CLAbot](https://cla-assistant.io/xsf/xeps?pullRequest=476)?

louiz:
Done.

xep-0321.xml
```diff
@@ -55,7 +55,7 @@
   <p>This document addresses the following requirements:</p>
   <ol>
     <li>Make it possible for remote services or entities to manage user's roster by the same mechanisms that descibed in the &rfc6121;.</li>
-    <li>Provide a way for user to control which services have permission to manage his roster.</li>
+    <li>Provide a way for user to control which services have permission to manage their roster.</li>
```
jcbrand: `for user` should be either `for the user` or `for users`. I personally think plural is better, in which case to above sentence also needs to be changed to plural, otherwise it should be `a user's roster`.

SamWhited: "their roster" works fine in the singular case (and sounds better than "a user's roster" IMO, although that's pretty bikesheddy)

jcbrand: +1 for `Provides` and `Makes`.

I'm not referring to the singular, non-gendered usage of `their`, I'm referring to the way user is referenced without an article, which IMO is not proper English (unless `user` is used as a proper noun, but then it should be capitalized).

So `Provide a way for user` should be `Provides a way for the user` or `Provides a  way for the user` (either is fine).

Similarly there's a missing article in the previous sentence.

horazont: I find it odd to continue the sentence "This document addresses the following requirements" with "provides foobar". I find "provide foobar" more natural.

Agreeing on the missing articles for "user". This, however, affects the whole XEP.

jcbrand: Ah, you're right, if `requirements` is plural, then it's just `provide`.

horazont: In any case, in the spirit of progress I’d merge this with "Provide a way for user**s**" (and some unrelated minor editing).

louiz: I agree, plural is better.

PR #477, opened 2017-06-16
XEP-0234: Fix a date in examples

linkmauve:


vanitasvitae:
Could you replace the TimeZone by 'Z'? XEP-0234 Table 2 states, that the timestamp MUST be UTC, so 'Z' would fit better here.

Flowdalic:
> XEP-0234 Table 2 states, that the timestamp MUST be UTC

I don't think that it does formally specify that (no MUST). For some obscure reasons it states "UTC Timestamp …", but I don't think it should require UTC (no other XEP does that AFAIK) and that we should simply remove 'UTC. Having "...which MUST conform to the DateTime profile of XMPP Date and Time Profiles (XEP-0082)" should be sufficient.

linkmauve:
I fixed this oversight, thanks. :)

lovetox:
@Flowdalic

https://xmpp.org/extensions/xep-0203.html

And lets not start to include unnecessary timezone informations in XEPs that absolutely not need them at all.

The only XEP where a timezone is a valid addition in my opinion is Entity Time.

Flowdalic:
You always need to specify TZD when the format is required to be XEP-0082 compatible, and you need a parser which is able to handle all valid TZD values. So what's the benefit of having UTC as default (and always explicitly stating that)? Furthermore some embedded devices could be only aware of their local timezones, but not the current offset from UTC (because of, well, timezone offsets are complicated).

lovetox:
You are grasping for straws.
If a device doesnt know his UTC Offset, it essentially cant deal correctly with delay timestampes (read MAM)

Also im not an expert on namespace bumps but this feels like one, and if you go on and change 0203 (because everything else would be quite inconsistent, we have to think of the devices that dont know its UTC offset) thats definitely one because, it there says MUST.

stpeter:
I don't know, this feels like a simple spec bug in the examples - changing 2015-07-26T21:46:00 to 2015-07-26T21:46:00Z seems like the right fix to me.

lovetox:
@Flowdalic editing your comment, lets look mine now out of context.

1. You dont need a parser that is able to handle all valid TZD values when the XEP states that only UTC is to be used. Of course its recommended that every client has such a method.

2. Even the 0082 security considerations state:

> To avoid this issue developers are advised to convert local time to UTC before sending ISO 8601 encoded times into the XMPP network.

So for me its a perfectly good way to limit XEPs that dont benefit from a non-UTC timestamp, and force developers to do the right thing if it makes no sense any other way. Instead of saying, do what you want but read up there is a sentence on the bottom where you can read what you actually should do, and what makes sense.

If you have the opinion anyway that every client has a method that parses all valid 0082 timestamps, then this would make no difference to you at all.

stpeter:
I don't have a problem with specifying UTC only (because a device might not know what timezone it's in), but adding "Z" to the examples is necessary.

lovetox:
@stpeter the example was wrong and fixing it was not the point of discussion.

Its the ongoing trend to allow all possible combinations of DateTime Profile even if it makes no sense. Then pointing to the security considerations. see https://xmpp.org/extensions/xep-0319.html

Flowdalic:
> You dont need a parser that is able to handle all valid TZD values when the XEP states that only UTC is to be used. 

As soon as you want to support one XEP which does not have the "TZD must be UTC" requirement  you do need one. 

It doesn't really matter that much to me. It's just a little bit strange to explicitly state the only valid option. Furthermore, not enforcing UTC doesn't mean you can not use it all the time.

linkmauve:
@lovetox, @stpeter, a device doesn’t need to know its own timezone to convert from the sender’s local timezone to UTC, you simply have to subtract the offset from the rest of the datetime.  And if the device doesn’t know its timezone, it won’t be able to deal with UTC either.

I am in favour of dropping the UTC requirement, it doesn’t bring anything privacy-wise, since it’s much more accurate to analyse activity patterns to pinpoint someone on a map if that’s what you are after.

PR #483, opened 2017-07-07
XEP-0357: Use server JID as 'from' address for notifications

weiss:
Let example 12 and 13 in XEP-0357 use the server JID rather than the user JID for publishing notifications, as suggested [on the list][1].

[1]: https://mail.jabber.org/pipermail/standards/2017-July/033050.html

### DIFF
--- a/xep-0357.xml
+++ b/xep-0357.xml
@@ -30,6 +30,12 @@
             <jid>lance@lance.im</jid>
         </author>
         <revision>
+            <version>0.3</version>
+            <date>2017-07-07</date>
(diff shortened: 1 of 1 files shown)


weiss:
I force-pushed a commit that also updates the [business rules][1] accordingly.  The app server is now expected to accept (at least) the server JID and the bare user JID as the `from` address for notifications.

[1]: https://xmpp.org/extensions/xep-0357.html#sect-idm45739234934960

legastero:
+1, merge it

PR #485, opened 2017-07-11
Split the dockerfile into two

mathieui:
- Create a `Dockerfile.base` that contains all the dependencies and required directory structure
**that image will need to be published to the hub** (as xmppxsf/xeps-base currently)
- Put minimal content into the `Dockerfile` and allow the number of make jobs and the make targets to be parametrized through docker build arguments.

This should, once someone publishes the base image to the hub, cut the full build time in half or more (because the old dockerfile was mixing copy and install operations, therefore invalidating part of the cache whenever the source changed).

Providing the `NCORES` build arg lets the editor/user choose how many jobs should be used at the same time, which defaults to 1, but should be set as high as the number of cores in the build machine. From my (very unscientific) tests, build speed increases proportionally with the number of jobs running, since latex output and CPU performance is the bottleneck.
The `TARGETS` build arg also allows customization of what should be running.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,16 @@
+# Builds all XEPs by default, HTML & PDF
+# docker build . --build-arg NCORES=X --build-arg TARGETS="target1 target2"
+# from Dockerfile.base
 FROM xmppxsf/xeps-base:latest
+
+ARG NCORES=1
+ARG TARGETS="html pdf"
 
 COPY *.xml xep.* *.css *.xsl *.js *.xsl Makefile /src/
 COPY resources/*.pdf /src/resources/
 
 WORKDIR /src
-RUN OUTDIR=/var/www/html/extensions/ make -j5 html pdf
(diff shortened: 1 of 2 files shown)


Kev:
Looks fine to me. The right way for us to get the base image onto the hub is to break it out into its own repo, and then I can set up an automated build for that too.

Kev:
I've created the (unpopulated) https://github.com/xsf/xep-docker-base repo. I've also set up automated builds for it to go to xmppxsf/xeps-base. So I think if the base dockerfile is uploaded there, it'll trigger a build, and once that's done, merging this (without the base or with) would build.

guusdk:
I have now added the 'base' image to https://github.com/xsf/xep-docker-base (I've also applied some changes - feel free to comment on those here, or there).

guusdk:
The base image is now available on dockerhub. I'll merge this PR now (to later remove the base dockerfile). 

As this PR does not touch the XEPs, not having a signed CLA shouldn't be an issue.

PR #487, opened 2017-07-17
XEP-0319: Make the schema more precise about a date being a xs:dateTime

linkmauve:
This section is non-normative, and is already specified more strongly in XEP-0082, so it shouldn’t change anything.


horazont:
The reasoning depends on #469, will at least need to figure out if council is needed for that change.

PR #489, opened 2017-07-30
XEP-0234: Add missing length attribute to XML Schema

vanitasvitae:
fileTransferRangeType was missing the length attribute in the XML schema.


horazont:
@vanitasvitae Please add a revision block.

vanitasvitae:
@horazont fixed :)

PR #492, opened 2017-08-07
XEP-0359: clarifications and removed 'overwrite' wording

iNPUTmice:
This commit changes some of the wording to clarify that the by attribute
for messages tagged by the server should be either the account jid or
the room jid (as opposed to the server jid for example)

To further clarify this the examples (including the disco example) were
all changed to have the same jid/by-attribute. (A room in that example)

The business rules had some wording that an entity would overwrite the
ID in existing 'faked' stanza-ids instead of removing the faked and
re-adding a non-faked one. However that doesn't work for faked messages
that are not archived (don't have a stanza-id).

rendered version [here](https://gultsch.de/files/xep-0359.html)

### DIFF
--- a/xep-0359.xml
+++ b/xep-0359.xml
@@ -9,7 +9,7 @@
 <xep>
 <header>
   <title>Unique and Stable Stanza IDs</title>
-  <abstract>This specification describes unique and stable IDs for stanzas.</abstract>
+  <abstract>This specification describes unique and stable IDs for messages.</abstract>
   &LEGALNOTICE;
   <number>0359</number>
   <status>Experimental</status>
@@ -23,6 +23,17 @@
   <supersededby/>
   <shortname>stanza-id</shortname>
   &flow;
+  <revision>
+	  <version>0.5.0</version>
+	  <date>2017-08-23</date>
+	  <initials>dg</initials>
+	  <remark>
+		  <ul>
+			  <li>Business rules clarifications</li>
+			  <li>Define stanza-ids for messages only</li>
+		  </ul>
+	  </remark>
+  </revision>
   <revision>
     <version>0.4.1</version>
     <date>2017-05-20</date>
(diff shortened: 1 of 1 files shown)


xep-0359.xml
```diff
@@ -135,25 +135,25 @@
...
+	<li>Stanza ID generating entities, which encounter a &stanza-id; element where the 'by' attribute matches the 'by' attribute they would otherwise set, MUST delete that element even if they are not adding their own stanza ID.</li>
 	<li>Stanzas MUST possess, in the direct child level of the stanza, at most one &stanza-id; extension element with the same XMPP address as value of the 'by' attribute.</li>
 	<li>Every &stanza-id; extension element MUST have the 'id' attribute and the 'by' attribute set.</li>
 	<li>Every &stanza-id; and &origin-id; extension element MUST always possess an 'id' attribute and MUST NOT have any child elements or text content.</li>
-	<li>The value of the 'by' attribute MUST be the XMPP address of the entity assigning the unique and stable stanza ID. Note that XMPP addresses are normalized as defined in &rfc6122;</li>
+	<li>The value of the 'by' attribute MUST be the XMPP address of the entity assigning the unique and stable stanza ID. For regular messages the assigning entity is the account. In groupchats the assigning entity is the room. Note that XMPP addresses are normalized as defined in &rfc6122;.</li>
```
iNPUTmice: Probably not. Other ideas? chat messages? One-on-one messages? non-group chat messages?

horazont: @Flowdalic, @iNPUTmice, what’s the status here?

iNPUTmice:
* Changed wording of 'regular' messages to one-on-one messages.
* Removed first business rule (the one about entities not removing stanza-ids which stands in contrast to the other business rule about deleting unsafe ids.). I think server not deleting arbitrary elements is the default thus it doesn't need explicit mentioning. 
* Made this XEP about messages only as per discussion on the list.

horazont:
@Flowdalic Please check if the new commits address your comments.

Flowdalic:
> I think server not deleting arbitrary elements is the default thus it doesn't need explicit mentioning.

I think it's common practice in at least some (commercial) MUC service implementations. That was my rationale for explicitly stating it.

iNPUTmice:
@Flowdalic I reintroduce that rule in a way that hopefully doesn't conflict with the other business rules.

Flowdalic:
Looks good to me. BTW: Feel free to add yourself to the 'acknowledgments' section if you want to. :)

iNPUTmice:
squashed. Please merge.

horazont:
@iNPUTmice In general, please avoid squashing after the end of an discussion. I don’t know of a way to verify that the changes in the squash are the same which were ack-ed by the author earlier unless I have followed the discussion closely.

SamWhited:
Personally, I disagree: we should *always* be squashing at the end if the changes are one logical change. It keeps the history clean. If you want to verify (though I think it's it's fine to trust the authors) you can verify that the list of changed files is the same and take a checksum of each file before and after.

horazont:
@SamWhited Except that I can’t access the "before" commits anymore after the squash to do the verification.

SamWhited:
Sure you can; make a new branch before you reset on the squashed branch or dig them out of the reflog if you already updated them.

horazont:
I think you’re assuming that I already pulled the branch before the squash happened (which is usually not the case).

SamWhited:
Let's take further discussion to the chat room, but either way I don't think this is a problem and we should be squashing.

PR #495, opened 2017-08-20
README: move editor readme to markdown readme

SamWhited:
I'm not sure if it's possible (GitHub's permissions system is neither very clear, or very flexible as far as I can tell) to give other people permission to sort / organize / etc. issues, but I've added a section on that just in case. Maybe it would be helpful (or maybe no one would do it, no idea; take it as a suggestion).

Also included a section on accepting protoxeps and updating existing xeps.

Left out anything about archiving because it's still manual and might change soon.

Thoughts? Better or worse than the existing document?

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,4 +50,128 @@ For more information try `make help`.
 Using Docker
 ------------
 
-A full set of HTML and PDFs can be generated inside a docker container, with no dependencies on the host other than Docker itself, and served by nginx in the container. To build the template `make docker`, to run it `make testdocker` (serves on http://localhost:3080), and to stop/delete it afterwards `make stopdocker`
+A full set of HTML and PDFs can be generated inside a docker container, with no
+dependencies on the host other than Docker itself, and served by nginx in the
+container. To build the template `make docker`, to run it `make testdocker`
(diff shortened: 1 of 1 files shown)


stpeter:
👍

horazont:
This is better since markdown is more readable than the XML document. Even though the XML document can be built by anyone competent enough to be an editor.

Re github permissions: I don’t think you can give people permission to change issues without push permission, however, you can *protect* branches. So it should be possible to have people added as collaborators (should give them permission to modify issues), but protect the master branch so that only editors can push to it. Collaborators would be able to create branches, but that should not affect us. Would require support from iteam or someone with more permissions than I have on the xeps repository.

guusdk:
I don't think we should rely to much on the flexibility of the Github permission system. I'm not a big fan of the workaround that's described - it's not really transparant, and will start to introduce confusion pretty fast. I don't have a better suggestion than asking people to submit PRs for gardening-purposes.

SamWhited:
Yah, sounds like that's not really possible with GitHub. I'll leave the gardening section in case it's a process that editors like (if not of course, feel free to suggest alternatives), but it doesn't have to be for other colaborators.

PR #497, opened 2017-08-21
XEP-0234, XEP-0300: Add hash-used element to hashes and make use of it in Jingle File Transfer

vanitasvitae:
As discussed in https://mail.jabber.org/pipermail/standards/2017-August/033116.html

### DIFF
--- a/xep-0234.xml
+++ b/xep-0234.xml
@@ -25,6 +25,12 @@
   &stpeter;
   &lance;
   <revision>
+    <version>0.18.2</version>
+    <date>2017-08.21</date>
+    <initials>ps</initials>
+    <remark><p>Make use of &lt;hash-used/&gt; from XEP-0300.</p></remark>
+  </revision>
+  <revision>
     <version>0.18.1</version>
     <date>2017-05-20</date>
     <initials>egp</initials>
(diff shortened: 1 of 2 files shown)


horazont:
Soo, @vanitasvitae, this PR will be my first victim, being new to the editor team. So let’s try to make this roll!

I requested a review from @tfar for the XEP-0300 part, and will try to hunt down Lance or Peter for comments on the XEP-0234 part (and possibly the XEP-0300 part).

stpeter:
@horazont You can find Lance via @legastero

xep-0234.xml
```diff
@@ -354,7 +360,12 @@
     <tr>
       <td>hash</td>
       <td>A hash of the file content, using the &lt;hash/&gt; element defined in &xep0300; and qualifed by the 'urn:xmpp:hashes:2' namespace. Multiple hashes MAY be included for hash agility.</td>
-      <td>REQUIRED when offering a file, otherwise OPTIONAL</td>
+      <td>See &lt;hash-used/&gt;</td>
+    </tr>
+    <tr>
+      <td>hash-used</td>
+      <td>Alternatively to a &lt;hash/&gt; element, the initiator can also include a &lt;hash-used/&gt; element. This comes handy to avoid reading the file twice to calculate the hash.</td>
```
stpeter: I don't think we mean "Alternatively" here because it seems that the <hash-used/> element is sent in addition to the <hash/> element (at least if I'm reading the text below correctly). An example would be helpful. English nit: "This comes in handy..." (or, less colloquially, "This avoids the need to read...").

Flowdalic: > it seems that the "hash-used" element is sent in addition to,

Just to clarify: The idea is that it is **not** send in addition to. You either send a hash element, which includes the hash algo and the hash value, OR you send a hash-used, which just includes the hash algo.

stpeter: If so, then I must be misreading the text around line 799, which says:

> In such a case however, the session-initiate message MUST contain a &lt;hash-used/&gt; element....  Additionally, the &lt;checksum/&gt; element MUST contain a &lt;file/&gt; element which MUST contain at least one &lt;hash/&gt; element....

In this situation, would there be both a "hash-used" element and a "hash" element?

vanitasvitae: I probably forgot to change the second sentence. It should be changed to say "at least one hash or hash-used element".

stpeter:
That'd be better. ;-) +1

vanitasvitae:
I changed the wording.

horazont:
@stpeter Do you, as author, approve the XEP-0234 changes in this PR and/or shall we wait for Lance’ comments?

legastero:
+1 @horazont

stpeter:
+1 here as well @horazont

PR #498, opened 2017-08-23
Editor Tooling revamp + inbox build & cleanup

horazont:
Re-work of the email sending utilities.

New stuff:

1. ``./tools/extract-metadata.py`` extracts metadata from all XEPs into a single XML (written to stdout; use ``make build/xeplist.xml`` for a convenient file).

2. ``./tools/send-updates.py`` writes emails based on the difference between two xeplist XML files, like the previous ``announce.py`` and ``inxep.py`` did. It has a configuration file (``--help`` will tell you more) to configure mail server, user, password and sender address.

3. ``./tools/deferrals.py`` lists all XEPs which are Experimental and have not been updated for more than one year.

4. ``./tools/archive.py`` copies changed XEPs (again based on the diff between two XEP lists) into the attic.

In addition:

* Cleaned up the inbox so that everything builds.
* Add build targets to build the inbox.
* Make the inbox and xeplist being built in Docker.


dwd:
I noticed you've got a sasl2.html file being added - is that right?

horazont:
@dwd Ha, good catch! That’s obviously a mistake.

I recovered the original sasl2.xml from the git log and added that instead.

inbox/Makefile
```diff
@@ -0,0 +1,8 @@
+OUTDIR?=../build/
+
+xeps=$(wildcard *.xml)
+
+.PHONY: all
+all: html
+
+html: $(patsubst %.xml, $(OUTDIR)/%.html, $(xeps))
```
SamWhited: This doesn't appear to actually do anything; we probably shouldn't copy/paste the logic into the inbox anyways, I think building inbox files can be done from the main makefile too with some tweaks.

Makefile
```diff
@@ -33,17 +49,33 @@ help:
 .PHONY: all
 all: html
 
+$(OUTDIR)/inbox/%: build/inbox
+build/inbox:
```
SamWhited: This should be `$(OUTDIR)/inbox` or, alternatively, tweak the `$(OUTDIR):` rule to `mkdir -p $(OUTDIR)/inbox` and always create both and remove this

Makefile
```diff
@@ -33,17 +49,33 @@ help:
 .PHONY: all
 all: html
 
+$(OUTDIR)/inbox/%: build/inbox
```
SamWhited: This can be deleted; I'm reasonably sure that this rule will never be hit if a more specific rule exists (and if one doesn't, this will do nothing so it can be removed)

Makefile
```diff
@@ -54,19 +86,26 @@ xep-%.html: $(OUTDIR)/xep-%.html ;
...
+
 $(EXAMPLESDIR)/%.xml: xep-%.xml $(XMLDEPS) examples.xsl $(EXAMPLESDIR)
 	xsltproc --path $(CURDIR) examples.xsl "$<" > "$@" && echo "Finished building $@"
 
 $(REFSDIR)/reference.XSF.XEP-%.xml: xep-%.xml $(XMLDEPS) ref.xsl $(REFSDIR)
 	xsltproc --path $(CURDIR) ref.xsl "$<" > "$@" && echo "Finished building $@"
 
-$(OUTDIR)/%.html: %.xml $(XMLDEPS) $(HTMLDEPS)
+$(all_xep_htmls): $(OUTDIR)/%.html: %.xml $(XMLDEPS) $(HTMLDEPS)
+	mkdir -p $(OUTDIR)/inbox
```
SamWhited: This should probably have already been created by one of the prerequisites.

SamWhited:
General comment: large PRs like this should probably be split up into several smaller PRs. It was lucky that Dave caught the added HTML file, with a diff this large it's really hard to spot problems like that.

tools/extract-metadata.py
```diff
@@ -160,34 +161,47 @@ def main():
...
+        try:
+            with xepfile.open("rb") as f:
+                parsed = open_xml(f)
+
+            tree.append(make_metadata_element(
+                number,
+                extract_xep_metadata(parsed),
+                True,
+            ))
+        except Exception as exc:
```
SamWhited: What exceptions can this throw and are they likely? I think in CI it would be nicer if this just raised the acception and crashed since it's probably only likely to happen if some OSException occures (which is actually exceptional). Having a `has_error` flow is confusing and feels brittle too.

inbox/Makefile
horazont: Ops, that makefile should not have been commited, right!

tools/extract-metadata.py
horazont: Especially for CI it makes sense to print all errors before exiting so that you see what went wrong. I can make this more specific to do that only for XML parser errors, that probably makes sense (since the error message won’t be helpful for other errors which could occur due to incomplete XML).

horazont:
> General comment: large PRs like this with many unrelated changes should probably be split up into several smaller PRs. It was lucky that Dave caught the added HTML file, with a diff this large it's really hard to spot problems like that.

I agree. If it still makes sense, I can split the inbox changes into a separate PR.

tools/extract-metadata.py
SamWhited: XML parsing errors will already be found when building the file, I wouldn't make this more complicated than it needs to be unless we have a specific use for this later.

PR #499, opened 2017-08-23
XEP-0045: Clarify wording for a client re-syncing to a MUC

ge0rg:
TL;DR: whenever a client (re)sends a join, the MUC service should return
everything a newly joining client needs to know.

Discussion in
https://mail.jabber.org/pipermail/standards/2016-June/031180.html

### DIFF
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -1385,6 +1385,7 @@
 </presence>
 ]]></example>
       <p>Before attempting to enter the room, a MUC-compliant client SHOULD first discover its reserved room nickname (if any) by following the protocol defined in the <link url='#reservednick'>Discovering Reserved Room Nickname</link> section of this document.</p>
+      <p>When a MUC service receives an &lt;x/&gt; tagged join stanza from an already-joined client (as identified by the client's full JID), the service SHOULD assume that the client lost its synchronization, and therefore it SHOULD send exactly the same stanzas to the client as if it would actually join the MUC. The server MAY also send a presence update to the other participants according to the received join presence.</p>
     </section3>
 
     <section3 topic='Presence Broadcast' anchor='enter-pres'>
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -1385,7 +1385,7 @@
 </presence>
 ]]></example>
       <p>Before attempting to enter the room, a MUC-compliant client SHOULD first discover its reserved room nickname (if any) by following the protocol defined in the <link url='#reservednick'>Discovering Reserved Room Nickname</link> section of this document.</p>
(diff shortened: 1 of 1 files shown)


stpeter:
Seems like a fine idea.

ge0rg:
Rendered version: last paragraph in https://op-co.de/tmp/xep-0045.html#enter-muc

xep-0045.xml
```diff
@@ -1385,6 +1385,7 @@
 </presence>
 ]]></example>
       <p>Before attempting to enter the room, a MUC-compliant client SHOULD first discover its reserved room nickname (if any) by following the protocol defined in the <link url='#reservednick'>Discovering Reserved Room Nickname</link> section of this document.</p>
+      <p>When a MUC service receives an &lt;x/&gt; tagged join stanza from an already-joined client (as identified by the client's full JID), the service SHOULD assume that the client lost its synchronization, and therefore it SHOULD send exactly the same stanzas to the client as if it would actually join the MUC. The server MAY also send a presence update to the other participants according to the received join presence.</p>
```
SamWhited: > SHOULD assume that the client lost its synchronization, and therefore it

nit: I'd strike this and just say "the service SHOULD send…"

> send exactly the same stanzas to the client as if it would actually join the MUC
 
I think this needs a reference to what those stanzas are "(See section x.x)". Also,  minor nit, "as if it would actually join the MUC" needs to be rephrased, it doesn't sound right.

ge0rg: > strike this and just say "the service SHOULD send…"

Having rationale for design decisions is a very important part of a good specification. I understand that the placement of "SHOULD" is not perfect here, and I'm open for better wording suggestions that don't remove the rationale.

> [...] needs to be rephrased

Would "as if it actually joined the MUC" be a better wording? I'm not a native speaker, and I struggled a bit with writing this. Again, I'm open for suggestions.

SamWhited: This isn't a rationale for design decisions, it's repeating what's said in the next sentence.

horazont:
Discussed on 2017-08-23 in board meeting, see [Trello Card](https://trello.com/c/Q7An3b0T/).

xep-0045.xml
ge0rg: I think there is a difference between the two sentences. First, what is the situation:

*the service SHOULD assume that the client lost its synchronization,*

then, how to handle it:

*and therefore it SHOULD send exactly the same stanzas to the client as if it would actually join the MUC*

Maybe it's not a full-fledged rationale, but at least an explanation of why we want that behavior. Anyway, feel free to propose a better wording that's more to the point :)

> What does "actually joined" mean?

The server should send everything to the client which it would send as response if the client had requested a join and hadn't been part of the MUC before. Does that sound better?

SamWhited:
s/board/council/ :)

PR #502, opened 2017-08-25
XEP-0388: Fixes, clarifications, and schemas

dwd:
Fixes:

* That's a lot of XML errors. Sorry.
* Clarified whose additional-data this is.
* &lt:next> uses &lt;initial-response>.

Added a few example flows.

Finally, added a (probably wrong) schema.

### DIFF
--- a/xep-0388.xml
+++ b/xep-0388.xml
@@ -21,6 +21,19 @@
   <shortname>sasl2</shortname>
   &dcridland;
   <revision>
+    <version>0.2.1</version>
+    <date>2017-08-24</date>
+    <initials>dwd</initials>
+    <remark>
+      <ul>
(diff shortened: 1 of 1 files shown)


xep-0388.xml
```diff
@@ -21,6 +21,19 @@
...
   &dcridland;
   <revision>
+    <version>0.2.1</version>
+    <date>2017-08-24</date>
+    <initials>dwd</initials>
+    <remark>
+      <ul>
+          <li>That's a lot of XML errors. Sorry.</li>
+          <li>Clarified whose additional-data this is.</li>
+          <li>&lt:next> uses &lt;initial-response>.</li>
```
horazont: Please fix the broken XML syntax there.

dwd: Sorry - fixed.

PR #504, opened 2017-09-01
Extend the readme with new tooling and workflow

horazont:
I essentially wrote down what I did when I merged a bunch of PRs recently. I’ll put this here in case anyone wants to take a look at it, otherwise I’ll simply merge it next week or so.


jcbrand:
Thanks for writing this. A lot of this is still new to me, so I've been spending quite a bit of time reading up, including this README.

PR #505, opened 2017-09-01
Fixed two typos

jcbrand:
* 'in' should be 'is'
* 'approbved' should be 'approved'.

I assume this doesn't warrant a new revision section, correct?


SamWhited:
> I assume this doesn't warrant a new revision section, correct?

I never bother for simple editorial fixes like this.

Kev:
> I assume this doesn't warrant a new revision section, correct?

We decided a few years ago that XEP versions should be stable, but that it wasn't worth the normal version bump for minor editorial changes, so they'd get a new version that only changes the patch version part (or whatever you want to call the x part of 0.0.x) - in this case bumping from 1.21.1 to 1.21.2.

jcbrand:
> in this case bumping from 1.21.1 to 1.21.2.

Done.

SamWhited:
Up to you of course, but I wouldn't waste my time with it.

jcbrand:
> Up to you of course, but I wouldn't waste my time with it.

If @Kev is ok with it, I'll remove it again, and if both of you are happy I'll then squash and merge.

PR #506, opened 2017-09-06
xep-0234: fix broken internal link

fippo:
as noted in https://mail.jabber.org/pipermail/standards/2017-August/033128.html

### DIFF
--- a/xep-0234.xml
+++ b/xep-0234.xml
@@ -399,7 +399,7 @@
       <td>OPTIONAL</td>
     </tr>
   </table>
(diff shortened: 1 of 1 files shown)


jcbrand:
This has been discussed on the standards list, so my understanding is that we don't need the author's explicit approval. Therefore adding "ready-to-merge" label.

fippo:
@stpeter @legastero FYI (-:

jcbrand:
@horazont Should an email be sent out for this one as well? There's no version change.

horazont:
I wouldn’t bother for purely editorial changes.

Kev:
If there's a new version published (that is, the content, other than stylesheet, changes on the website), there should be a new version number (although a z from x.y.z), which implies also an announcement.

SamWhited:
I also don't bother for small editorial changes.

PR #508, opened 2017-09-12
XEP-0385 (SIMS): Use 'hashes:2' (XEP-0300)

Flowdalic:


linkmauve:
This missed bumping the namespace to JingleFT:5, which is the one introducing hashes:2.

Flowdalic:
> This missed bumping the namespace to JingleFT:5, which is the one introducing hashes:2.

Good catch. I wonder if we could jingleft to 5 and leave sims at 1. In other words: Is a sims namespace bump required when we bump jingleft to 5 in SIMS?

tfar:
IMHO, SIMS is pretty independent of ft and hashes versions. While the `<media-sharing xmlns='urn:xmpp:sims:2'>` element contains the `file` element at a specific version that contains a `hash` element at a specific version.

For 1-to-1 compatibility, clients can see what versions the other party supports and send the correct version accordingly, if it supports them. In multi-party scenarios it gets trickier, but that's also the case if the version of the media-sharing element would increment together with the version of the sub elements. Or am I missing something?

Flowdalic:
I've added a commit on top of this PR to avoid the sims namespace bump, which is IMHO not necessary. 

> I also have a WIP version of JingleFT bumping it to :6, which I hope won’t need any more bump in the future, it may be sensible to stall this PR until I’m done with it.

I don't think there is much gain in waiting besides a little less editor work. If editors are overloaded then sure, but otherwise I would just continue processing this PR.

PR #509, opened 2017-09-12
Update JET

vanitasvitae:
Get rid of dependencies on OMEMO and OX. Instead use stub encryption methods.
This commit prepares for addon specs (jet-omemo, jet-ox).


horazont:
For future reference: XEP-0001 states that updates to ProtoXEPs can be made before council decides based on the feedback from the discussion.

PR #511, opened 2017-09-17
Limit text width of XEPs on large viewports

SamWhited:
XEPs are currently very hard to read even on fairly standard sized 1080p monitors because the text is more-or-less full width. It would also be nice to scale the fontsize based on the viewport size at a later date, but incremental changes.

/cc @guusdk because I think the website serves this file from somewhere else and it might have to be copy/pasted (sigh)


horazont:
This is a massive improvement. Good work!

(I don’t think we should be scaling the font size based on viewport though. That’s up to the UA / user config, not something each website should be doing individually.)

PR #516, opened 2017-09-22
XEP-0359: make the by attr optional on origin IDs

SamWhited:
* Optionally allow the by attr for origin IDs
* Update reference to XMPP-Addr
* Retab mixed spaces/tabs

/cc @Flowdalic

### DIFF
--- a/xep-0359.xml
+++ b/xep-0359.xml
@@ -24,15 +24,26 @@
   <shortname>stanza-id</shortname>
   &flow;
   <revision>
-	  <version>0.5.0</version>
-	  <date>2017-08-23</date>
-	  <initials>dg</initials>
-	  <remark>
-		  <ul>
-			  <li>Business rules clarifications</li>
-			  <li>Define stanza-ids for messages only</li>
-		  </ul>
-	  </remark>
+    <version>0.5.1</version>
+    <date>2017-09-22</date>
+    <initials>ssw</initials>
+    <remark>
+      <ul>
+        <li>Make the by attribute optional on origin IDs</li>
+        <li>Update XMPP addr reference</li>
+      </ul>
+    </remark>
+  </revision>
+  <revision>
+    <version>0.5.0</version>
+    <date>2017-08-23</date>
+    <initials>dg</initials>
+    <remark>
+      <ul>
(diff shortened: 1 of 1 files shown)


Flowdalic:
I implicitly assumed that the value of the 'by' attribute of origin-id is just the same as the value of the 'from' attribute of the stanza. I know that this is not true with MUC reflection and semi-anonymous rooms. But I never thought of this of a problem. Could you elaborate on the motivation for this change?

Also stuffing whitespace fixes, minor changes and semantic changes in the same commit doesn't really make reviewing easy. I'm surprised to see that coming from you, as it's something I'd expect you to say too :) Could you split it into multiple commits?

SamWhited:
> I implicitly assumed that the value of the 'by' attribute of origin-id is just the same as the value of the 'from' attribute of the stanza. I know that this is not true with MUC reflection and semi-anonymous rooms.

This was the use case I had in mind; I want to know who set the ID, not necessarily who sent me the message.

> Also stuffing whitespace fixes, minor changes and semantic changes in the same commit doesn't really make reviewing easy. I'm surprised to see that coming from you, as it's something I'd expect you to say too :)

I generally would, but this is so tiny I didn't bother.

horazont:
I [moved the technical discussion to standards@](https://mail.jabber.org/pipermail/standards/2017-September/033423.html). Please continue it there.

On Freitag, 22. September 2017 17:51:43 CEST Sam Whited wrote:
> > Also stuffing whitespace fixes, minor changes and semantic changes in the
> > same commit doesn't really make reviewing easy. I'm surprised to see that
> > coming from you, as it's something I'd expect you to say too :)
>
> I generally would, but this is so tiny I didn't bother.

FWIW, I found it hard to figure out what this PR does due to the mix of changes. The diff highlighting is completely broken by mixing the indentation changes with the technical changes. So it’s especially important if the technical change is minor, I think.

(It also took me some time to realize why you "added" two version blocks there. I’m bad at reading diffs.)

SamWhited:
This is a discussion between me and the author, no need to spam standards@, let's continue here.

Flowdalic:
I'd actually see that discussion on standards@.

PR #517, opened 2017-09-23
XEP-0030: Service Discovery - Removed the order restriction of identities/features in the disco-info XML schema

Syndace:
I was using the XML schema files provided by XEP-0030 to validate some XMPP traffic and stumbled upon a stanza that did not pass the validation process because the order of identities and features was wrong.

I carefully read the specification and asked the jdev muc for advice, we came to the conclusion that the schema file is too restrictive and the order of identities and features should not be forced.

This PR removes the order forced by the current schema.


horazont:
After discussion in the editor MUC, we judged this change to be editorial and it’ll be merged soon. Thanks!

PR #532, opened 2017-11-01
update unicode ent to reference latest version

SamWhited:
Also add example of referencing a specific version.

These don't appear to be in use anywhere, so they won't break anything existing. I will be using them "soon" so I thought I'd go ahead and update them.


stpeter:
Yes, it's best for us to reference http://www.unicode.org/versions/latest/ ... although, note that PRECIS and IDNA are now stuck on an older version, as mentioned in section 13.5 of RFC 8264. :(

PR #533, opened 2017-11-02
XEP-0387: bump compliance suite year to 2018

SamWhited:
Bumps the year on the compliance suites to 2018.


horazont:
I don’t think this is editorial, at all. But the change has been discussed on-list and was proposed by the author, so [Ready to Merge] holds IMO.

SamWhited:
I forgot about this; it sounded like everyone was for doing it this way, so I'm going to go ahead and change it and the council will vote on the updated version with 2018 in the title. Thanks for the reminder!

PR #546, opened 2017-11-24
XEP-0107: The defined child of mood is REQUIRED, see §2.1

linkmauve:


horazont:
I don’t think this is correct. Some of the use cases publish empty <mood/> to signify un-publishing.

horazont:
I’m going to close this for inactivity. Feel free to re-open if you have any more feedback on this.

PR #547, opened 2017-11-24
XEP-0313: do not store type groupchat in user archives

iNPUTmice:
An update to the MAM XEP to reflect real word situation where implementations don't actually store group chat messages into the user archive.

Change made after discussion on mailing list. cc @mwild1


mwild1:
@Kev That's exactly what it's saying. Contacts shouldn't be able to choose what happens to your archive.

iNPUTmice:
If this:
> A MUC archives allows a user to view the conversation within a room. All messages sent to the room that contain either a &lt;body&gt; element **or a &lt;store/&gt; hint (&xep0334;)** SHOULD

is the line that is causing issues here I'm happy to remove in order to get the way more important not storing type=groupchat in user archives statement merged.

horazont:
@Kev ping.

Kev:
I already took the discussion back to the mailing list yesterday.

jcbrand:
The mailing list discussion is here: https://mail.jabber.org/pipermail/standards/2017-November/thread.html#33917

The conclusion appears to be to allow servers to store groupchat messages in the user's own MAM archive, but to not return them by default.

E.g. `Saying it shouldn’t store them isn’t helpful, but I’m ok with not returning them by default. We can add further filtering in other XEPs, I think.` by @Kev 

In light of this, can this PR be closed, or should it be updated?

Kev:
No activity in years.

PR #548, opened 2017-11-26
XEP-0144: The specification refers to a bare JID but the example was a FULLJID

detrout:


detrout:
Had to read the CLA and decide if it was a sufficiently free software compatible license before I signed it.

horazont:
@detrout Thank you! I’ll take care of this today.

PR #550, opened 2017-11-29
0.13

stpeter:
Addressed feedback received during Last Call.


SamWhited:
Thanks! I'll merge this shortly as #551 which is the same commit but with a few editor tweaks.

PR #553, opened 2017-12-03
XEP-0363: amend security considerations and introduce retry element

iNPUTmice:
* Added `<retry/>` element to error response
* Extend Security Considerations

### DIFF
--- a/xep-0363.xml
+++ b/xep-0363.xml
@@ -29,6 +29,17 @@
     <jid>daniel@gultsch.de</jid>
   </author>
   <revision>
+    <version>0.4.0</version>
+    <date>2017-12-03</date>
+    <initials>dg</initials>
+    <remark>
+      <ul>
+        <li>Added &lt;retry/&gt; element to error response</li>
+        <li>Extend Security Considerations</li>
+      </ul>
+    </remark>
+  </revision>
+  <revision>
     <version>0.3.1</version>
     <date>2017-04-21</date>
     <initials>dg</initials>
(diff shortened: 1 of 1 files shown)


xep-0363.xml
```diff
@@ -210,6 +221,7 @@
   </error>
 </iq>]]></example>
 <p>For any other type of error the service SHOULD respond with appropriate error types to indicate temporary or permanent errors.</p>
+<p>For temporary errors such as exceeding a personal quota the service MAY include a &lt;retry/&gt; element qualified by the urn:xmpp:http:upload:0 namespace as a child of the &lt;error/&gt; element. That element MUST include an attribute 'stamp' which indicates the time at which the requesting entity MAY try again. The format of the timestamp must adhere to the date-time format specified in &xep0082; and MUST be expressed in UTC. The service SHOULD NOT impose sanctions on an entity simply for retrying earlier than the specified time.</p>
```
SamWhited: * s/That element MUST/The retry element MUST/
* s/requesting entity MAY try again/requesting entity may try again/ (this is passively describing what the entity may be doing, so I don't think it's actually 2119 language?)
* "entity ~simply~ for retrying earlier"

xep-0363.xml
```diff
@@ -251,6 +265,7 @@
 <section1 topic='Security Considerations' anchor='security'>
   <ul>
   <li>Client implementors MUST consider the fact that without additional end-to-end-encryption files uploaded to a service described in this document will store those files in plain text on that service. Client implementors SHOULD either use this only for semi public files (for example files shared in a public MUC or a PEP Avatar) or implement appropriate end-to-end encryption.</li>
+  <li>Requesting entities MUST strip any newline characters from the HTTP header names and values before making the PUT request.</li>
```
SamWhited: I wonder if we should just say that headers must be canonicalized as described in the HTTP spec? I can't remember if this includes newlines or if it's just the capitalization of the header keys, but I'm pretty sure there's a section in there about that (although I can't find it now).

iNPUTmice: There is already wording in the XEP that the actual upload is out of scope for this document thus repeating (or referencing) HTTP spec is unnecessary IMHO. I see this part of the security section just as a friendly reminder in case some HTTP libraries don't validate the header input.

SamWhited: Sounds good to me

PR #554, opened 2017-12-06
XEP-0048: Suggest that xep48 implementations support xep49

Kev:
Per Council discussion.


horazont:
Rejected by council ([see mailinglist archive](https://mail.jabber.org/pipermail/standards/2018-January/034161.html)).

PR #555, opened 2017-12-07
XEP-0060: Register pubsub#persist_items as a Precondition publish-option

iNPUTmice:
` pubsub#persist_items` is used in XEPs like XEP-0048 or XEP-0223 and needs to be registered according to the wording in the publish-options section. Otherwise implementations will just reject the examples given in those XEPs.

PR #556, opened 2017-12-07
XEP-0060: change wording on publish-options to refer to node-configuration

iNPUTmice:
This is an alternative approach #555 that spares us from having to re-add every node-configuration as a precondition to the publish-options registry. 

* Specify that unregistered publish-options will fall back to preconditions to node configurations of the same name
* Get rid of per-item OVERRIDE
* Remove registration for the obsolete pubsub#access_model publish-options

### DIFF
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -51,11 +51,24 @@
   &ralphm;
 
   <revision>
+    <version>1.15.0</version>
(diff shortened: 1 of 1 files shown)


horazont:
Closing in favour of merged #557.

PR #568, opened 2018-01-10
New Proto-XEP: Easy User Onboarding

ge0rg:
This document defines a protocol and URI scheme for user invitation in order to allow a third party to register on a server. The goal of this is to make onboarding for XMPP IM newcomers as easy as possible.

Rendered version: https://op-co.de/tmp/user-invite.html


horazont:
Marc Schink agreed outside of github.

PR #569, opened 2018-01-10
XEP-0387: Address feedback made in the period of the new Council

Kev:
This addresses, I believe, all the issues I raised. Where I was wrong (84), or it was contentious (220), I've dropped the point. Where Sam felt it was worth watering down (needing 153, but only read-only), I've done so.

### DIFF
--- a/xep-0387.xml
+++ b/xep-0387.xml
@@ -266,9 +266,9 @@
         <tr>
           <td><strong>Persistent Storage of Private Data via PubSub</strong></td>
           <td align='center'>&#10005;</td>
-          <td align='center'>N/A</td>
+          <td align='center'>&#10005;</td>
           <td align='center'>&#10003;&component;</td>
-          <td align='center'>N/A</td>
+          <td align='center'>&#10003;</td>
           <td>&xep0223;</td>
         </tr>
         <tr>
--- a/xep-0387.xml
+++ b/xep-0387.xml
@@ -2,10 +2,12 @@
 <!DOCTYPE xep SYSTEM 'xep.dtd' [
 <!ENTITY % ents SYSTEM 'xep.ent'>
 %ents;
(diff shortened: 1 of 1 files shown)


xep-0387.xml
```diff
@@ -244,6 +263,22 @@
           <td align='center'>&#10003;</td>
           <td>&xep0048;</td>
         </tr>
+        <tr>
+          <td><strong>Persistent Storage of Private Data via PubSub</strong></td>
+          <td align='center'>&#10005;</td>
+          <td align='center'>N/A</td>
+          <td align='center'>&#10003;&component;</td>
+          <td align='center'>N/A</td>
```
ge0rg: The "advanced" client needs to be able to read and store PEP bookmarks as well. I'm not sure the regular PEP support from Core is sufficient for that, or if we should actually make a checkmark here.

horazont:
@SamWhited Please comment on this so it can be merged before the council meeting agenda deadline.

(Or merge it yourself :-))

SamWhited:
As the author I am absolutely against merging this until we do the 2019 suites, as I said before.

dwd:
I don't think we'll get to publish it unless the comments are addressed somehow.

SamWhited:
They will be handled in the 2019 ones. These don't need to be perfect, and there is no point to doing compliance suites if we can't be on time with them so as mentioned before I don't want to encourage us to submit feedback after the deadline is over. This should have all been in during the last call and not had a second last call used as an excuse to submit things late.

dwd:
That was mandated by XEP-0001. We literally had no choice there. We can get this published tomorrow if this PR is either merged, or the comments addressed in some other way. Otherwise I don't see how a vote will pass.

SamWhited:
Retracting this (#573) since I'm tired of working on an XEP with constantly shifting goal posts. If someone else wants to take this up and work on the compliance suites, please re-open this PR.

PR #579, opened 2018-02-02
XEP-0198: Stream should be closed when 'h' value is to high.

guusdk:


horazont:
I’d suggest to add the specific type of stream error which SHOULd be emitted. I can’t find something sensible in RFC 6120, so maybe undefined-condition + a stream management specific error condition.

dwd:
In my view as author, I think this is valuable and a sensible addition, but needs list discussion (particularly for what error should be returned).

horazont:
Right, list discussion probably makes more sense than council for now.

horazont:
@guusdk Can you take this to the list please?

horazont:
@guusdk could you update your PR with the error conditions suggested on-list so that we can send this on to council?

Flowdalic:
Also could you please introduce a stream management specific stream error condition so that the stream error receiving entity could disable stream management in subsequent connections, avoiding reconnection loops.

xep-0198.xml
```diff
@@ -388,7 +399,12 @@
 </failed>
 ]]></example>
   <p>In both of these failure cases, the server SHOULD allow the client to bind a resource at this point rather than forcing the client to restart the stream negotiation process and re-authenticate.</p>
-  <p>If the former stream is resumed and the server still has the stream for the previously-identified session open at this time, the old stream SHOULD be terminated.</p>
+  <p>If the former stream is resumed and the server still has the stream for the previously-identified session open at this time, the server SHOULD send a 'conflict' stream error and close that stream</p>
+<example caption='Server returns error if client attempts to enable stream management before resource binding'><![CDATA[
```
horazont: I think this could use a paragraph of text, otherwise the example kinda comes out of nowwhere

xep-0198.xml
```diff
@@ -409,6 +425,13 @@
 </failed>
 ]]></example>
   <p>Stream management errors SHOULD be considered recoverable; however, misuse of stream management MAY result in termination of the stream.</p>
+  <p>When a remote entity acknowledges that it has handled a number of stanzas that is higher than the amount of stanzas that it was sent (by sending an 'h' value that is to high), the local entity SHOULD generate a stream error and close the stream:</p>
+  <example caption='Server closes old stream that was replaced by new stream'><![CDATA[
+<stream:error>
```
Flowdalic: As it was pointed it <conflict/> may not be the best choice. I'm also still missing the stream management specific stream error.

```
<stream:error>
      <undefined-condition
          xmlns='urn:ietf:params:xml:ns:xmpp-streams'/>
      <text xml:lang='en' xmlns='urn:ietf:params:xml:ns:xmpp-streams'>
        You acknowledged 42 stanzas while I only send you 12 so far.
      </text>
      <reported-handled-count-to-high xmlns='urn:xmpp:sm:3'/>
</stream:error>
```

xep-0198.xml
```diff
@@ -409,6 +425,13 @@
 </failed>
 ]]></example>
   <p>Stream management errors SHOULD be considered recoverable; however, misuse of stream management MAY result in termination of the stream.</p>
+  <p>When a remote entity acknowledges that it has handled a number of stanzas that is higher than the amount of stanzas that it was sent (by sending an 'h' value that is to high), the local entity SHOULD generate a stream error and close the stream:</p>
+  <example caption='Server closes old stream that was replaced by new stream'><![CDATA[
```
Flowdalic: Is the caption text right? Shouldn't it be something like "Closing a stream because a remote entity reported a impossible number of handled stanzas".

xep-0198.xml
guusdk: I actually made a minor change in the line preceding the example. I feel that more text to illustrate the error puts the thing out of proportion.

xep-0198.xml
guusdk: I believe I messed up a copy/paste somewhere. I'll correct it.

xep-0198.xml
guusdk: Same copy/paste error. This should've gone elsewhere.

xep-0198.xml
```diff
@@ -409,6 +426,17 @@
 </failed>
 ]]></example>
   <p>Stream management errors SHOULD be considered recoverable; however, misuse of stream management MAY result in termination of the stream.</p>
+  <p>When a remote entity acknowledges that it has handled a number of stanzas that is higher than the amount of stanzas that it was sent (by sending an 'h' value that is to high), the local entity SHOULD generate a stream error and close the stream:</p>
+  <example caption='Entity closes stream because peer acknowledges more stanzas than it was sent'><![CDATA[
+<stream:error>
+  <undefined-condition xmlns='urn:ietf:params:xml:ns:xmpp-streams'/>
+  <handled-count-too-high xmlns='urn:xmpp:sm:3' h='10' send-count='8'/>
+  <text xml:lang='en' xmlns='urn:ietf:params:xml:ns:xmpp-streams'>
+    You acknowledged 10 stanzas, but I only send you 8 so far.
```
horazont: s/send/sent/

horazont:
From what I can tell, this now reflects the list discussion and we can send this to council.

xep-0198.xml
```diff
@@ -28,6 +28,12 @@
   &fabio;
   &dcridland;
   &mwild;
+  <revision>
+    <version>1.5.4</version>
```
horazont: I think this should bump the minor and not the patch level version number

Flowdalic:
I just saw that the text does not specify how the stream should be closed not is <handled-count-to-high/> defined. This should probably be changed before being merged.

horazont:
@Flowdalic Good catch about the schema. @guusdk could you fix that, too?

guusdk:
> I just saw that the text does not specify how the stream should be closed not is defined.

Que?

Flowdalic:
> 
I just saw that the text does not specify how the stream should be closed nor is <handled-count-to-high/> defined.

Sorry, github swallowed the XML. The issue is that we do not consider examples as normative, hence we need to specify the behavior once this happens and we need to specify the <handled-count-to-high/> application-specific stream error information element

guusdk:
That makes sense. I've incorporated the changes now.

guusdk:
Applied feedback from Council (too -> to), rebased.

linkmauve:
Fixed a few more typos.

PR #582, opened 2018-02-06
XEP-0060: `retrieve-default-sub` feature missing in XML Schema.

sco0ter:


sco0ter:
Please not that this feature is also missing in the registrar:
https://xmpp.org/registrar/disco-features.html

PR #585, opened 2018-02-12
XEP-0401: Use Data Form for IBR and fix error stanzas

af8a524db1:
* Use Data Form for In-Band Registration as required by XEP-0077
 * Fix error stanzas as required by RFC 6120 § 8.3.2
 * Change error of "invalid-token" error to "modify"

### DIFF
--- a/xep-0401.xml
+++ b/xep-0401.xml
@@ -31,6 +31,16 @@
     <surname>Schink</surname>
   </author>
   <revision>
+    <version>0.2.0</version>
+    <date>2018-02-11</date>
+    <initials>ms</initials>
+    <remark>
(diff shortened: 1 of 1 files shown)


Flowdalic:
Doesn't this require a namespace bump since the registration fields are now returned in a form?

dwd:
That'd be my thought too. (But note it's the Editor's role to do that as needed).

Also, I'd highly recommend raising this change on the mailing list, if only to get some conversation going about the spec.

horazont:
Removing ready to merge since @ge0rg has feedback he wants to send.

ge0rg:
Pinged the list at https://mail.jabber.org/pipermail/standards/2018-February/034323.html

horazont:
Discussion thread is https://mail.jabber.org/pipermail/standards/2018-February/034315.html

PR #591, opened 2018-02-21
XEP-0050 Ad-Hoc Commands: Clarify 'execute' actions equivalence.

Flowdalic:
The XEP is slightly ambiguous regarding the equivalence of the
'execute' action, it states that:

- The action "execute" is always allowed, and is equivalent to the
  action "next".
- The "next" action is typically the "next" button or option in a
  wizard. If <next/> is not contained by the <actions/>, it is
  disabled.

So if "next" action is disabled, execute is allowed but equivalent to
a disabled action.

This change tries to clarify that.

### DIFF
--- a/xep-0050.xml
+++ b/xep-0050.xml
@@ -27,7 +27,7 @@
   &linuxwolf;
   <revision>
     <version>1.2.3</version>
-    <date>2018-02-21</date>
+    <date>2018-04-19</date>
     <initials>fs</initials>
     <remark>Clarify 'execute' actions equivalence.</remark>
   </revision>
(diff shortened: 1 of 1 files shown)


xep-0050.xml
```diff
@@ -520,7 +526,7 @@
       </li>
       <li>If there is an &lt;actions/&gt; element, the user-agent usually uses a multi-stage dialog or view, such as a wizard.
         <ul>
-          <li>The action "execute" is always allowed, and is equivalent to the action "next".</li>
+          <li>The action "execute" is always allowed, and is equivalent to the action "next" if it is an allowed action, or otherwhise equivalent to "complete".</li>
```
horazont: There’s a typo ("otherwhise") (which you can either fix or let the editor fix).

Flowdalic:
Fixed the typo, thanks @horazont 

Also here is a link to the related standards@ thread: https://mail.jabber.org/pipermail/standards/2018-February/034324.html

horazont:
From my understanding, #598 supersedes this. Re-open if I’m wrong.

Flowdalic:
I personally would prefer this over #598, as it's shorter and not so complex.

#591 (this PR) simply defines a sane default value for 'execute' if 'next' is not allowed
#598 forbids 'execute' to point to a invalid action

Both changes will probably cause some implementations to violate the rules, but only #591 allows initiating entities to simply step through the ad-hoc command using 'execute'.

horazont:
I’ll re-open and put this on councils plate to decide.

SamWhited:
Closing this for now to clean up the "Needs Council" label; council vetoed a few months ago and discussed rewording to make the intention clear. If the author would like to submit a new draft or go on list for further feedback and discussion feel free. Thanks!

PR #593, opened 2018-02-26
XEP-0153: Clarify encoding of update hash

horazont:
As per list discussion https://mail.jabber.org/pipermail/standards/2018-February/034362.html .


stpeter:
LGTM as author of XEP-0153.

dwd:
Added to agenda for 2018-03-07

PR #596, opened 2018-03-05
XEP-0390: Explain that ASCII Seperators are invalid only in XML *1.0*

Flowdalic:
but not in XML 1.1.


Ppjet6:
@horazont thoughts?

This looks like a harmless clarification to me.

horazont:
LGTM indeed, I postponed this because I wanted to do things, but then I got sidetracked. let’s get this merged.

PR #598, opened 2018-03-05
XEP-0050. Try to clarify usage of 'execute'

Kev:


xep-0050.xml
```diff
@@ -520,17 +520,19 @@
       </li>
       <li>If there is an &lt;actions/&gt; element, the user-agent usually uses a multi-stage dialog or view, such as a wizard.
         <ul>
-          <li>The action "execute" is always allowed, and is equivalent to the action "next".</li>
+          <li>The action "execute" is always allowed, and is equivalent to the action "next". A form that has an &lt;actions/&gt; element, no "execute" attribute, and for which &lt;next/&gt; is not an available action is therefore invalid.</li>
```
horazont: Okay, I might miss context here. But I suggest this wording instead:

> The action "execute" is equivalent to the action specified in the "execute" attribute. If the "execute" attribute is absent, it defaults to "next". A form which has an ``<actions/>`` element and an "execute" attribute which evaluates (taking the default into account if absent) to an action which is not allowed is therefore invalid.

xep-0050.xml
```diff
@@ -520,17 +520,19 @@
...
           <li>The "complete" action is typically the "finish" or "done" button or option in a wizard. If &lt;complete/&gt; is not contained by the &lt;actions/&gt;, it is disabled.</li>
           <li>If the &lt;actions/&gt; possesses the "execute" attribute, that value is the default button or option. If the &lt;actions/&gt; does not possess the "execute" attribute, there is no default button or option.</li>
+          <li>As the "execute" command is equivalent to another command ("next" in the absence of an "execute" attribute), it is not suggested that a user-agent shows both the "execute" and the command that "execute" is an alias of in their interface.</li>
         </ul>
       </li>
     </ol>
     <p>Responders SHOULD use the following guidelines when providing &lt;actions/&gt;:</p>
     <ul>
       <li>The "execute" attribute SHOULD NOT specify a value that does not match one of the allowed actions.</li>
+      <li>As user-agents treat the absence of an "execute" attribute as equivalent to "next", when the &lt;actions/&gt; element is provided, it is invalid to return a command that has no "execute" attribute, an actions element, and no &lt;next/&gt; action</li>
```
Flowdalic: It appears that this just duplicates what is added above. If so, I'd suggest to only state the requirement once.

horazont:
@Kev Will you update the PR?

Kev:
Updated, thanks for the wording. For the semi-duplicated text - I think it's worth leaving in, given the confusion that this has obviously caused, can see what Council think and tweak accordingly.

horazont:
Closing this. If you want to re-submit this proposal for the current council, please comment / reopen.

PR #603, opened 2018-03-12
Fix which witch is which

Kev:


dwd:
Best commit message ever.

Also I concur with the change, and consider it editorial.

PR #604, opened 2018-03-12
XEP-0163: Clarify how +notify works

Flowdalic:
It sounded like annoucing the jabber.org/protocol/foo+notify feature
causes one to receive notifications from *all* nodes related to the
jabber.org/protocol/foo protocol, but as per XEP-0060 § 9.2, one will
only receive notifications of the node whose NodeID equals
'jabber.org/protocol/foo'.


ralphm:
This is indeed a bit clearer. Thanks!

PR #608, opened 2018-03-15
XEP-0223: Add a warning about publish-options support

horazont:
In light of [CVE-2018-6591](https://gultsch.de/converse_bookmarks.html), it seems reasonable to make this clearer.

Variants (Council, feel free to mix and match):

* Raise the discovery to a MUST.
* Move "Determining Support" before the examples of the publish flow.

### DIFF
--- a/xep-0223.xml
+++ b/xep-0223.xml
@@ -26,6 +26,12 @@
   <shortname>N/A</shortname>
   &stpeter;
   <revision>
+    <version>1.1</version>
+    <date>2018-03-28</date>
+    <initials>jwi</initials>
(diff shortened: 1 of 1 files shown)


Kev:
This should really talk about pubsub, not PEP.

dwd:
Council decision here is to apply this but with the discovery raised to MUST and s/PEP/pubsub/

I'm holding a -1 until these are done for form's sake, so nudge me when those changes are done and I can change my vote.

horazont:
@dwd Changes applied, please review.

dwd:
In case that wasn't clear, you can merge this one @horazont

PR #611, opened 2018-03-18
bookmarks2: remove password element, add examples, update security considerations

jcbrand:
Password removal as suggested by @SamWhited in the list.
Updated security considerations as suggested by @iNPUTmice 

@dwd What do you think?


SamWhited:
Wih my editor hat on I'd prefer to merge changes while it's in the protoxep state. It's not likely that this would cause anyone to veto either way, and it's easier to make changes before we have to start versioning things.

PR #644, opened 2018-05-22
XEP-0373: Add pubsub example

vanitasvitae:
This PR adds an example about how the metadata node is being fetched.
It also fixes a small typo in one of the anchors.

### DIFF
--- a/xep-0373.xml
+++ b/xep-0373.xml
@@ -281,7 +281,7 @@
   <p>Two PEP node types are invovled: A "medatata node" is used to store meta information about
   OpenPGP keys used by an entity while the actual public keys are stored in "data nodes".</p>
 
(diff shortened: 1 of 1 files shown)


vanitasvitae:
8c752ba is probably a breaking change.

horazont:
@vanitasvitae Indeed, 8c752ba is breaking. Why did you include it in this PR? What is the rationale?

horazont:
@Flowdalic please look at this

vanitasvitae:
> Why did you include it in this PR? What is the rationale?

Basically it was triggering my OCD :wink:. Should I maybe exclude this change for now, so that it can be included in a later (breaking) change?

vanitasvitae:
I split this PR and moved the breaking change (8c752ba) to #669

PR #648, opened 2018-05-26
Fix a typo in XEP-0334

labdsf:


labdsf:
Judging from the commit history in this repository, adding typo fixes to revision history is not required.

PR #650, opened 2018-05-29
XEP-0157: fix data form example

horazont:


Kev:
No. As discussed in xsf@, it's not actually desirable.

PR #653, opened 2018-06-02
XEP-0045: Add a feature for the voice request flow

horazont:
via @Zash


Flowdalic:
Can we make the feature name more explicit, e.g., muc#voice_request?

horazont:
@Flowdalic Using ``#request`` is in line with the FORM_TYPE used for the request. Technically, it could also be used to request moderator privileges since the role which the occupant wants is a field in the form.

horazont:
Veto-ed.

PR #655, opened 2018-06-05
XEP-0369 MIX-CORE update to sort JID addressing

stevekille:


xep-0369.xml
```diff
@@ -580,28 +588,31 @@
...
       </p>
       <p>
-       A Nick MAY be associated with a participant.  The nick associated with the user is stored in a &lt;nick/&gt; child element of the &lt;participant/&gt; element.   The nick for each channel participant MUST be different to the nick of other participants.
+       A Nick MAY be associated with a participant, which provides a user-oriented  description of the participant.  The nick associated with the user is stored in a &lt;nick/&gt; child element of the &lt;participant/&gt; element.   The nick for each channel participant MUST be different to the nick of other participants.
```
Kev: I'm not convinced Nicks should be optional (I thought we'd come to that conclusion on-list).

xep-0369.xml
```diff
@@ -580,28 +588,31 @@
...
+      </p>
+      <p>
+        Where a nick is provided for a user, it is generally recommended to use this nick or the JID to display the user.   This enables consistent representation of participants for all participants in the channel.
```
Kev: Surely we should only recommend showing the nick? JIDs should be hidden from users whenever possible, and in this case a client showing JIDs might have unpleasant interactions in anonymous rooms.

xep-0369.xml
```diff
@@ -580,28 +588,31 @@
...
       </p>
       <p>
         The real JID of the user MAY be held in the participants node.   When the real JID is not present, the procedures defined in MIX-ANON must be followed.
         The user's JID is stored in a &lt;jid/&gt; child element of the &lt;participant/&gt; element.
       </p>
       <p>
-        When a user joins a channel, the user's bare proxy JID is added to the participants node by the MIX service.   When a user leaves a channel, the user's bare proxy JID is removed from the participants node.  The participants node MUST NOT be directly modified using pubsub.
+        When a user joins a channel, an item representing the is added to the participants node by the MIX service.   When a user leaves a channel, the user's item is removed from the participants node.  The participants node MUST NOT be directly modified using pubsub.
```
Kev: 'an item representing the is added'?

xep-0369.xml
```diff
@@ -668,7 +679,7 @@
   <ol>
     <li>IQ.  All of the IQ errors of &rfc6120; MUST be supported.</li>
     <li>Messages.  If a message is received and an error situation is encountered, a message of type error MUST be sent back to the message sender. This message format is specified in &rfc6121; containing an error defined in &rfc6120;, which is the same error set as for IQs.</li>
-    <li>Presence.  Any responses to presence messages MUST follow the rules of &rfc6121;.</li>
+    <li>Presence. If MIX-PRESENCE is supported,  any responses to presence messages MUST follow the rules of &rfc6121;.</li>
```
Kev: There's no such thing as a "presence message".
It's not clear to me what following the rules of 6121 would mean in this case?

xep-0369.xml
```diff
@@ -997,7 +1010,7 @@
...
 </message>
 ]]></example>
-      <p>The user that has been added to the channel is identified by the item id of the item added to the Participants node, which is the proxy JID of the new channel participant.  The &lt;participant&gt; element MUST include a &lt;jid&gt; element with the JID of the participant, unless MIX-ANON is being followed to hide participant JIDs.    The &lt;nick&gt; element will not be included at this point, unless it is automatically generated by the channel.   In the likely event that a Nick is subsequently added, an update with the &lt;nick&gt; element will be distributed.
+      <p>The user that has been added to the channel is identified by the item id of the item added to the Participants node, which is the Stable Participant ID of the new channel participant.  The &lt;participant&gt; element MUST include a &lt;jid&gt; element with the JID of the participant, unless MIX-ANON is being followed to hide participant JIDs.    The &lt;nick&gt; element will not be included at this point, unless it is automatically generated by the channel.   In the likely event that a Nick is subsequently added, an update with the &lt;nick&gt; element will be distributed.
```
Kev: I thought the plan was that you specified your nick when joining? Having a situation where participants are nick-less because their join processing didn't complete seems unhelpful.

xep-0369.xml
```diff
@@ -1132,9 +1145,11 @@
       </p>
       <ol>
         <li>A &lt;nick&gt; element that contains the Nick of the message sender, taken from the Participants Node.  This MUST be present if a Nick is defined for the user.</li>
-        <li>A &lt;jid&gt; element containing the real JID of the sender. This MUST be present, unless following the "JID Hidden" model defined in MIX-ANON.  If this element is omitted, the &lt;nick&gt; and &lt;proxy-jid&gt; elements MUST be present.</li>
-        <li>A &lt;proxy-jid&gt; element contains the sender's proxy JID.  This MUST be present if &lt;jid&gt; is omitted and MUST not be present if &lt;jid&gt;  is present.</li>
+        <li>A &lt;jid&gt; element containing the real JID of the sender. This MUST be present, unless following the "JID Hidden" model defined in MIX-ANON.  If this element is omitted, the &lt;nick&gt; element MUST be present.</li>
       </ol>
+      <p>
+        MIX messages are distributed by the channel with the from using the JID of the channel, with the Stable Participant ID of the sender in the resource.    This enables a receiving system to distinguish messages based on sender using on the JID.
```
Kev: Using on...ly?

xep-0369.xml
Kev: <Steve> I can see sensible applications for MIX, for example IoT, where Nicks are not sensible.

xep-0369.xml
Kev: &lt;Steve&gt; The JID will be known to all. I can see some situations where you’d prefer to use the definitive name. The key point is to not use local references. I’m fine to delete “or the JID” if you prefer.

stevekille:
I’m not sure what you are saying here or what change you want

 


S

xep-0369.xml
Kev: &lt;Steve&gt; Fixed

xep-0369.xml
Kev: &lt;Steve&gt; Corrected

stevekille:
Just figured it.   These are MY comments under @kev

xep-0369.xml
Kev: &lt;Steve&gt; The protocol has setting Nicks independent of Join (it has always been this way). I don’t recollect any plan to change this.

xep-0369.xml
Kev: &lt;Steve&gt; Fixed

xep-0369.xml
Kev: So make it a "MUST except..."? Them being optional seems very unhelpful for the majority of cases.

xep-0369.xml
Kev: I think that would be better.

xep-0369.xml
Kev: > I’m not sure what to write here.

What is the text trying to say?

xep-0369.xml
Kev: 0.1 had a note to include the nickname in the join, but the text wasn't yet written.

Kev:
> In MIX-PRESENCE. Core does not need them.

I don't believe that's true.

PR #656, opened 2018-06-05
XEP-403 MIX-PRESENCE  - updates for addressing and IQ relay

stevekille:


xep-0403.xml
```diff
@@ -61,32 +76,49 @@
     <li>The mechanism must work cleanly for participants with multiple clients.</li>
     <li>Standard presence messages must be used to share presence.</li>
     <li>Nick changes should be visible as changes (and not as a new user).</li>
-    <li>Where &xep0404; is not used,  participants must be able to directly contact other participants.</li>
+    <li>Where &xep0404; (MIX-ANON) is not used,  participants must be able to directly contact other participants.</li>
```
Kev: This says must, but earlier it was optional.

xep-0403.xml
```diff
@@ -61,32 +76,49 @@
...
+    <li>IQ stanzas, including vCard, can be communicated through a channel.</li>
```
Kev: Is saying 'through a channel' potentially confusing?

xep-0403.xml
```diff
@@ -61,32 +76,49 @@
...
 
   </ol>
 
 </section1>
 
 <section1 topic='Concepts' anchor='concepts'>
+  
+ <section2 topic="Participant Information in Presence" anchor="concept-participant-info">
+   <p>
+     A MIX channel MAY distribute presence information about channel participants.   In order to share JID and Nick information about a participant, this information is encoded in the presence message.   This allows full presence information to be shared for each participant without the need for the client to perform any lookup.
```
Kev: Is this strictly OPTIONAL? I think if someone's implementing 403 it's actually mandatory isn't it?

xep-0403.xml
```diff
@@ -61,32 +76,49 @@
...
+  
 
-
+<section2 topic="Partipant JID Addressing for Presence and IQ relay" anchor="concept-jid-address">
+  <p>
+    Participants in Presence messages and in IQ messages relayed through a channel are identified by an encoded JID of the form are encoded, using the format "&lt;stable-participant-identifier&gt;#&lt;channel&gt;@&lt;mix domain&gt;".   This provides a stable JID for each participant. This JID format is used in three places:
+  </p>
+  <ol>
+    <li>The 'from' of presence stanzas generated by the channel.</li>
+    <li>The 'to' of IQ stanzas being sent to the channel.</li>
```
Kev: This is wrong isn't it? If you're sending to the channel you're using the channel's JID.

xep-0403.xml
```diff
@@ -61,32 +76,49 @@
...
+    <li>The 'from' of IQ stanzas coming from the channel.</li>
```
Kev: Same.

xep-0403.xml
```diff
@@ -61,32 +76,49 @@
...
+  </ol>
+  <p>
+    These JIDs will be used to represent specific JID clients.   The resource associated with the encoded JID MUST be the resource value from the associated client JID, except where MIX-ANON is followed.
```
Kev: I think this is wrong and we always use a client-id map. This allows sensible JID shortening, otherwise you'll break with long resources.

xep-0403.xml
```diff
@@ -61,32 +76,49 @@
...
+  </p>
+</section2>
+  
+  
   <section2 topic="User Presence in MIX" anchor="concepts-presence">
     <p>
-      MIX channels store presence of each online client for a user that chooses to publish presence.  Presence is stored in the presence node and is encoded using a full proxy JID.  Where a user publishes presence for one or more clients, this information is available to all users subscribing to the channel presence.
+      MIX channels store presence of each online client for a user that chooses to publish presence, in the format that is distributed as presence.  Presence is stored in the presence node with an item for each client that is publishing presence.  Each item is named with an encoded JID and contains the presence information that is distributed.    Where a user publishes presence for one or more clients, this information is available to all users subscribing to the channel presence.
```
Kev: Do we need to use the encoded JID for ids, or is it sufficient/better to just do user-id/client-id or something?

Kev: Maybe there's merit in it being the same as the from of the live updates, but I'm not yet sure.

xep-0403.xml
```diff
@@ -126,16 +158,27 @@
...
+        The presence node contains the presence value for clients belonging to participants that choose to publish presence to the channel. A MIX channel MAY require that all participants publish presence, so that active channel participants are visible.  It is not possible to enforce this in the server, so participants in a channel with this option MUST publish presence.  Each item in the presence node is identified by an encoded JID. The presence is encoded as a standard a presence stanza using a &lt;presence/&gt; element qualified by the 'jabber:client' namespace.
       </p>
-
+    <p>
+      MIX extends the &lt;presence&gt; stanza using a &lt;mix&gt; element qualified by the 'urn:xmpp:mix:presence:0' namespace.  This enables any receiver of presence  to see identify the client to which presence refers and to have a nick to display.  This element contains two child elements:
```
Kev: See identify?

xep-0403.xml
```diff
@@ -207,64 +254,143 @@ A user MAY share presence information with the channel, for one or more online c
...
+      </mix>
+</presence>]]></example>
 
+    </section2>
 
-    </section3>
 
+<section2 topic="Relaying IQ Stanzas" anchor="usecase-iq-relay">
+  <p>
+    MIX channels MAY relay IQ stanzas between participants.   This is often useful to obtain client information where a direct request to the client would be blocked.   When a client sends an IQ stanza through a MIX channel, it will set the 'from' to its own JID and set the 'to' to the encoded JID of the recipient.    The MIX channel will modify the JIDs in the outgoing message, so that the 'to' is the full JID of the recipient and the 'from' is the encoded JID of the sender.   This is illustrated in the vCard section below.
```
Kev: I think this elides that there is both bare-JID (which should probably be in 369) and full-JID routing.

xep-0403.xml
```diff
@@ -207,64 +254,143 @@ A user MAY share presence information with the channel, for one or more online c
...
+  </p>
+</section2>
+  
+  
+  
+  <section2 topic="Requesting a vCard through a Channel" anchor="usecase-vcard">
+    <p>A client  MAY request the vCard of a channel participant through a MIX channel, for example to get an avatar.  The MIX channel MAY pass this request on or MAY block it.  vCard requests MAY use &xep0054; (vcard-temp) or &xep0292; (vCard4 over XMPP).  The MIX channel does not process the vCard requests, but simply relays them on to real bare JID of the target.  A MIX service MAY choose to relay one or both protocols.   Where a MIX service relays  one or both of these protocols, each protocol relayed MUST be advertised as a feature of the MIX service.   In the following example, using vcard-temp, the requesting client sends a message to the  encoded JID of the channel participant for which the vCard is desired.</p>
```
Kev: This seems out of place. The stuff about advertising vcard support seems wrong (you advertise vcards if you support vcards), and it's just a case of forwarding iqs, so this seems to be saying more than is needed. There's also more MAYs in here than there should be, I think. I don't think this is optional-to-implement.

xep-0403.xml
```diff
@@ -207,64 +254,143 @@ A user MAY share presence information with the channel, for one or more online c
...
+    <example caption="Channel passes on vCard request to the User&apos;s Server" ><![CDATA[
+<iq from='123456#coven@mix.shakespeare.example/6789'
+    id='lx09df27'
+    to='peter@shakespeare.example'
+    type='get'>
+  <vCard xmlns='vcard-temp'/>
+</iq>
+]]></example>
+    <p>
+      The user's server, on behalf of the user, MAY send a response or reject with an error.   The user's server will send the vCard back to the channel.
```
Kev: This isn't right, it's an iq so the server MUST reply or error.

horazont:
Could you please keep technical discussion on standards@? This is extremely hard to read both on the email and the web interface, and not everyone potentially interested is subscribed to this.

stevekille:
Jonas,

From my perspective,  email is best.   To stadards@ for substantive stuff.   Nits to me, so I can fix them.

I responded to the Git email (not realizing this was bad), which messed things up further.

My plan is to address issues in updates.

Steve

Kev:
> Could you please keep technical discussion on standards@?

You know my opinion on stuff happening on github instead of the lists, but Steve asked me to comment on these reviews, so I did. It is at least largely just clarification of stuff that we've already discussed, rather than whole new discussions.

PR #657, opened 2018-06-05
XEP-405 MIX-PAM editorial to align with MIX-CORE/MIX-PRESENCE

stevekille:


Kev:
I think this also needs a note that if the server does blocking of unknown JIDs it'll potentially want to allow through encoded JIDs for channels in which the user resides.

PR #660, opened 2018-06-06
MIX-ANON Clarify use of PMs &  Editorial

stevekille:


horazont:
Going to bump the version number to 0.3 for non-editorial changes.

PR #667, opened 2018-06-18
XEP-0045: Added standardized roominfo fields for allowinvites and the…

iNPUTmice:
… room name.

The standard doesn’t commend muc services to set the identity name of a room to be the name configured in the room configuration. Some implementations set the identity name to the local part of the jid if the configured room name is empty. Therefore implementations are unable to discover if a room name is empty (=not set) when running disco#info. To circumvent this and to let clients discover (reliably) if a name is set (or not!) this PR adds a standardized roominfo field for the name.

Discovering whether or not an occupant is allowed to invite other members is also useful (clients can hide the button if that’s not the case.)

Rendered version here: https://gultsch.de/files/xep-0045.html

### DIFF
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -46,6 +46,14 @@
   <registry/>
   &stpeter;
   <revision>
+    <version>1.32.0</version>
+    <date>2018-06-18</date>
+    <initials>dg</initials>
+    <remark>
+      <p>Added standardized roominfo fields for allowinvites and the room name.</p>
+    </remark>
+  </revision>
+  <revision>
     <version>1.31.1</version>
     <date>2018-03-12</date>
     <initials>kis</initials>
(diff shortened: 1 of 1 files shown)


zinid:
For the record, seems like `muc#roominfo_changesubject` and `muc#roominfo_pubsub` is missing in the form definition, but they present in Example 10.

zinid:
> There is muc#roomconfig_changesubject in the same example a few lines bellow which serves the same purpose and is specified.

Indeed, LOL

zinid:
> Also they don’t follow the same nameing scheme. For example there is roominfo_description and roomconfig_roomdesc

Don't we increase the discrepancy by introducing `muc#roominfo_name`? Because there is `muc#roomconfig_roomname` in the document. This is sort of bikeshedding of course, but may lead to copy-paste "typo" bugs in implementations.

iNPUTmice:
Yes we are. I didn’t get that you can simply put roomconfig things in there last night.
I should probably close this PR.

PR #671, opened 2018-06-28
XEP-0198: Clarify the note about counters

Flowdalic:
The note was only 50% correct (the counter the note talks get
initialized and started when the entity receives <enable(d)/>, and not
when it is sent) and partly misleading (there are typically four
counters per stream). Most XEP-0198 Stream Management implementations
I am aware of didn't get the counters right at the first time. The
tricky part is to get the time right where the counters should be set
to zero and get started. The new note's text tries to help
implementers with this. As bonus, it is also shorter than the original
note.

### DIFF
--- a/xep-0198.xml
+++ b/xep-0198.xml
@@ -29,6 +29,12 @@
   &dcridland;
   &mwild;
   <revision>
+    <version>1.5.3</version>
+    <date>2017-06-27</date>
+    <initials>fs</initials>
+    <remark><p>Improve the note about stream management counters in section 4.</p></remark>
+  </revision>
+  <revision>
     <version>1.5.2</version>
     <date>2016-12-08</date>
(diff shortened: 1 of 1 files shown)


mwild1:
Would it be less confusing to just say "Each entity maintains two counters"? People are typically going to be focused on what they need to store in the implementation they are currently writing, and saying there are four counters may still be confusing (I see why you wrote it though - because the original text also took the "view from both sides" approach).

Flowdalic:
Possible. On the other hand, the "Every side of the stream maintains a counter of its own send stanzas and a counter of the received and handled stanzas ('h')." is pretty close to that. That doesn't mean I wouldn't be looking forward for you patch, just that I personally see no need to change the wording.

mwild1:
> That doesn't mean I wouldn't be looking forward for you patch, just that I personally see no need to change the wording.

Submitted, see #673.

PR #672, opened 2018-06-28
XEP-0059: Add 'exact(-index)' attribute to RSM's <count/> and <first/>

Flowdalic:
RSM specifies that the number or position included in a RSM response
may be approximate. But some protocols benefit from knowledge whether
the value is exact or approximate. For example a MAM synchronization
algorithm, which syncs a slave archive, potentially having multiple
holes, from a master archive, may use an optimized algorithm if the
master archive returns exact counts.

This commit adds signaling to RSM that the values are exact and not
approximate in a backwards compatible fashion.

### DIFF
--- a/xep-0059.xml
+++ b/xep-0059.xml
@@ -34,6 +34,12 @@
     <jid>jlseguineau@im.antepo.com</jid>
   </author>
   <revision>
+    <version>1.1</version>
+    <date>2018-06-2f</date>
+    <initials>fs</initials>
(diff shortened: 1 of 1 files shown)


Flowdalic:
I don't think that "Needs Board" is correct as this is not a procedural XEP. Should be "Needs Council".

jcbrand:
You're right, sorry.

horazont:
[Vetoed in Council meeting on 2018-07-11](https://mail.jabber.org/pipermail/standards/2018-July/035245.html).

PR #674, opened 2018-06-29
XEP-0313 State explicitly that there shall be no 'to' attribute

bartekgorny:
Previous phrasing suggested that it might be present and equal to "", which is not legal for a jid. See convo on mailing list.


jcbrand:
Having read the mailing list discussion, it's not clear to me whether board still needs to approve.

Mailing list discussion starts here: https://mail.jabber.org/pipermail/standards/2018-June/035182.html

jcbrand:
Ok, looking at the README, since this XEP is still experimental and the change has been discussed in the mailing list, it may be merged without approval by standards body.

Kev:
Somewhat belatedly as I didn't see this was here, but consider this author approval.

PR #677, opened 2018-07-19
Three small editorial fixes

linkmauve:
I didn’t make a release for the third commit since there is exactly no functional change, I can do one if you prefer.


horazont:
The last commit modifies the rendering, and it was argued recently that anything which changes how the XEP looks on the page needs a revision block (with a patch version bump).

linkmauve:
There, revision block added.

PR #681, opened 2018-07-22
XEP-0050: Remove the status attribute from the request

linkmauve:
Remove a suggestion to set the status attribute by the requester, since this attribute MUST be ignored by the responder if set (§4.1).


horazont:
Rejected because council, see https://mail.jabber.org/pipermail/standards/2018-August/035301.html

PR #682, opened 2018-07-25
XEP-0065, XEP-0363: Document exposing the service on the user’s domain

linkmauve:
This practice is quite widespread already, and shouldn’t have any
overhead given the domain’s disco#info should already be cached by
XEP-0115/XEP-0390 in the <stream:features/>.

### DIFF
--- a/xep-0065.xml
+++ b/xep-0065.xml
@@ -30,6 +30,16 @@
   &stpeter;
   &infiniti;
   <revision>
+    <version>1.9</version>
+    <date>2018-07-25</date>
+    <initials>egp</initials>
+    <remark>
+      <ul>
+        <li>Standardise the user’s domain exposing this feature.</li>
+      </ul>
+    </remark>
+  </revision>
+  <revision>
     <version>1.8.1</version>
     <date>2015-09-17</date>
     <initials>fs</initials>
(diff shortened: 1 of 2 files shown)


horazont:
@iNPUTmice Please comment for the change to XEP-0363.

xep-0363.xml
```diff
@@ -205,6 +215,40 @@
...
+  <p>Alternatively, the upload service may be exposed directly on the Requester’s server. In which case, the Requester should use Service Discovery on this domain, like for every previous disco#items results.</p>
+  <example caption='Requester Sends Service Discovery Request to Server'><![CDATA[
+<iq from='requester@example.com/foo'
+    id='a58f1cc2'
+    to='example.com'
+    type='get'>
+  <query xmlns='http://jabber.org/protocol/disco#info'/>
+</iq>
+]]></example>
+  <p>The server returns its information and the Requester inspects it to determine if it contains an identity of category "proxy" and type "bytestreams".</p>
```
lovetox: Copy paste error, `proxy`, `bytestream`

horazont:
Rejected by council due to wording, author of the other affected XEP does not reply.

lockywolf:
I would appreciate not needing a separate DNS entry and a tls setting for each component. It would be a good idea.

xep-0065.xml
```diff
@@ -271,6 +281,32 @@
     <feature var='http://jabber.org/protocol/bytestreams'/>
   </query>
 </iq>
+]]></example>
+  <p>Alternatively, the proxy may be exposed directly on the Requester’s server. In which case, the Requester should use Service Discovery on this domain, like for every previous disco#items results.</p>
```
Flowdalic, line 285: nit: a quick grep in the XEPs repo gives the impression that 'requester' is usually written lowercase

PR #686, opened 2018-07-31
XEP-0384: Make examples use item id='current'

linkmauve:
This makes it respect XEP-0060 §12.20 on singleton nodes.

### DIFF
--- a/xep-0384.xml
+++ b/xep-0384.xml
@@ -28,6 +28,12 @@
     <jid>andy@strb.org</jid>
   </author>
   <revision>
+    <version>0.2.2</version>
+    <date>2018-07-31</date>
+    <initials>egp</initials>
+    <remark><p>Make examples show items published to the id "current", as per <cite>XEP-0060</cite> §12.20.</p></remark>
+  </revision>
+  <revision>
     <version>0.2.1</version>
     <date>2018-05-21</date>
     <initials>mb</initials>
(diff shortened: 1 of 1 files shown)


horazont:
I don’t think this passes as patch level version bump. Also, @iNPUTmice please ack.

horazont:
Also, on a personal note, I’d appreciate if there was a paragraph giving a rationale for ``id="current"`` (by referencing the XEP-0060 wording on singleton nodes).

goffi-contrib:
A paragraph explaining that `current` **SHOULD** be used for `id` would be necessary IMHO.
I would use **SHOULD** instead of **MUST** to avoid namespace bump.

Flowdalic:
@goffi-contrib isn't that what

        <p>NOTE: as per <link url='https://xmpp.org/extensions/xep-0060.html#impl-singleton'><cite>XEP-0060</cite> §12.20</link>, it is RECOMMENDED for the publisher to specify an ItemID of "current" to ensure that the publication of a new item will overwrite the existing item.</p>

adds (Note that 'RECOMMENDED' equals 'SHOULD')?

goffi-contrib:
@Flowdalic yes it is, it seems that I've missed it :)

jcbrand:
@strb We need you as author to review this PR please.

strb:
I apologize for the delay, this somehow slipped through the cracks.

The change looks good to me. It doesn't really change the mechanics of anything, just makes things more explicit, so the patch-level change thing is fine IMO.

PR #690, opened 2018-08-02
XEP-0184: Make the schema require @id in <received/>

linkmauve:
Even though the text only says “[an entity] SHOULD include an 'id' attribute”, there is really nothing to do without one.

### DIFF
--- a/xep-0184.xml
+++ b/xep-0184.xml
@@ -28,6 +28,12 @@
   &stpeter;
   &hildjj;
   <revision>
+    <version>1.4.0</version>
+    <date>2018-08-02</date>
+    <initials>egp</initials>
+    <remark><p>Make the 'id' attribute required, this extension makes no sense otherwise.</p></remark>
+  </revision>
+  <revision>
     <version>1.3.0</version>
     <date>2019-05-15</date>
     <initials>gl</initials>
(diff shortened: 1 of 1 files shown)


ge0rg:
You are adding an inconsistency here. If the schema enforces the @id, then the text needs to be changed to "MUST". I think when this was discussed in the Council, there was a rough consensus that this change won't be blocked. Could you please change the wording accordingly?

horazont:
Rejected by council, because of the schema inconsistency. Feel free to update and resubmit.

linkmauve:
Sorry I missed your two messages.

I have now implemented this change, and hereby ask Council to review it again.

Thanks!

xep-0184.xml
```diff
@@ -203,7 +209,7 @@
   <received xmlns='urn:xmpp:receipts' id='richard2-4.1.247'/>
 </message>
 ]]></example>
-  <p>When the recipient sends an ack message, it SHOULD ensure that the message stanza contains only one child element, namely the &lt;received/&gt; element qualified by the 'urn:xmpp:receipts' namespace. In addition, it SHOULD include an 'id' attribute that echoes the 'id' attribute of the content message. Naturally, intermediate entities might add other extension elements to the message when routing or delivering the receipt message, e.g., a &lt;delay/&gt; element as specified in &xep0203;.</p>
+  <p>When the recipient sends an ack message, it SHOULD ensure that the message stanza contains only one child element, namely the &lt;received/&gt; element qualified by the 'urn:xmpp:receipts' namespace. In addition, it MUST include an 'id' attribute that echoes the 'id' attribute of the content message. Naturally, intermediate entities might add other extension elements to the message when routing or delivering the receipt message, e.g., a &lt;delay/&gt; element as specified in &xep0203;.</p>
```
Kev: I suggest instead of "namespace. In addition, it MUST" we simply use ", which MUST". Otherwise it's unclear if the stanza or payload must have the id.

linkmauve: Thanks for the suggestion.

linkmauve:
I rebased on top of master, this PR can now be merged.

PR #695, opened 2018-08-18
XEP-0060: Add an example on returning fewer items than requested

linkmauve:
That shows what happens when the client requests some items from a node and none are present.

### DIFF
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -51,6 +51,14 @@
   &ralphm;
 
   <revision>
+    <version>1.15.3</version>
+    <date>2018-08-18</date>
+    <initials>egp</initials>
+    <remark>
+      <p>Add a new sentence and example about what to do if there are fewer items than requested present in a node.</p>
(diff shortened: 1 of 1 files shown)


xep-0060.xml
```diff
@@ -2387,6 +2395,17 @@ And by opposing end them?
     </items>
   </pubsub>
 </iq>
+]]></example>
+    <p>If there are fewer items in the node than requested, the service should just return all of them.</p>
```
SamWhited: "just" isn't necessary, "the service should return all of them" reads fine.

xep-0060.xml
```diff
@@ -2387,6 +2395,17 @@ And by opposing end them?
...
+    <example caption='Service returns no items if it has none'><![CDATA[
```
SamWhited: Nit: "Service has no items" or similar would be shorter, it obviously can't return any items if it doesn't have any.

PR #696, opened 2018-08-18
XEP-0367: Update to specify the correct ID to use within MUCs

mwild1:
The current version of the XEP currently only uses the 'id' attribute, which is a namespace shared by all clients in a MUC, leading to collisions and race conditions.

This update switches it to use the stanza-id assigned to the message by the MUC.

### DIFF
--- a/xep-0367.xml
+++ b/xep-0367.xml
@@ -168,9 +168,9 @@
 <section1 topic='Security Considerations' anchor='security'>
   <p>
     Clients that implement message attachments MUST display the attachments
-    in such a way that they could be confused with the original message and
-    cause someone viewing the conversation to assume they were sent by the
-    sender of the message being attached to.
+    in such a way that they could not be confused with the original message
+    and cause someone viewing the conversation to assume they were sent by
+    the sender of the message being attached to.
   </p>
   <p>
     When matching a received attachment to the original message, clients must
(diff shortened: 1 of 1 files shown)


SamWhited:
LGTM

xep-0367.xml
```diff
@@ -109,42 +117,67 @@
...
 <section1 topic='Security Considerations' anchor='security'>
   <p>
-    Clients that implement message attachments MUST be careful not to display
-    the attachments in such a way that they could be confused with the original
-    message and cause someone viewing the conversation to assume they were sent
-    by the sender of the message being attached to.
+    Clients that implement message attachments MUST display the attachments
+    in such a way that they could be confused with the original message and
+    cause someone viewing the conversation to assume they were sent by the
+    sender of the message being attached to.
```
jcbrand: @mwild1 Is there a "not" missing here?

I.e. "in such a way that they could NOT be confused with the original message"

mwild1: @jcbrand @SamWhited Ha, sorry for the late response. Yes, obviously there is a 'not' missing :)

Flowdalic:
You may want to consider adding XEP-0359 to the dependencies meta information:

````
  <dependencies>
    <spec>XMPP Core</spec>
  </dependencies>
````

PR #697, opened 2018-08-20
XEP-0384: Receiving PreKeyMessages should trigger signed prekey rotation

vanitasvitae:
See the [protocol audit](https://conversations.im/omemo/audit.pdf) chapter 4:
```
The  lifetime  of  (signed)  prekeys  should  be  mentioned  in  the  standard.   Signed  prekeys
should be changed regularly in order to achieve forward secrecy.  This should at least be done
after every time the user receives a PreKeySignalMessage that uses the latest signed prekey,
but  it  can  be  done  more  often  (based  on  time)  to  ensure  the  forward  secrecy  of  dropped
messages.
```


jcbrand:
@strb Would you mind reviewing this PR?

vanitasvitae:
ping @strb

strb:
I apologize for the delay, this somehow slipped through the cracks.

This change is good.

The only thing that bothers me a little is the change to the acknowledgements. I'm not sure if this is just "how it's done", but it feels a bit weird to me, almost like having words put into my mouth. Don't get me wrong, I'm not at all against contributors being acknowledged properly. I'm very grateful that people are interested in OMEMO and doing work on it. But I'd prefer if those were listed separately. The list of names that was added to in this revision was not really a list of contributors, but a list of people whose input I had specifically sought out when drafting the initial version of the standard, though I can see how it'd seem that way, so I don't blame you for doing it like this.

I'd be in favor of just adding a separate paragraph that says something like "Thanks to X, Y, Z for their contributions to the standard" or something like that. This way, in the future, further contributors would then have an obvious place to add themselves to if they want. (e.g. @linkmauve in #686) Not sure how this is handled in other XEPs though?

The revision entry w/ version bump is missing, but since this, #686 and #707 are independent changes the ordering doesn't matter, so this can be resolved in any order.

vanitasvitae:
> But I'd prefer if those were listed separately.

Sure, I can fix that :)

vanitasvitae:
I reverted the changes to the acknowledgements for now.

horazont:
@strb Please check whether the PR is good now.

PR #706, opened 2018-09-24
Update BCP 14 language to comply with RFC 8174

SamWhited:
Clarify that normative RFC 2119 language MUST be in all caps.

### DIFF
--- a/xep.xsl
+++ b/xep.xsl
@@ -481,7 +481,15 @@ content: "XEP-<xsl:value-of select='/xep/header/number'/>";
         <p class='indent'>Errata can be sent to &lt;<a href='mailto:editor@xmpp.org'>editor@xmpp.org</a>&gt;.</p>
         <!-- CONFORMANCE TERMS-->
         <h3 id='appendix-conformance'>Appendix F: Requirements Conformance<xsl:call-template name='anchor-link'><xsl:with-param name='anchor' select='"appendix-conformance"'/></xsl:call-template></h3>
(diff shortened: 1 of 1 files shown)


Flowdalic:
I like the idea, but I am a bit worried as this is essentially a non-backwards compatible change: All non-capitalized RFC 2119 keywords in previous XEP texts (if there are any) will lose their semantic. I wonder if we should introduce a appendix-conformance-v2 and just change the template to use this instead, so that previous XEPs are not affected by the change.

SamWhited:
I really doubt it matters; we can clean it up later if we notice any. As far as I'm aware all recent XEPs are using caps.

Kev:
I'm not sure why this would be a good thing - 2119 doesn't require the words to be in caps, why would we benefit from doing so?

Flowdalic:
> I'm not sure why this would be a good thing

I'd say for the same reason RFC 8174 was written.

Kev:
Yep, being stupid, ta. I'd have thought Council would be a sensible venue for discussion. I'd have thought we'd need to bump the versions of all the affected XEPs, too, as this affects the reading of them, and probably grepping through to check that we don't have any non-upper-case uses that this would affect.

horazont:
Okay, editors. Since council accepted this, do you have opinions on how we merge this?

* Simply merge it: Changes the text of all XEPs without revision block.
* Bump the minor version number of all the XEPs. Possible, I can script that (I already have tools to auto-add revision blocks, so that’s not too bad). Creates a huge diff, a large amount of new revisions (I wouldn’t send emails for that).

Opinions?

SamWhited:
Just merge it. It won't change anything in practice and it's not a change to the actual XEPs themselves, just a change to some metadata associated with all XEPs (more or less).

Kev:
If revision blocks can be added by script, that is best - our policy is that all XEP changes should have a new revision block, and this is a non-editorial change to many XEPs so would certainly be better with one.

SamWhited:
Rebased. Ping seeking a decision on my last comment. Thanks!

SamWhited:
Ping again. This was approved quite a while ago.

PR #707, opened 2018-10-05
XEP-0384: Disallow backup of session data

vanitasvitae:
Some applications allow backups of session related data in the cloud. Restoring these backups will likely break sessions with other devices.


horazont:
@strb Please comment.

PR #709, opened 2018-10-08
XEP-0384: Add message counter logic to determine stale devices

vanitasvitae:
This PR adds a business rule that helps to determine stale/inactive devices and specifies a way to deal with them.

This is a fix for a protocol vulnerability which might lead to loss of forward secrecy.

Wording might be subject to change :)


vanitasvitae:
I changed the wording a bit to exclude "ping messages" from incrementing the counter. That way a device which is overdoing ratchet forwarding will not consider peer devices as read-only prematurely.

horazont:
@strb Please comment.

PR #715, opened 2018-11-05
XEP-0045: Add missing disco#info feature to example 4, 9, 78 and 218

Flowdalic:
As XEP-0030 § 3.1 states that "…every entity MUST support at least the
'http://jabber.org/protocol/disco#info' feature" it needs to be
present in every disco#info result response.


dwd:
(Rejected by Council, 2019-01-23)

PR #716, opened 2018-11-06
XEP-0030: Clarify 'disco#info' feature in 'disco#info' responses

Flowdalic:
Strictly speaking, every disco#info IQ result must contain the
disco#info feature. But looking at the existing examples in various
XEPs and implementations, this feature is usually omitted. Hence we
clarify in XEP-0030 that announcing the 'disco#info' feature is
optional, as support for it is implicitly acknowledged by returning a
result IQ to a 'disco#info' IQ request.

After thinking about the situation a while, this is currently my preferred solution. But I am not sure if it may causes issues regarding entity capabilities 1.0 and 2.0 (Should 'disco#info' be part of the hash input or not? Probably it should.) . Hence it possibly deserves some discussion by the community and by council.

For a motivation see
- https://github.com/xsf/xeps/pull/715
- https://github.com/processone/ejabberd/issues/2661


horazont:
> Hence it possibly deserves some discussion by the community and by council.

Please bring it to the standards@ list then. This is not the place for technical discussions.

horazont:
Brought it to the list https://mail.jabber.org/pipermail/standards/2018-November/035434.html

PR #718, opened 2018-11-08
XEP-0409: Incorporate my mailing list feedback

horazont:
Since nobody objected, here’s a concrete PR with my proposal. Summary:

- minor wording changes
- make server include the ``<im-ng/>`` marker for messages addressed to the full JID
- strip destination resource for forking non-IM-NG messages mapped to IM-NG
- transient hint


Ppjet6:
@Kev ping?
As it's been added in Future development for 423 (still LC), it might be good to take look at 409 again.

lnjX:
@horazont I hope this hasn't been answered yet, but about `<transient/>`: Is a new processing hint really needed here, can't we rely on `<no-store/>`?

PR #719, opened 2018-11-08
XEP-0001: add state transition Proposed -> Experimental

horazont:
Rationale: Sometimes, XEPs linger in Last Call state. This is probably for the same reason as XEPs which linger in Experimental state (and then become Deferred). However, currently there is no way to put Proposed (= Last Call) XEPs back into Experimental state if one follows the XEP-0001 process by the letter.

This is unfortunate, because it puts council into the position to either:

- Reject a XEP which is actually useful, but needs work before going to Draft (note that "Rejected" is a final state in the transition graph), or
- Advance a XEP to Draft which isn’t ready yet, or
- Burn cycles of the Editors and Council on perpetuating a Last Call which won’t lead to a conclusion at the current time.

(The "burning" of "cycles" is mostly due to the noise created by either officially extending the LC all the time, or having a zombie Proposed XEP hanging around in the list and having to recognize it as zombie and ignoring it.)

### DIFF
--- a/xep-0001.xml
+++ b/xep-0001.xml
@@ -22,6 +22,12 @@
   &stpeter;
   &dcridland;
   <revision>
+    <version>1.22.0</version>
+    <date>2018-11-08</date>
+    <initials>jsc</initials>
+    <remark><p>Officially allow moving XEPs back to Experimental after Last Call.</p></remark>
+  </revision>
+  <revision>
     <version>1.21.1</version>
     <date>2016-11-16</date>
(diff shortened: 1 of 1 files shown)


ralphm:
The Board approves of this changes (per today's vote), but wanted to know if @stpeter could say if it was intentional that the transition wasn't there before.

stpeter:
This was an oversight and not intentional. We probably just figured that things would always keep moving forward. :-)

PR #727, opened 2018-12-13
Obsolete some deferred XEPs

linkmauve:
As an attempt to clean up our deferred XEPs into the ones we may want to resurrect in the future, and the ones we know we will never want again, I’m obsoleting these three XEPs which all have a more modern equivalent.

### DIFF
--- a/xep-0008.xml
+++ b/xep-0008.xml
@@ -10,7 +10,7 @@
   <abstract>This specification provides historical documentation of an IQ-based protocol for exchanging user avatars.</abstract>
   &LEGALNOTICE;
   <number>0008</number>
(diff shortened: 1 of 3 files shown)


horazont:
This is not only an editorial change. Please increase the second version number.

linkmauve:
I took example on 4c5db548 in order to do that, since there is no change in functionality I don’t think incrementing major or minor would be any useful.

Or did we decide on only editorial changes being patch versions?

ivucica:
I noticed this was on agenda for the Council meeting yesterday, 2022-02-16. The changelog dates should likely be bumped before merge.

PR #731, opened 2019-01-05
XEP-0369: Remove MIX xmlns from pubsub <item/>

lnjX:
It's defined differently here: https://xmpp.org/extensions/xep-0369.html#info-node


horazont:
That place **must** have a namespace declaration. The other definition is wrong.

PR #736, opened 2019-01-13
XEP-0308: Allow message correction in MUC.

linkmauve:
Corrections done in MUC used to always be untrusted when coming from the
room’s history, because —according to this XEP— another entity could
have taken the same in-room full JID.

This addition makes the room responsible for rejecting such illegal
correction attempts, and advertise that in its disco#info, which clients
can then trust properly.

This also relaxes the rule on identity checking in non-anonymous rooms,
by making it rely on the real bare JID of the occupant.

### DIFF
--- a/xep-0308.xml
+++ b/xep-0308.xml
@@ -22,6 +22,12 @@
   <shortname>message-correct</shortname>
   &ksmith;
   <revision>
+    <version>1.1.0</version>
+    <date>2019-01-11</date>
+    <initials>egp</initials>
+    <remark><p>Add support for corrections in MUCs, including in history.</p></remark>
+  </revision>
+  <revision>
     <version>1.0</version>
     <date>2013-04-08</date>
     <initials>psa</initials>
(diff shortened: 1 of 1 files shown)


Kev:
I think it's not me this needs, but Council, isn't it?

xep-0308.xml
```diff
@@ -100,11 +106,17 @@
   <p>While it's not possible to prevent this protocol from being used in such a way, it is not intended that it provides a way to continue expanding a previous message indefinitely and clients, in as much as it is sensible, should present use of this extension only for correction rather than for providing a continuous stream, for which &xep0301; can be used instead.</p>
   <p>Correction MUST only be used to change the details of a stanza (e.g. the message body) and not to change the nature of the stanza (e.g. correction MUST NOT be used to turn a chat message into a pubsub notification)</p>
   <p>While it is possible to use this protocol to correct messages older than the most recent received from a full JID, such use is out of scope for this document and support for this SHOULD NOT be assumed without further negotiation.</p>
+  <p>A &xep0045; room SHOULD reject any correction attempt made by a different real bare JID than the one from which the message originates.  In order to notify clients of this behaviour, it MUST include the 'urn:xmpp:message-correct:0' feature in its disco#info if it applies this rule.</p>
```
dwd: Why SHOULD and not MUST?
Also, does the reverse hold true - must a room always allow correction when the bare jids are the same?

xep-0308.xml
```diff
@@ -100,11 +106,17 @@
...
-  <p>When used in a &xep0045; context, corrections must not be allowed (by the receiver) for messages received before the sender joined the room - particularly a full JID leaving the room then rejoining and correcting a message SHOULD be disallowed, as the entity behind the full JID in the MUC may have changed.</p>
+  <p>When used in a &xep0045; context, the receiver of a correction should check if the room advertises the 'urn:xmpp:message-correct:0' feature in its disco#info.  If it does so, the receiver can trust any message correction received through this room, be it live or from history.  Otherwise, if this feature isn’t advertised by the room, the receiver should verify each correction coming its way, for instance:</p>
+  <ul>
+    <li>If the room is non-anonymous and the bare JID of the sender of both the message and its correction is the same, allow it.</li>
+    <li>If the room is semi-anonymous and the message came live from the same occupant, allow it.</li>
+    <li>If the room is semi-anonymous and the message came from the history, from another sender, or from someone with the same nick but who left and came back, reject it.</li>
+  </ul>
```
dwd: This seems to be lacking in normative language. Are these MUSTs? Is the "should" actually a "SHOULD"?

horazont:
Vetoed during council meeting on 2019-04-17 due to the unexplained implications of the DoS vector this creates and other reasons. Please make a new PR or reopen this one with fixes and ping an editor.

PR #739, opened 2019-01-15
XEP-0410: pre-LC and LC feedback

ge0rg:
This WIP PR is meant to accumulate feedback from the XEP-0410 LC.


horazont:
(Note to editors: Don’t merge until first round of LC is over.)

PR #745, opened 2019-01-20
XEP-0234: Use <hash-used/> to signify the hash function being used, instead of an (invalid) empty <hash/>

Ri0n:
Most likely smeone just forgot to update the exampe after introducing hash-used in this xep.

### DIFF
--- a/xep-0234.xml
+++ b/xep-0234.xml
@@ -873,7 +873,7 @@ a=file-range:1024-*]]></code>
           <name>test.txt</name>
           <range/>
           <size>6144</size>
-          <hash xmlns='urn:xmpp:hashes:2' algo='sha-1'/>
+          <hash-used xmlns='urn:xmpp:hashes:2' algo='sha-1'/>
         </file>
       </description>
       <transport xmlns='urn:xmpp:jingle:transports:s5b:1'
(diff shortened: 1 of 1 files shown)


horazont:
This is inconsistent. The text above the example explicitly talks about an empty ``<hash/>`` element. This, of course, is in conflict with what XEP-0300 says. So please update the PR so that it also fixes the text above the example.

horazont:
@stpeter @legastero Please review.

This changes the effective Schema and might require a namespace bump to fix, even though the previous text was (I think) invalid use of XEP-0300.

horazont:
@Ri0n As you can see, I notified the authors for feedback. This may take a few days.

Ri0n:
Thanks

stpeter:
Hmm, this does seem to be an error in the text and examples, and would also necessitate a change to the schema (which we should include in this PR). As to whether this requires a namespace bump, I leave that to the Council.

horazont:
@stpeter Thanks! I’ll try to squeeze this into today’s Council meeting.

@Ri0n Could you make the changes to the schema?

Ri0n:
Added to schema. Not sure if it's correct.
By the way the XEP also notices hash-used can be used in <checksum>. I believe it's somewhat wrong. Checksum should always provide hash value.

dwd:
(Passed Council, 2019-02-20)

PR #751, opened 2019-02-05
xep-0284: Fix xmlns in disco#info examples

Zash:
XEP-0030 was from the `http://jabber.org/protocol/*` era,
nothing else uses `http://xmpp.org/protocol`.
This appears to be a mistake, which is corrected in this PR.

I have not added a revision block, but can do if it's needed.


ralphm:
Looks good to me!

PR #758, opened 2019-02-11
XEP-0060: Expose pubsub#access_model and pubsub#publish_model 

edhelas:
This will allow clients to display proper information regarding the nodes configuration in their UI, especially when combined with the Affiliations and Subscriptions states (hide the publish icon, display an "open/public" node icon…).


dwd:
(Passed Council, 2019-05-15)

PR #760, opened 2019-02-13
XEP-0045: Add avatar support using XEP-0084

linkmauve:
This is a second attempt at getting MUC room avatars, after [Council rejected my previous proposal](https://mail.jabber.org/pipermail/standards/2018-September/035351.html).  This new proposal avoids introducing a new XEP, and instead will eventually allow deprecation/obsolescence of two of the mentioned ones, leaving only XEP-0084.

### DIFF
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -46,6 +46,12 @@
   <registry/>
   &stpeter;
   <revision>
+    <version>1.32</version>
+    <date>2019-02-13</date>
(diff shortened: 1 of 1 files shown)


horazont:
Vetoed by Council on 2019-02-27.

Summary of the rationale was that list discussion was needed, especially around the PubSub-like wire format. Please take this to the list and re-propose.

PR #765, opened 2019-03-12
ProtoXEP: DNS Queries over XMPP (DoX)

moparisthebest:
There is [running code](https://github.com/moparisthebest/jDnsProxy/tree/dox) that implements this, as well as a resolver anyone can poke running at [dns@moparisthebest.com/listener](xmpp:dns@moparisthebest.com/listener)

I would humbly suggest this might be accepted as a XEP on the first of next month, if council approves. ;)  Feel free to contact me via XMPP with any questions.

Rendered version: https://www.moparisthebest.com/dox.html

Thanks!

### DIFF
--- a/inbox/dox.xml
+++ b/inbox/dox.xml
@@ -1,0 +1,181 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE xep SYSTEM 'xep.dtd' [
+  <!ENTITY % ents SYSTEM 'xep.ent'>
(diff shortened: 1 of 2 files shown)


moparisthebest:
I've updated namespace from 'https://moparisthebest.com/dns' to 'urn:xmpp:dox:0' as suggested in the XSF channel, the running code at `dns@moparisthebest.com/listener` is still using the old namespace for now though.

PR #766, opened 2019-03-12
Update references to new XEPs and RFCs

zinid:
Add entities for XEP-0415, XEP-0416 and for referenced RFCs.


horazont:
Thanks, I forgot to add the entities.

PR #767, opened 2019-03-18
Mix CORE - XEP-0369

stevekille:
Address Dave Cridland Review Comments

Remove MIX xmlns from pubsub channel info &lt;item/&gt; example;
      Make Message Archiving Optional;
      Clarify that participant node only contains real JIDs;

### DIFF
--- a/xep-0369.xml
+++ b/xep-0369.xml
@@ -36,6 +36,14 @@
   &ksmithisode;
   &skille;
   <revision>
+    <version>0.14.2</version>
+    <date>2019-03-18</date>
+    <initials>sek</initials>
+    <remark><p>
+      Replace &lt;submission-id/&gt; with XEP-0359;
+    </p></remark>
+  </revision>   
+  <revision>
     <version>0.14.1</version>
     <date>2019-01-05</date>
     <initials>lnj</initials>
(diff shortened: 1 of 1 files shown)


stevekille:
Could you hold off pushing this for a bit?    There might be some more changes coming, and it seems sensible to include them, particularly as namespace has just been bumped, and we can include changes to protocol within it.

stevekille:
I believe discussion has died down, and this set of changes is now good to publish

horazont:
@stevekille For the future: If you’re not sure that something should be merged yet, please edit the title to prefix it with ``WIP:``. That sends a clear message.

stevekille:
Thanks for the WIP hint - I did not know this.

In this case, I thought it WAS ready to submit.  It was discussion with Dave C that suggested holding off for a bit, and adding the comment.

horazont:
@stevekille You can (I think) edit the title at any time. If you can't sorry for the noise.

I noticed this only now, please use a minor version bump (in this case, 0.14.1 -> 0.15.0) instead of a patch-level version bump (0.14.1 -> 0.14.2) for non-editorial changes, especially when bumping the namespace. It’s too late now as I already published everything, but please try to remember this for the next time so that we have two levels of validation here.

stevekille:
My apologies.  This was mostly editorial.   However, I did drop a bit of protocol and Dave made me bump the namespace.    I forgot to change the version to 15.

PR #771, opened 2019-03-24
XEP-0313: Relax 'communicating archive ID' requirement

guusdk:
The requirement keyword changed in this commit stems from the time when this XEP used ``<archived id='...' by='...' />`` instead of XEP-0359.

The 'MUST NOT' requirement keyword implies that something will go horribly wrong, if the ID is sent. This is not the case here.

Sharing the stanza-id on a need-to-know basis is a sensible default, but should not be a requirement.

### DIFF
--- a/xep-0313.xml
+++ b/xep-0313.xml
@@ -220,7 +220,7 @@
 	</section2>
 	<section2 topic='Communicating the archive ID' anchor='archives_id'>
 		<p>When a message is archived, the server MUST add an &lt;stanza-id/&gt; element as defined in &xep0359; to the message, which informs the recipient of where and under what ID the message is stored. When doing this the server MUST follow the business rules defined in XEP-0359. The 'by' attribute MUST be set to the address of the archive. For regular users that’s the bare JID of the account and for MUC that’s the bare JID of the room.</p>
-		<p>Servers MUST NOT include the &lt;stanza-id/&gt; element in messages addressed to JIDs that do not have permissions to access the archive, such as a users’s outgoing messages to their contacts. However servers SHOULD include the element as a child of the forwarded message when using &xep0280;</p>
+		<p>Servers SHOULD NOT include the &lt;stanza-id/&gt; element in messages addressed to JIDs that do not have permissions to access the archive, such as a users’s outgoing messages to their contacts. However servers SHOULD include the element as a child of the forwarded message when using &xep0280;</p>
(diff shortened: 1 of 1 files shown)


mwild1:
Looks good to me. Consequences of revealing the archive ID may be bad if the ID contained any state (e.g. allowing someone to determine how many messages you exchanged). However XEP-0313 defers to XEP-0359 on the format of IDs, and that says:

> The value of the 'id' attribute should not provide any further information besides the opaque ID itself. Entities observing the value MUST NOT be able to infer any information from it, e.g. the size of the message archive. The value of 'id' MUST be considered a non-secret value.

So as long as implementations that break this 'SHOULD NOT' still obey that rule, it's fine.

Kev:
This /does/ reveal information about whether the message has been archived though, which is itself a thing you might not want?

mwild1:
@Kev that seems like a sensible thing to add for implementations that violate the "SHOULD NOT", indeed.

@guusdk assuming you're still keen on getting this changed, would you add such a note and a '(see security considerations)' or something along those lines?

guusdk:
Do we really need to clutter the XEP with a warning like that? Worst case, the receiving end (that apparently already was permitted to receive the original message) is informed that the text is archived. I have to force myself to see a security issue with this. "SHOULD NOT" is sufficient, I'd argue.

Kev:
This makes me wonder whether we should drop the 359 use and go back to just using a MAM-specific ID. It's not clear to me why reducing from MUST NOT to SHOULD NOT would be a beneficial thing - it certainly creates issues - but I'm assuming the desire to do so is somehow related to re-use of 359.

Flowdalic:
> This /does/ reveal information about whether the message has been archived though

I don't think it does. Having a <stanza-id/> just tells you that the entity with the address found in the 'by' attribute assigned this ID to the message.

Kev:
If we don't assume that a MAMming entity injecting an ID means that the message has been archived, we're going to have things going badly wrong with archive lookups. So I think that adds some weight to my previous wondering about reverting 359 out of 313.

Flowdalic:
> So I think that adds some weight to my previous wondering about reverting 359 out of 313.

I am not sure I agree with that. See also the recent discussion on standard@ (where this should have been raised in the first place)

horazont:
Please discuss technical issues on standards@.

Kev:
No activity in years.

PR #779, opened 2019-04-05
XEP-0184: add a box about types and JIDs

ge0rg:
Because it led to ambiguity in the past, here is a small informational box about reflecting the message type in Receipts.


dwd:
Council vote PASSED

PR #780, opened 2019-04-06
XEP-0001: Make only editorial changes affect Deferred state

horazont:
See https://trello.com/c/U3OJ4sQx/328-clarify-what-happens-when-a-typo-or-equivalent-is-fixed-in-a-deferred-xep


xep-0001.xml
```diff
@@ -295,6 +295,7 @@
...
   <p>If an Experimental XEP is inactive (i.e., no updated versions are published) for a period of twelve (12) months, the XMPP Extensions Editor shall automatically change the status of the XEP to Deferred unless it is in the queue of XEPs under active consideration for advancement by the Approving Body; upon submission of an updated version, the XMPP Extensions Editor shall change the status back to Experimental.</p>
+  <p>Only substantial, non-editorial changes count as <em>activity</em> (or <em>updates</em>) for the purpose of considering moving a XEP from or to Deferred state.</p>
```
Kev: I suggest tying this into the versioning scheme as that's more concrete (and is consistent with the existing text that only understands substantive updates). e.g. something like 
```suggestion
  <p>Only substantial, non-editorial changes (e.g. those that would cause an updated version of 0.1 to 0.2, not editorial updates from 0.1.1 to 0.1.2) count as <em>activity</em> (or <em>updates</em>) for the purpose of considering moving a XEP from or to Deferred state.</p>
```

horazont: I thought about that, too, but XEP-0001 does not exactly define how we version things. So to be able to refer to that, we’d have to go ahead and write that down clearly. Which I’m not opposed to, in general, but time and stuff.

Kev: I think it does, actually, from memory define "Updated Versions" as those that have a 0.x.0 minor version bump, talks about when x.0.0 major bumps happen, and the patch 0.0.x versions aren't mentioned (i.e. out of scope of xep1).

lnjX: @SeveFP I agree that this reads a little odd.

iNPUTmice:
Council has accepted this.

ralphm:
If @Kev is ok with this change without his suggestion, then please resolve the comment.

PR #781, opened 2019-04-09
XEP-0308. Correct contents only

Kev:


dwd:
Vote complete, passed.

PR #783, opened 2019-04-10
make xml: Include  XEP XML dependencies in output

Zash:
This should help with "undefined entity" errors when viewing the XEP XML
files online.

E.g. https://xmpp.org/extensions/xep-0030.xml currently shows:

```
XML Parsing Error: undefined entity
Location: https://xmpp.org/extensions/xep-0030.xml
Line Number 11, Column 5:
    &LEGALNOTICE;
----^
```


Zash:
Looks like at least Firefox won't ever load entities from an external file, so this approach won't work.

PR #786, opened 2019-04-17
XEP-0045: Move section about GC 1.0 further down

Zash:
It's weird that the first thing you see under "Entering a room" is how
to do it in the old way, instead of the new way, which might mislead
implementers and make them miss the new way entirely.


horazont:
Pending. Seems to be included in #787.

Zash:
This was meant as a smaller editorial-only variant of #787 so this can be closed if that was accepted.

PR #787, opened 2019-04-17
XEP-0045: kill GC1.0 with fire

ge0rg:
This patch set delivers (with a bit of delay) on my promise of providing a PR for GC1.0 removal, as discussed and agreed upon in https://mail.jabber.org/pipermail/standards/2018-April/034768.html

It does the following:
* change wording that claims compatibility with GC1.0
* move the GC1 section from the top to the bottom of 'Entering a Room' (thx Zash)
* mandate that a service SHOULD kick a non-occupant sending a GC1.0 join / presence update (thx Kev)

### DIFF
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -534,7 +534,7 @@
 </section1>
 
 <section1 topic='Requirements' anchor='reqs'>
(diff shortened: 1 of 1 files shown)


dwd:
Council vote PASSED

PR #789, opened 2019-04-28
XEP-0380: Add OMEMO encryption method

lnjX:


lnjX:
Probably needs the approval of @linkmauve.

lnjX:
@horazont friendly reminder :)

PR #793, opened 2019-06-11
XEP-0166: Relax transport element requirement

Ri0n:
It's possible a responder may want to replace transport before session/content accept.
In this case it's necessary to send initial transport offer twice (with replace and with accept) which basically looks wrong.
Instead it's better to omit transport element when it was already negotiated.

This change probably needs ns bump.

### DIFF
--- a/xep-0166.xml
+++ b/xep-0166.xml
@@ -1195,7 +1195,7 @@ PENDING  o-----------------------+   |
     </section3>
     <section3 topic='session-accept' anchor='def-action-session-accept'>
       <p>The <strong>session-accept</strong> action is used to definitively accept a session negotiation (implicitly this action also serves as a content-accept). A session-accept action indicates a willingness to proceed with the session (which might necessitate further negotiation before media can be exchanged). The session-accept action indicates acceptance <em>only</em> of the content definition(s) whose disposition type is "session" (the default value of the &CONTENT; element's 'disposition' attribute), not any content definition(s) whose disposition type is something other than "session" (e.g., "early-session" for early media).</p>
-      <p>In the session-accept stanza, the &JINGLE; element MUST contain one or more &lt;content/&gt; elements, each of which MUST contain one &lt;description/&gt; element and one &lt;transport/&gt; element.</p>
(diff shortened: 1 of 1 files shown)


fippo:
hey @Ri0n 

I don't think this is the correct way to do it. `session-initiate` is always the first message in a jingle session and the only thing that can create a session id, see https://xmpp.org/extensions/xep-0166.html#concepts
This means you can't do a transport-replace earlier as you don't have a session-id.

Now transport-replace is difficult... XEP-0260 uses this in odd ways to fallback from s5b to ibb. There is no negotiation of the transport to use (even though session-accept says there can be multiple transports but the session-initiate can only contain a single one?!) so this is probably done via disco/caps.

I do think there is something to fix here.

(maybe we should move to standards@ to keep discussions in one place)

Ri0n:
@fippo 
I say nothing about session-initiate. It's about session-accept or content-accept (when session was already accepted before) and xep-0166 says we can do a lot of negotiation before session-accept. So nothing wrong is here. We can do transport-replace before session-accept.

Regarding multiple transports in session-initiate/accept there is a way to add multiple content elements with same name but different transports. Basically I already implemented a support for this in Psi (for incoming requests). But none of XEPs says it's a valid case explicitly.

a standards@ thread is already started https://mail.jabber.org/pipermail/standards/2019-June/036204.html

fippo:
> I say nothing about session-initiate

Ah, the spec is written in confusing ways. The section where you made the first change and that specifies what elements are contained in the <content/> is labelled `Initiation` and talks about the session-initiate but in a way that makes it sound a bit like its about the content element in general.

Moving to standards@ for the rest

Ri0n:
Oh. maybe I did the change a little bit blindly. I'll review/change it. Yep let's keep discussing at standards@

horazont:
It was mentioned already, but technical discussion must not take place in the issue tracker, but on the mailing list.

I’ll close this for now, please re-open once list discussion has settled.

iNPUTmice:
Council has rejected this modification.

PR #794, opened 2019-06-26
xep-0371: fix gathering-complete text

fippo:
Fixes the text to use the correct element name which is used in the sample and the schema


stpeter:
+1 LGTM

PR #796, opened 2019-06-30
XEP-0368: clarify what happens when a `.` target is published

horazont:
This is proposal to formalise the results of the [list discussion][1].

   [1]: https://mail.jabber.org/pipermail/standards/2019-June/036235.html

Further (non-editorial) comments should go in that thread.


horazont:
```
PASSED (-0:1:+4)
PR #796 - XEP-0368: clarify what happens when a `.` target is published - https://github.com/xsf/xeps/pull/796
Dave: +1
Georg: +1 (this is just a clarification of RFC 2782)
Jonas: +1
Kev: [abstained]
Link: +1 (definitely!)
```

PR #797, opened 2019-07-01
XEP-0128: Remove 'unlikely' statement.

guusdk:
With various XEPs (XEP-0157, XEP-0232, XEP-0363) depending on Service Discovery Extensions,
it is no longer 'unlikely' that more than one extension is added.


horazont:
Passed council vote.

PR #801, opened 2019-07-19
XEP-0368: clarify fallback behavior on SRV failure

moparisthebest:
Half of this proposes the opposite of #796


horazont:
[Veto-ed unanimously by council on 2019-07-31](https://logs.xmpp.org/council/2019-07-31#2019-07-31-0e2255bfa44d2e7f). Most agreed to the rationale summarised by myself:

> - the port choice is, as Link Mauve put it, "yolo", which isn’t really appropriate
> - It doesn’t clarify like the title suggests, but instead changes behaviour from the previous business rules
> - Contradicts how SRV works normally
> - Isn’t logically coherent (don’t trust DNS, but trust DNS)

horazont:
```
2019-07-31 (expired 2019-08-14)

VETOED (-5:0:+0)
PR #801 - XEP-0368: clarify fallback behavior on SRV failure - https://github.com/xsf/xeps/pull/801
Dave: -1 (don't think I like possibly anything about this)
Georg: -1 (agree with Jonas)
Jonas: -1 (inappropriate port choice; doesn't clarify, but changes behaviour; contradicts normal SRV working; logically incoherent regarding trust of DNS)
Kev: -1 (don't like it at all; in comparison, #796 seems logical)
Link: -1 (same reasons as Jonas)
```

PR #802, opened 2019-07-19
Symlink xslt file in inbox

Ppjet6:
XEPs in inbox/ refer to "xep.xsl" that doesn't exist.


Ppjet6:
Merged as 64009c2.

PR #809, opened 2019-08-07
XEP-0364: Fix link to Daniels article

jubalh:
Let me know if I should also add a new revision entry.


horazont:
Note to editors: fix the typo right around the diff, too.

PR #810, opened 2019-08-09
XEP-0372: Fix incorrect pubsub URIs

Ppjet6:
All these URIs were missing a ";".

### DIFF
--- a/xep-0372.xml
+++ b/xep-0372.xml
@@ -29,7 +29,7 @@
     <version>0.3</version>
     <date>2019-09-04</date>
     <initials>(mb)</initials>
-    <remark>Make Pubsub URIs conform to RFC5222 in sections 3.3 and 3.4</remark>
+    <remark>Make Pubsub URIs in sections 3.3 and 3.4 conform to &xep0147;.</remark>
   </revision>
   <revision>
     <version>0.2</version>
--- a/xep-0372.xml
+++ b/xep-0372.xml
@@ -29,7 +29,7 @@
     <version>0.3</version>
     <date>2019-09-04</date>
     <initials>(mb)</initials>
-    <remark>Make Pubsub URIs in sections 3.3 and 3.4 conform to &xep0147;.</remark>
+    <remark>Make Pubsub URIs in sections 3.3 and 3.4 conform to XEP-0147.</remark>
   </revision>
   <revision>
     <version>0.2</version>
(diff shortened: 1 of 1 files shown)


horazont:
@Ppjet6 Sorry, the change is not obvious to me. Where is the URL format documented?

Ppjet6:
https://logs.xmpp.org/xsf/2019-08-09#2019-08-09-4deb7f079b78f055 This is the discussion on `xsf@`, that was confirmed by ralphm. And the URI format is documented in RFC 5122.

https://tools.ietf.org/html/rfc5122#section-2.5

> Note: The delimiter between key-value pairs is the ";" character instead of the "&" character used in many other URI schemes.  This delimiter was chosen in order to avoid problems with escaping of the & character in HTML and XML applications.

horazont:
@Ppjet6 Could you please form a revision block which includes a reference to RFC 5122?

Ppjet6:
Is that ok? I also rebased on master.

xep-0372.xml
```diff
@@ -25,6 +25,12 @@
   <supersededby/>
   <shortname>Refs</shortname>
   &ksmithisode;
+  <revision>
+    <version>0.3</version>
+    <date>2019-09-04</date>
+    <initials>(mb)</initials>
+    <remark>Make Pubsub URIs conform to RFC5222 in sections 3.3 and 3.4</remark>
```
ralphm: This should be RFC 6122? You want to use the `&rfc6122;` XML entity reference for this. And maybe you also want to reference `&xep0147;`.

ralphm: Gah. I meant 5122 of course.

horazont: You cannot use those entities in ``<remark/>``:

```
xep-0372.xml:32: element remark: validity error : Element span is not declared in remark list of possible children

  <remark>Make Pubsub URIs in sections 3.3 and 3.4 conform to &xep0147;.</remark

                                                                               ^

xep-0372.xml:32: element remark: validity error : Element note is not declared in remark list of possible children

  <remark>Make Pubsub URIs in sections 3.3 and 3.4 conform to &xep0147;.</remark

                                                                               ^

Makefile:110: recipe for target '/var/www/html/extensions//xep-0372.html' failed
```

Ppjet6:
I'm using the shortened version of the name, as other XEPs do (grepping in the repo).

Ppjet6:
@Kev I need you to ACK this :)

Kev:
Ack.

PR #812, opened 2019-08-28
XEP-0084: Bump bytes datatype from unsignedShort to unsignedInteger

vanitasvitae:
Related Mailing List Posting: https://mail.jabber.org/pipermail/standards/2019-August/036376.html

### DIFF
--- a/xep-0084.xml
+++ b/xep-0084.xml
@@ -589,7 +589,7 @@
     <xs:complexType>
       <xs:simpleContent>
         <xs:extension base='empty'>
-          <xs:attribute name='bytes' type='xs:unsignedShort' use='required'/>
+          <xs:attribute name='bytes' type='xs:unsignedInteger' use='required'/>
           <xs:attribute name='height' type='xs:unsignedShort' use='optional'/>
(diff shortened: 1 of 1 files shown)


xep-0084.xml
```diff
@@ -37,6 +37,12 @@
   &pgmillard;
   &temas;
   &xvirge;
+  <revision>
+    <version>1.1.3?</version>
```
horazont: @vanitasvitae Please don’t do that questionmark thing. Git will show me a conflict either way if another version was released inbetween. I might not look closely at the version after initially reviewing the PR, however, unless there’s a conflict.

(And I don’t have a label for "It has a revision block, but I definitely need to pay attention to it beyond the date.)

vanitasvitae: Sorry, I'll patch that out pretty quick :)
For the future: Do you prefer to do the revision block yourselves or what should be filled out by the PR creator?

horazont: @vanitasvitae I prefer if the PR creator does it, because I don’t always know the preferences regarding initials and remark style. I reserve the right to criticize them though ;)

PR #814, opened 2019-09-02
XEP-0393: clarify quote termination and trimming

SamWhited:
Tweak whitespace trimming in block quotes to indicate that only the first space should be trimmed. Also clarify how block quotes are terminated.


horazont:
@SamWhited This doesn’t read like an editorial change to me, so I suggest to bump the minor version number.

SamWhited:
@horazont bumped

PR #815, opened 2019-09-03
Correct form definition to comply with standards

rufferson:
Minor fix: Correct Data Forms in examples to comply with XEP-004 and XEP-086 requirements - mandatory type attributes.


horazont:
@Kev review please.

PR #817, opened 2019-09-05
Reaction

Nyco:
Not "rection"


dwd:
I sincerely hope that was the intent.

Kev:
The change is obviously sensible, although it's now a XEP rather than in the inbox.
(It's also editorial so doesn't need review)

PR #820, opened 2019-09-09
XEP-0280: remove error note, add XEP-0333

ge0rg:
These are some editorial changes of the Carbon Eligibility Rules:

* removed the "note" regarding non-tracking of message errors, because it was just a note and message errors **are** important.
* added XEP-0333 as another example for "typical" IM payloads

### DIFF
--- a/xep-0280.xml
+++ b/xep-0280.xml
@@ -298,8 +298,8 @@
     <ul>
       <li>it is of type "chat".</li>
       <li>it is of type "normal" and contains a &lt;body&gt; element.</li>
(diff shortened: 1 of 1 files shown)


ge0rg:
Somehow, I ended up [retracting this](https://logs.xmpp.org/council/2019-09-11?p=h#2019-09-11-b9941df0417574c7) because it would need a bump of the carbon-rules feature, and also a conclusive list of IM-related XEPs.

> Okay, in that case I retract the PR until somebody comes up with a conclusive list of all IM-related payloads.

PR #824, opened 2019-09-12
XEP-0060: Add pubsub#public in Publish-Subscribe features

edhelas:
This feature is similar to 'muc#roomconfig_publicroom' but applied to Pubsub nodes.


Ppjet6:
Council seems to have forgotten about this.

https://mail.jabber.org/pipermail/standards/2019-October/036511.html
Only Kev has replied to this with a "I don't understand what this is for" comment. And the vote has expired.

I seem to remember mentions of "things like this" being more appropriate in the registry that is in the process of getting fixed. You might want to plead your case and give them a bit more details after this fix?

PR #825, opened 2019-09-12
XEP-0328 (jid-prep): Update to 0.2

Flowdalic:
Rendered version: http://geekplace.eu/xeps/xep-jidprep/xep-jidprep.html
Diff: http://geekplace.eu/xeps/xep-jidprep/diff.html
Diff (side-by-side): http://geekplace.eu/xeps/xep-jidprep/diff-side-by-side.html

To quote @legastero: 

> 20190908T17:54:10Z 000 <Lance>  flow: consider 328 yours to update. browsers have exposed string normalization functions since that was written in 2012/2013, so it is possible to do prepping in the browser now without needing half a mb of data tables

The XSF Editor may want to add rfc8264 to `xep.ent`.

### DIFF
--- a/xep-0328.xml
+++ b/xep-0328.xml
@@ -6,11 +6,11 @@
 <?xml-stylesheet type='text/xsl' href='xep.xsl'?>
 <xep>
 <header>
-  <title>JID Prep</title>
-  <abstract>This specification defines a way for an XMPP client to request an XMPP server to prep and normalize a given JID.</abstract>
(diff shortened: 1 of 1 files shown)


horazont:
@Flowdalic Can you please move the entity for RFC 8264 into xep.ent right away? Saves us a round-trip.

Flowdalic:
jc already added RFC 8264 into xep.ent with c3cba21f ("Another xeps.ent update"). I have updatedxep-0328.xml accordingly.

PR #827, opened 2019-09-18
XEP-0419: fix potential vulnerability typo

sonnyp:


sonnyp:
After debating with my peers and careful considerations, we have come to the conclusion that it is safer to use the British form rather than the American form due to the Edward Snowden revelations.

However it has come to our attention that most [CVEs](https://cve.mitre.org/) are written in English which indicates that English is not a safe language. We will come back to the XSF with a proposal to adopt a different language.

PR #828, opened 2019-09-18
XEP-0045: add muc#roominfo_webchat_url form field as used in the wild

horazont:


horazont:
[Vetoed](https://mail.jabber.org/pipermail/standards/2019-September/036489.html)

PR #833, opened 2019-09-25
New protoXEP: Message Moderation

jcbrand:
* Send an IQ to a groupchat service to retract a message
* The groupchat service is responsible for sending out a retraction message
* Use XEP-0422 for the retraction message sent out by the service

~Note: I'm making this PR against the `message-retractions` branch, because it relies on that branch being merged first.~


jcbrand:
@horazont or @Ppjet6 would you mind taking a look?

Ppjet6:
It builds on my machine, no obvious issues, the rest is council realm I guess.

PR #834, opened 2019-09-25
XEP-0410: treat remote-server-{not-found,timeout} like timeout

horazont:
Rationale is inline. Summary: those errors indicate unreachability
of the server and most certainly not (with confidence) that the
client has been removed from the room.

### DIFF
--- a/xep-0410.xml
+++ b/xep-0410.xml
@@ -30,6 +30,12 @@
     <jid>georg@yax.im</jid>
   </author>
   <revision>
+    <version>1.1.0</version>
+    <date>2019-09-25</date>
+    <initials>jsc</initials>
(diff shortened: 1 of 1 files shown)


horazont:
@ge0rg Review in addition, please.

ge0rg:
I approve this change.

horazont:
Passed

```
15:09:21 Link Mauve> I’m +1 on this.
15:09:22 jonas’> +1 on that one
15:09:41 Ge0rG> As this is only a client behavior recommendation, I don't consider this a breaking change and thus I'm also +1 with my Council hat on
15:09:50 dwd> This seems sensible. I'm going to +1 on this.
15:14:31 Kev> +1
```

PR #835, opened 2019-09-28
XEP-0402: Fix issues found during Stockholm XMPP Sprint

linkmauve:
This PR might be more easily read commit per commit.

Rendered version: https://linkmauve.fr/extensions/xep-0402.html

I was extremely surprised at the difference of treatment for notifications between publishing an item (XEP-0060 §7.1.2.1) and purging all items (§8.5.2), both sending a notification if the node is configured for that, and retracting an item (§7.2.2.1) where the **publisher** additionally has to include a `notify='1'` attribute on the `<retract/>` element.

Another issue we’ve identified during the Sprint is password migration, as much as I dislike this feature there are existing MUCs being protected by passwords, and we most likely keep those until we decide on either removing this feature from MUC altogether, or find a more secure way to store passwords on the server.

Otherwise, this specification is pretty nice and should be ready for advancing to draft soon I think.

### DIFF
--- a/xep-0402.xml
+++ b/xep-0402.xml
@@ -175,6 +175,9 @@
                 <field var='pubsub#max_items'>
                     <value>10000</value>
                 </field>
+                <field var='pubsub#send_last_published_item'>
+                    <value>never</value>
+                </field>
                 <field var='pubsub#access_model'>
                     <value>whitelist</value>
                 </field>
--- a/xep-0402.xml
+++ b/xep-0402.xml
@@ -130,13 +130,15 @@
   <pubsub xmlns='http://jabber.org/protocol/pubsub'>
     <items node='urn:xmpp:bookmarks:0'>
       <item id='theplay@conference.shakespeare.lit'>
-          <conference name='The Play&apos;s the Thing'
+          <conference xmlns=']]>&namespace;<![CDATA['
+                      name='The Play&apos;s the Thing'
                       autojoin='true'>
             <nick>JC</nick>
           </conference>
       </item>
       <item id='orchard@conference.shakespeare.lit'>
-          <conference name='The Orchard'
+          <conference xmlns=']]>&namespace;<![CDATA['
+                      name='The Orchard'
                       autojoin='true'>
             <nick>JC</nick>
           </conference>
@@ -155,7 +157,8 @@
     <pubsub xmlns='http://jabber.org/protocol/pubsub'>
         <publish node='urn:xmpp:bookmarks:0'>
             <item id='theplay@conference.shakespeare.lit'>
-                <conference name='The Play&apos;s the Thing'
+                <conference xmlns=']]>&namespace;<![CDATA['
+                            name='The Play&apos;s the Thing'
                             autojoin='true'>
                     <nick>JC</nick>
                 </conference>
--- a/xep-0402.xml
+++ b/xep-0402.xml
@@ -172,6 +172,9 @@
                 <field var='pubsub#persist_items'>
                     <value>true</value>
                 </field>
+                <field var='pubsub#max_items'>
+                    <value>10000</value>
+                </field>
                 <field var='pubsub#access_model'>
                     <value>whitelist</value>
                 </field>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
> In order not to replace old bookmarks when pushing a new one, we need to request a good max_items value.  This one has been picked as some random large enough number, bikeshed welcome!

Also all clients will have to agree on one value because otherwise your preconditons will constantly fail when other clients have used a different value before. Which in turn means we really need support for max_items=0 to mean server side max.

Ppjet6:
This is for @dwd and/or @jcbrand to review.

xep-0402.xml
```diff
@@ -25,6 +25,19 @@
...
+  <revision>
+    <version>0.3.0</version>
+    <date>2019-09-28</date>
+    <initials>egp</initials>
+    <remark><ul>
+      <li>Fix examples.</li>
+      <li>Explain the expected workflow better.</li>
+      <li>Encourage clients to set pubsub#send_last_published_item to never and pubsub#max_items to some high value.</li>
+      <li>Add examples about notifications and +notify.</li>
+      <li>Request clients to add notify='1' on retract requests, so they do trigger notifications (see <link url='https://xmpp.org/extensions/xep-0060.html#publisher-delete-success-notify'>XEP-0060 §7.2.2.1</link> for a rationale).</li>
```
jcbrand: Nitpick: `retract requests` should probably be `retraction requests` to make it grammatically correct, alternatively `retract` should be put in quotes.

xep-0402.xml
```diff
@@ -102,25 +115,42 @@
...
+        <p>Note also that a conference element has no truly mandatory attributes or child elements, thus the following is legal:</p>
         <example caption='Minimalist conference element'><![CDATA[
 <conference xmlns=']]>&namespace;<![CDATA['/>
 ]]></example>
     </section1>
 
-    <section1 topic='Examples' anchor='examples'>
+    <section1 topic='Workflow' anchor='workflow'>
+
+        <section2 topic='Registering to notifications' anchor='enable-updates'>
```
jcbrand: `Registering to notifications` is grammatically incorrect. You register "for" something, not "to" unless "to" is shorthand for "in order to".

I therefore suggest "Registering to receive notifications".

xep-0402.xml
```diff
@@ -102,25 +115,42 @@
...
+            <p>A client interested in bookmarks SHOULD include the 'urn:xmpp:bookmarks:0+notify' feature in its &xep0115;, as per &xep0163;, so that it receives notifications for updates done by other clients of the user, and react accordingly.  The actual notifications are explained in the <link url='#notifications'>Bookmark Notifications</link> section of this specification.</p>
```
jcbrand: "react" should be "reacts".

xep-0402.xml
```diff
@@ -102,25 +115,42 @@
...
+
+            <example caption='Client replies with +notify to a XEP-0030 query from its server'><![CDATA[
```
jcbrand: Should XEP-0030 be linkified? e.g. `&xep0030;`

xep-0402.xml
```diff
@@ -102,25 +115,42 @@
...
+<iq from='juliet@capulet.lit/balcony' to='capulet.lit' type='result' id='disco1'>
+  <query xmlns='http://jabber.org/protocol/disco#info'>
+    <identity category='client' type='pc' name='poezio'/>
```
jcbrand: I don't think we should put subtle advertisements for clients in the XEPs.

xep-0402.xml
linkmauve: Using literal XML tags inside of an attribute is invalid in XML, so we can’t add any markup here.

xep-0402.xml
linkmauve: Should I invent a non-existing client instead?  I remember seeing mentions of Psi or Swift in some XEPs, so I thought using an existing client was something acceptable, especially since otherwise the XEP-0030 answer is invalid when there is no identity.

linkmauve:
I fixed the English issues, thanks for giving me a nice explanation!

xep-0402.xml
jcbrand: I guess it's ok.  My first impression is against it but I'm not going to argue over it.

jcbrand:
@linkmauve LGTM

PR #837, opened 2019-10-03
Reduce final image size by 92%

horazont:
By performing the build in one image (with all the needed tools)
and then copying over the result files into an image based on
stock nginx, the image size is reduced from 1.4 GB to 111 MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Builds all XEPs by default, HTML & PDF
 # docker build . --build-arg NCORES=X --build-arg TARGETS="html pdf"
 # from Dockerfile.base
-FROM xmppxsf/xeps-base:latest
+FROM xmppxsf/xeps-base:latest as build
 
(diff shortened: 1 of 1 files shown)


horazont:
Requested review from someone from iteam to be sure.

Ppjet6:
Not from iteam, but LGTM :)

Dockerfile
```diff
@@ -14,7 +14,10 @@ COPY texml-xsl/*.xsl /src/texml-xsl/
 
 WORKDIR /src
 RUN OUTDIR=/var/www/html/extensions/ make -j$NCORES $TARGETS
+RUN bash -c 'rm -f /var/www/html/extensions/*.{log,aux,toc,tex,tex.xml,out}'
```
Zash: Why delete `*.xml`?

horazont: That’s ``.tex.xml``, not ``.xml``.

PR #839, opened 2019-10-04
XEP-0391 & XEP-0396: JET improvements

vanitasvitae:
This PR adds improvements to both XEP-0391: Jingle Encrypted Transports, as well as XEP-0396: JET-OMEMO:

* Removed redundant `name=` attribute from the security element
* Specify content of the `<size/>` element
* Move cipher list from XEP-0391 to profile XEPs to enable profiles to specify own ciphers
* Describe simple JET security element for use with SCE
* Bump XEP-0391 to 0.2.0
* Bump XEP-0396 to 0.2.1

### DIFF
--- a/xep-0396.xml
+++ b/xep-0396.xml
@@ -35,6 +35,16 @@
   <discuss>jingle</discuss>
   &paulschaub;
   <revision>
+    <version>0.2.1</version>
+    <date>2019-10-04</date>
+    <initials>ps</initials>
+    <remark>
+      <p>Declare usable ciphers</p>
+      <p>Remove redundant name attribute from examples</p>
+      <p>Specify handling of AES-GCM auth tag</p>
+    </remark>
+  </revision>
+  <revision>
     <version>0.2.0</version>
     <date>2018-12-06</date>
     <initials>XEP Editor (jsc)</initials>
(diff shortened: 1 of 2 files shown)


horazont:
In addition I’m not sure if ``urn:xmpp:ciphers:`` should really be that, or:

- if it should be scoped to jingle in some way, or
- if it should simply be plain strings without any URN stuff (+ specifying that any non-standard extensions should use [Clark’s Notation](http://www.jclark.com/xml/xmlns.htm))

horazont:
Also, you have syntax errors which you need to fix. Please review the travis build log.

xep-0396.xml
```diff
@@ -34,6 +34,16 @@
   <registry/>
   <discuss>jingle</discuss>
   &paulschaub;
+  <revision>
+    <version>0.2.1</version>
```
horazont: @vanitasvitae Adding the cipher specification (and, in a later revision, the Registry entry) is not purely editorial, so you need to bump the version to 0.3.0 at least.

horazont:
@vanitasvitae Assign back to me once you’re done.

xep-0396.xml
vanitasvitae: Alright :)

PR #840, opened 2019-10-06
XEP-0060: Clarify unlimited behaviour for node config

iNPUTmice:
Not being able to set max_items to 'undefined' or 'unlimited' prevents publish_options to work with Bookmarks 2.

### DIFF
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -50,6 +50,12 @@
   &stpeter;
   &ralphm;
   <revision>
+    <version>1.17.0</version>
+    <date>2019-10-06</date>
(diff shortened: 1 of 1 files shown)


Flowdalic:
Instead of
`max_items="max"`
I think that
`max_items="unlimited"`
is better readable makes the intended semantic more clear.

iNPUTmice:
Council strongly suggested max.

Also it is not technically unlimited because the server might have a limit.

Ppjet6:
Voted by council on 2019-10-30:
https://mail.jabber.org/pipermail/standards/2019-November/036596.html

PR #846, opened 2019-10-19
XEP-0001: Replace Organizational type with Procedural

Ppjet6:
There's no "Procedural" type where there should be, and "Organizational" doesn't
seem to be used anywhere. Probably because historical reasons.


Ppjet6:
Merged as 2ac07d512739259fbacee5fca291c4f0c8c04bf8

PR #848, opened 2019-10-30
XEP-0292: Recommend using contact bare JIDs as item IDs

Zash:
Similar to XEP-0402. Allows easy association between items and roster
entries.

Relevant Standards threads:
https://mail.jabber.org/pipermail/standards/2011-October/025307.html
https://mail.jabber.org/pipermail/standards/2019-January/035710.html


horazont:
@stpeter Can you comment please?

stpeter:
This seems reasonable!

PR #852, opened 2019-11-07
XEP-0158: Use the correct FORM_TYPE for extended IBR, and use a proper MUC join as an example

linkmauve:
This was found while implementing IBR CAPTCHA in [Renga](https://github.com/HaikuArchives/Renga) at the [Stockholm sprint](https://wiki.xmpp.org/web/Sprints/2019_September_Stockholm).

### DIFF
--- a/xep-0158.xml
+++ b/xep-0158.xml
@@ -405,7 +405,7 @@
   <query xmlns='jabber:iq:register'>
     <x xmlns='jabber:x:data' type='form'>
       <field type='hidden' var='FORM_TYPE'>
-        <value>urn:xmpp:captcha</value>
+        <value>jabber:iq:register</value>
       </field>
       <field type='hidden' var='challenge'><value>F3A6292C</value></field>
       <field type='hidden' var='sid'><value>reg1</value></field>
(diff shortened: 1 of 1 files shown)


Ppjet6:
Merged as b107617. Thanks!

zinid:
Well, putting 'jabber:iq:register' there is also wrong, because CAPTCHA related fields are not registered in its form type: https://xmpp.org/extensions/xep-0077.html#registrar-formtypes

linkmauve:
https://xmpp.org/extensions/xep-0158.xml#registrar-formtypes-register is the process to add new fields to an existing FORM_TYPE’s registrar, you should be looking at the registrar instead of the original XEP (which shouldn’t get modified every time an extension adds something to its form).

zinid:
@linkmauve well I think this is just plain wrong of course.

linkmauve:
Well, according to [XEP-0001 §14](https://xmpp.org/extensions/xep-0001.xml#registrar), it definitely is the registrar’s function to register form fields, and in this case they have done so already: https://xmpp.org/registrar/formtypes.html#jabber:iq:register

PR #854, opened 2019-11-12
XEP-0045: Direct messages SHOULD be used over PMs in non-anonymous rooms

ge0rg:
I'm the last person to get rid of PMs, but I think they have no place in non-anon MUCs. Needs Council, obviously.

### DIFF
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -46,6 +46,12 @@
   <registry/>
   &stpeter;
   <revision>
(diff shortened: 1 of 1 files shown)


ge0rg:
This was [discussed and rejected by Council on 2019-11-13](https://mail.jabber.org/pipermail/standards/2019-November/036630.html). I'm preparing a second attempt with less strict wording and more rationale.

PR #856, opened 2019-11-16
WIP: MIX-PAM: Replace roster mechanism by private PEP node

lnjX:
WIP: https://mail.jabber.org/pipermail/standards/2019-November/036628.html


lnjX:
oops, didn't know this was still open, the solution was to stay with the roster extension

PR #862, opened 2019-12-19
Simplify Makefile and make it more portable

SamWhited:
I've been experimenting again with making the Makefile less of a tangled web of overlapping targets and allowing it to be used with non-GNU variants of Make.
This Makefile replicates much of the functionality of the old Makefile but is simpler and in theory should be more portable (it has been tested with GNU Make and NetBSD Make).

Let me know if there is any functionality missing that you used (although I deliberately left out a few rules which are covered by the more general versions of other rules now), or you find something broken on GNU Make.


SamWhited:
The build failures can be fixed by merging #863 first and allowing the container to rebuild on docker hub. CI doesn't currently test the local Dockerfile, which could be an issue.

Ppjet6:
You should be able to rebase :)

SamWhited:
Done, sorry for the delay, not sure why I didn't get an email from this issue.

horazont:
@SamWhited The build still fails.

SamWhited:
Looks like something in the Python, I don't have time to debug our hacky python infrastructure though, so unfortunately I'll have to close this for now.

PR #876, opened 2020-01-11
XEP-0001: Remove impossible guarantee from XSF Objectives

moparisthebest:


horazont:
I’d like to put this in context. The context of this is the discussion about XEPs which potentially currently require the use of GPL-licensed code or a provable clean-room design to be implemented. The apparent view of the community is that this constitutes an "encumberence" and would thus violate the objective removed by this PR.

@moparisthebest is, as I got it, of the opinion that GPL-licensed code should not count as encumberence. While I can understand where that is coming from, dropping the entire objective instead of clarifying that code licensing or at least some type of code licensing is not an issue is the wrong way to go about this.

winfried:
After reading the discussion leading to this patch, I am wondering about one thing: is there any objective against writing an OMEMO library with a more permissive license then GPL? The Python OMEMO library for example depends on two other libraries and those are MIT and Apache2/BSD licensed. If it is possible to write an OMEMO library with a more permissive license, then I see no conflict between the fourth principle from XEP-0001 and a XEP describing OMEMO.

horazont:
Please keep the discussion to editorial topics. If you want to discuss the meaning of this patch and the context, please open a thread on the [members mailing list](https://mail.jabber.org/mailman/listinfo/members) or take it [to the XSF muc](https://xmpp.org/chat?xsf).

stpeter:
These are objectives, not promises. I would prefer:

> To guarantee, as far as possible, that any person or entity can implement the protocols without encumbrance.

ralphm:
Board has discussed this PR and is aware of the broader discussions on our IPR Policy, the definition of Open Standards, and potentially allowing specifications that violate Objective 4 and/or the IPR Policy. In that light we'd like to see rough consensus to emerge from those discussions, and then suggest changes to this PR, or close it in favour of another PR that reflects said rough consensus.

horazont:
References:

- [Discussion about this PR on members@](https://mail.jabber.org/pipermail/members/2020-January/009076.html)
- [Board meeting logs from xsf@](https://logs.xmpp.org/xsf/2020-01-16#2020-01-16-362864a54e0eff07)
- [Related discussion about how to handle non-"Open" XEPs on members@](https://mail.jabber.org/pipermail/members/2020-January/009085.html)

moparisthebest:
We've discussed this quite a bit, and there is a plan to put in a PR for XEP-0001 clarifying things, no use considering this anymore until that is done, thanks all!

PR #878, opened 2020-01-15
XEP-0384: Weaken links to SignalProtocol

alohapersona:
- Clarify that use of SignalProtocol is not final
- Only experimental implementations use SignalProtocol
- Lower risk of IPR issues by not explicitly saying that SignalProtocol shall be used

### DIFF
--- a/xep-0384.xml
+++ b/xep-0384.xml
@@ -28,6 +28,12 @@
     <jid>andy@strb.org</jid>
   </author>
   <revision>
+    <version>0.3.1</version>
+    <date>2020-01-15</date>
+    <initials>ap</initials>
+    <remark>Weaken links to SignalProtocol</remark>
+  </revision>
+  <revision>
     <version>0.3.0</version>
     <date>2018-07-31</date>
     <initials>egp</initials>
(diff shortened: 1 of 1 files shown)


Ppjet6:
@strb, editors need an ACK from you before merging this.

strb:
I honestly don't really see the point of this change. Irrespective of the whole argument about whether the current reliance on signal protocol is desirable or not, this doesn't actually fix anything, it's just playing semantics. 

The change doesn't even go to the effort of re-writing the spec to use clear language throughout, and instead opts for adding just a disclaimer at the top, basically stating that "when we say signal protocol, we don't mean signal protocol". This doesn't "clarify" anything, it does just the opposite, especially the comment about "experimental" implementations. Even I am not actually sure what that is supposed to mean.

I don't think this should be merged, because I don't think it actually helps anybody to understand the protocol or its current use. The correct way to address these questions is to fix the underlying issue by properly adapting and specifying the parts that are still under-specified.

horazont:
@strb Thank you for the feedback. Closing this.

PR #881, opened 2020-01-20
SIG-E2EE: Only elected XSF members may act as embassadors

vanitasvitae:
As requested by Council, only elected members of the XSF are now able to represent the XSF to external entities.


horazont:
Going to apply this manually after acceptance.

PR #887, opened 2020-01-29
XEP-0431: No More Beer

dwd:
Kev and MattJ argued that Beer constituted an encumbrance.

This changes the language around there to simply use normative language to assert server developers can't be silly.


guusdk:
Glad to see you are having fun on the Eurostar.

PR #896, opened 2020-02-21
Add a Design Considerations section to the template

horazont:
Having the result and a summary of long or complicated discussions in the document itself instead of list or chat archives serves future readers.


Ppjet6:
Looks good to me! I also think this is a really good addition and is often missing in XEPs.

PR #898, opened 2020-02-27
XEP-0060: Specify that empty <item/> are invalid on publish

horazont:
Note that to trigger a notify-only publish, the ``<item/>`` must not
be included at all (as said at the beginning of #publisher-publish-request).

### DIFF
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -50,6 +50,12 @@
   &stpeter;
   &ralphm;
   <revision>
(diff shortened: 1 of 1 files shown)


edhelas:
That's a good change indeed :) Especially with the namespace check.
I'll then try to do another PR on https://xmpp.org/extensions/xep-0277.html#microblog_node_config to enforce Atom1.0 namespace in published items.

PR #904, opened 2020-03-10
XEP-0284: Various fixes, and add myself as an author

linkmauve:
I’m working on an implementation of this XEP, and I’d like to push it forward towards Draft in not so long, being an author will let me do that properly.


Ppjet6:
@stpeter, we need an ACK from authors to accept the PR! I don't know the handles for the other authors, happy if somebody can chime in.

linkmauve:
Fyi, Joonas Govenius’s email address doesn’t exist on Google’s servers anymore, and Tom Pusateri’s rejects emails sent from my server, so I couldn’t reach them.

pusateri:
I can take a look at this pull request. Not sure why email bounced but try again.

stpeter:
@pusateri I believe it's the message to Joonas that bounced. Thanks for taking a look!

linkmauve:
Hi, I just rebased, but does it really make sense to squash?  I usually prefer to keep the commits as is when they are self-contained but related to the same XEP.

PR #905, opened 2020-03-11
XEP-0371: migrate from RFC 5245 to 8445

Ri0n:
Changes:

* RFC 5245 is replaced with RFC 8445
* Introduced `ice2` transport attribute for backward compatibility, same as in RFC 8445
* Clarified ICE restart procedure
* Make `remote-candidate` MUST when `urn:ietf:rfc:3264` is advertised by the responder
* Send `remote-candidate` for all components at once to be better compatible with SIP gateways.
* Wrong reference to RFC 6455 was replaced with correct one: RFC 6544
* Allow gathering-complete element to be combined with remaining candidates

### DIFF
--- a/xep-0371.xml
+++ b/xep-0371.xml
@@ -17,7 +17,7 @@
   <dependencies>
     <spec>XMPP Core</spec>
     <spec>XEP-0166</spec>
-    <spec>RFC 5245</spec>
+    <spec>RFC 8445</spec>
     <spec>RFC 6544</spec>
     <spec>draft-ietf-ice-trickle</spec>
   </dependencies>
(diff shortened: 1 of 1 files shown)


Ri0n:
I'll update the PR to make `remote-candidates` mandatory as the related rfc states.

Ri0n:
Since xmpp.org is down I'm happy to have any feedback right here.

dwd:
Paging @stpeter

Ri0n:
ping

stpeter:
These changes seem fine (I might send the author an email or provide a PR with a few editorial suggestions). Because we're adding the 'ice2' attribute, I wonder if we need a namespace bump - especially because aggressive nomination is no longer supported (however, 'ice2' defaults to false, so this might not be necessary).

Ri0n:
ping @horazont

horazont:
@Ri0n Eh, sorry, this slipped through. Fun times.

So, problem here: you added another commit after the approval. In addition, the commit edits the already-existing version block, which is not allowed.

I would suggest that you remove that commit and create a separate PR for the gathering changes there.

Ri0n:
I was asked in @xsf to add another commit to change the date for changelog. I did.

Do you mean I have to split this one into two? First one - all the changes but without "changes" part and another other just "changes"?

Ri0n:
Looks like force-push showed wrong history here on github

horazont:
@Ri0n The one problem is that the newest commit must be the one adding the revision block, because that is the one which will be tagged to constitute the release of the new version. So you could squash them. The other problem is that the commit where you changed the date does more than just changing the date: https://github.com/xsf/xeps/pull/905/commits/c589749dd48df7122b4d832a221e037599619df9

So this is what confuses me; and this seems to be another non-editorial change we should ask the original authors about. Then again, I’m still pretty tired so I wonder if I’m missing something here.

Ri0n:
Probably I did git --amend instead of a new commit. To be honest I don't remember why,
But those extra changes in the latest commit were definitely there  before approval.

PR #907, opened 2020-03-11
XEP-0380: EME: Add omemo:1 as encryption method

lnjX:
omemo:1 is incompatible with OMEMO <= 0.3

### DIFF
--- a/xep-0380.xml
+++ b/xep-0380.xml
@@ -25,6 +25,14 @@
   <shortname>EME</shortname>
   &linkmauve;
   <revision>
+    <version>0.4.0</version>
+    <date>2020-03-11</date>
+    <initials>lnj</initials>
+    <remark>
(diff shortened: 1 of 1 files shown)


lnjX:
@Ppjet6 `Needs Version Block`? - I already added a version block.

Ppjet6:
oops! Thanks!

linkmauve:
Hi, and thanks for this PR!

But I don’t feel like this belongs to the XEP, instead the next version will request a registry from the registrar so that we don’t have to update the XEP every time we had a new encryption mechanism.

lnjX:
@linkmauve Yes that's the better solution!

Ppjet6:
I'm reopening this just to keep track. There's some other work also blocking on the registry atm..

lnjX:
#1127 was merged without using the registrar. However I think I can close this now?

Neustradamus:
@linkmauve, @Ppjet6: @lnjX has done a question, what do you think?

PR #908, opened 2020-03-12
XEP-0384: OMEMO Encryption

melvo:
Improvements

### DIFF
--- a/xep-0384.xml
+++ b/xep-0384.xml
@@ -173,7 +173,7 @@
     </p>
     <p>
       While in the future a dedicated key server component could be used to distribute
-      key material for session creation, the current specification relies on &xep0163; to publish
+      key material for session creation, the current specification relies on &xep0060; and &xep0163; to publish
       and acquire key bundles.
     </p>
   </section2>
--- a/xep-0384.xml
+++ b/xep-0384.xml
@@ -508,7 +508,7 @@
   <section2 topic='Receiving a message' anchor='usecases-receiving'>
     <p>When an OMEMO element is received, the client MUST check whether there is a &lt;keys&gt; element with a jid attribute matching its own bare jid and an inner &lt;key&gt; element with a rid attribute matching its own device id. If this is not the case the message was not encrypted for this particular device and a warning message SHOULD be displayed instead. If such an element exists, the client checks whether the element's contents are an OMEMOKeyExchange.</p>
     <p>If this is the case, a new session is built from this received element. The client MUST then republish their bundle information, replacing the used PreKey, such that it won't be used again by a different client. If the client already has a session with the sender's device, it MUST replace this session with the newly built session. The client MUST eventually delete the private key belonging to the PreKey after use (this is subject to the <link url="#business-rules">Business rules</link>).</p>
-    <p>If the element's contents are a OMEMOAuthenticatedMessage, and the client has a session with the sender's device, it tries to decrypt the OMEMOAuthenticatedMessage using this session. If the decryption fails or there is no session with the sending device, a warning message SHOULD be displayed instead. Also refer to the section about recovering from broken sessions in the <link url="#business-rules">Business Rules</link>.</p>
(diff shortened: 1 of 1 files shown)


xep-0384.xml
```diff
@@ -173,14 +173,14 @@
     </p>
     <p>
       While in the future a dedicated key server component could be used to distribute
-      key material for session creation, the current specification relies on &xep0163; to publish
+      key material for session creation, the current specification relies on &xep0060; and &xep0163; to publish
```
Syndace: This isn't necessary but doesn't hurt either I guess

xep-0384.xml
```diff
@@ -481,10 +481,10 @@
...
-        A &lt;key&gt; element has an attribute named 'rid' referencing the device id of the recipient device, and an attribute named 'kex' which defaults to 'false' and indicates if the enclosed encrypted message includes a key exchange. The ciphertext that is the key and HMAC encrypted using the long-standing OMEMO session for that recipient device is encoded using base64 and placed as text content into the &lt;key&gt; element.
-        The ciphertext that is the encrypted &content; element is encoded using base64 and placed as text content into the &payload; element.
+        A &lt;key&gt; element has an attribute named 'rid' referencing the device id of the recipient device, and an attribute named 'kex' which defaults to 'false' and indicates if the enclosed encrypted message includes a key exchange. The ciphertext containing the key and HMAC encrypted using the long-standing OMEMO session for that recipient device is encoded using base64 and placed as text content into the &lt;key&gt; element.
+        The ciphertext with the encrypted &content; element is encoded using base64 and placed as text content into the &payload; element.
```
Syndace: I don't think this makes things clearer and it technically also isn't correct. The ciphertext doesn't _contain_ the encrypted element, the ciphertext _is_ the encrypted element. `that is` is the better choice in both cases IMO.

xep-0384.xml
```diff
@@ -481,8 +481,8 @@
...
-        A &lt;key&gt; element has an attribute named 'rid' referencing the device id of the recipient device, and an attribute named 'kex' which defaults to 'false' and indicates if the enclosed encrypted message includes a key exchange. The ciphertext that is the key and HMAC encrypted using the long-standing OMEMO session for that recipient device is encoded using base64 and placed as text content into the &lt;key&gt; element.
-        The ciphertext that is the encrypted &content; element is encoded using base64 and placed as text content into the &payload; element.
+        A &lt;key&gt; element has an attribute named 'rid' referencing the device id of the recipient device, and an attribute named 'kex' which defaults to 'false' and indicates if the enclosed encrypted message includes a key exchange. The ciphertext containing the key and HMAC encrypted using the long-standing OMEMO session for that recipient device is encoded using base64 and placed as text content into the &lt;key&gt; element.
```
melvo:
```suggestion
        A &lt;key&gt; element has an attribute named 'rid' referencing the device id of the recipient device, and an attribute named 'kex' which defaults to 'false' and indicates if the enclosed encrypted message includes a key exchange. The key and HMAC encrypted using the long-standing OMEMO session for that recipient device are encoded using base64 and placed as text content into the &lt;key&gt; element.
```

xep-0384.xml
```diff
@@ -481,8 +481,8 @@
...
+        The ciphertext with the encrypted &content; element is encoded using base64 and placed as text content into the &payload; element.
```
melvo:
```suggestion
        The encrypted &content; element is encoded using base64 and placed as text content into the &payload; element.
```

xep-0384.xml
melvo: OK, thanks.
That was also my first understanding of it.
The problem was that those two lines were too long and complex.
Therefore, I thought that the first change would be what was intended by the authors.
The new suggestion should simplify the text without changing the meaning of the sentences.

Syndace: Thanks, I like the new suggestions :+1:

Syndace:
Looks good to me now *author approval*

PR #913, opened 2020-03-25
XEP-0068: Clarify that FORM_TYPE does not need to be type='hidden' in submission forms

mar-v-in:
The specification did not match actual usage and examples in this and other XEPs for submission forms.

Rendered diff: http://www.aptest.com/standards/htmldiff/htmldiff.pl?oldfile=https://xmpp.org/extensions/xep-0068.html&newfile=https://larma.de/xeps/xep-0068.html

### DIFF
--- a/xep-0068.xml
+++ b/xep-0068.xml
@@ -24,6 +24,12 @@
   &hildjj;
   &stpeter;
   <revision>
+    <version>1.2.1</version>
+    <date>2020-03-25</date>
+    <initials>mw</initials>
+    <remark>Clarify that FORM_TYPE does not need to be type=&quot;hidden&quot; in submission forms.</remark>
+  </revision>
+  <revision>
     <version>1.2</version>
     <date>2012-05-28</date>
     <initials>psa/jjh</initials>
@@ -115,7 +121,7 @@
 
 <section1 topic='Use Cases' anchor='usecases'>
     <section2 topic='Unspecified Form' anchor='usecases-unspecified'>
-      <p>These are forms that do not have a hidden field of name FORM_TYPE.
+      <p>These are forms with type=&quot;form&quot; or type=&quot;result&quot; that do not have a hidden field of name FORM_TYPE or forms with type=&quot;submit&quot; that do not have any field of name FORM_TYPE.
         Existing processing rules still apply.</p>
       <example caption='Message with no FORM_TYPE'><![CDATA[
 <message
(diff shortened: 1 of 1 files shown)


xep-0068.xml
```diff
@@ -115,7 +121,7 @@
 
 <section1 topic='Use Cases' anchor='usecases'>
     <section2 topic='Unspecified Form' anchor='usecases-unspecified'>
-      <p>These are forms that do not have a hidden field of name FORM_TYPE.
+      <p>These are forms with type=&quot;form&quot; or type=&quot;result&quot; that do not have a hidden field of name FORM_TYPE or forms with type=&quot;submit&quot; that do not have any field of name FORM_TYPE.
```
Flowdalic: nit: you don't need to use &quot; here and in every place you introduced it with this change.

xep-0068.xml
```diff
@@ -192,7 +198,7 @@
 ]]></example>
   </section2>
   <section2 topic='IQ Example' anchor='usecases-IQ'>
-    <p>The following example shows a user's interaction with a Multi-User Chat room in order to register with the room.</p>
+    <p>The following example shows a user's interaction with a Multi-User Chat room in order to register with the room. Note that during submission of the form, the FORM_TYPE field does not need to include the type=&quot;hidden&quot; attribute.</p>
```
Flowdalic: Maybe explain why: "… because of the rules defined in 'usescases-incorrect'"

xep-0068.xml
```diff
@@ -171,7 +177,7 @@
     </section2>
     <section2 topic='Incorrectly Specified FORM_TYPE' anchor='usecases-incorrect'>
 
-    <p>If the FORM_TYPE field is not hidden, it MUST be ignored as a context indicator.</p>
+    <p>If the FORM_TYPE field is not hidden in a form with type=&quot;form&quot; or type=&quot;result&quot;, it MUST be ignored as a context indicator.</p>
```
Flowdalic: Maybe elaborate on the consequence of this: "As consequence, FORM_TYPE fields MAY omit the hidden attribute for 'submit' type forms."

xep-0068.xml
mar-v-in: I actually do know, but did nonetheless for consistency with how it was before and because I wasn't sure if any tooling might have issues without.

xep-0068.xml
```diff
@@ -23,6 +23,12 @@
   <registry/>
   &hildjj;
   &stpeter;
+  <revision>
+    <version>1.2.1</version>
```
Ppjet6: This is fine if it's editorial changes, otherwise this needs to be 1.3.0. I'll let you ack/nack this :)

xep-0068.xml
horazont: If anything, tooling is more unhappy *with* `&quot;`. As a general rule of thumb, anything which is valid XML goes, though :).

horazont:
Definitely non-editorial since it effectively changes the meaning of some forms.

horazont:
This was vetoed by council in the vote started on 2020-04-01.

The rationale(s) can be found here:

- https://mail.jabber.org/pipermail/standards/2020-April/037251.html

PR #918, opened 2020-04-01
XEP-0436: Updates based on list feedback

jcbrand:
- Create a separate tag for sending the version number
- Include a reset token
- Remove the "additional measures" section and merge parts into "Requirements"
- Mandate that presences are always sent for affiliated users

### DIFF
--- a/xep-0436.xml
+++ b/xep-0436.xml
@@ -25,6 +25,20 @@
   &jcbrand;
   &mwild;
   <revision>
+    <version>0.2.0</version>
+    <date>2020-05-10</date>
+    <initials>jcb</initials>
+    <remark>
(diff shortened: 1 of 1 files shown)


jcbrand:
Adding the "needs changes" label because I have more changes planned before merging.

Flowdalic:
Glad that you are actively developing this specification. However, this ProtoXEP is currently proposed and under vote by council. Making changes to it is usually discouraged until council has voted to avoid ambiguity about the concrete XEP version being voted about.

jcbrand:
Thanks for the explanation @Flowdalic

jcbrand:
In the meantime, Council has accepted the protoXEP and it's been awarded the number XEP-0436

PR #921, opened 2020-04-03
xep-pubsub-mam: Split off from XEP-0313

mwild1:
PR 2 of 3 for XEP-0313 0.7

This is the pubsub querying protocol, split out into a separate document per community consensus.


edhelas:
A big +1 on that one

PR #922, opened 2020-04-03
XEP-0313: new revision 0.7

mwild1:
PR 3 of 3 for XEP-0313 0.7

Due to the well-established nature of this protocol at this point in time, I will be seeking feedback on list before considering this PR ok to merge.

### DIFF
--- a/xep-0313.xml
+++ b/xep-0313.xml
@@ -10,7 +10,7 @@
   <abstract>This document defines a protocol to query and control an archive of messages stored on a server.</abstract>
   &LEGALNOTICE;
   <number>0313</number>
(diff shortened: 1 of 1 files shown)


Ppjet6:
I'll add "Needs Author" so that we know. Ping us when you want to merge

mwild1:
Editors, I consider this (and the two related PRs) ready to merge!

mwild1:
Actually I've now had enough feedback that I would rather continue working on some more changes first.

mwild1:
Having incorporated the bit of feedback from the list, I now consider this ready to merge.

PR #923, opened 2020-04-06
XEP-OMEMO (384): add hint to PubSub 'persistent-items' feature

Flowdalic:
OMEMO implementations ideally want to determine if the PEP service is
suitable for OMEMO. This includes a xep30 feature discovery. This
change makes it easier for implementors to determine the features to
look out for by explicitly mention and linking to the
'persistent-items' feature


Syndace:
Sorry for letting you wait that long @Flowdalic :)
Looks like we finally have rough consensus that we like this change (following recent discussion in the context of #979).

That means @editors: _Author Approval_ :)

PR #926, opened 2020-04-15
XEP-0045: Clarify that the 307 status code should not be used alongsi…

mwild1:
…de 333 for user disconnects

Per discussion in xsf@ about the UX implications of showing user disconnects as kicks.

PR #927, opened 2020-04-15
XEP-0333 v0.4 - add MUC usage info

mwild1:


xep-0333.xml
```diff
@@ -236,6 +242,48 @@
...
+</example>
+
+<example caption='Example marker response to a markable MUC message'>
+<![CDATA[
+<message
+    from='coven@chat.shakespeare.lit/secondwitch'
+    to='coven@chat.shakespeare.lit'
+    id='message-2'
+    type='groupchat'
+    >
```
lnjX, line 278:
```suggestion
    type='groupchat'>
```

This fits better to the formatting of the other examples in other XEPs.

mwild1: Fixed, thanks!

Ppjet6:
Seems that only half of this requires council. I assume one won't go without the other though.

Ppjet6:
After talking with @mwild1, it seems the commit of #926 is also included in this but both PRs are separate. Only the XEP-0333 should be merged here.

Ppjet6:
This was cherry-picked in 928513d81d3b351f5c10861daeaa59274bb439b2.

PR #929, opened 2020-04-19
XEP-0389: major updates to add more flexibility

SamWhited:
A while back @Zash  reached out to me to ask if extensible IBR could be made compatible
with [XEP-0401: Easy User Onboarding]. While working on a solution to this
I decided it would also be nice if it could also allow the user to pick
between several flows (eg. "Verify by Phone" vs  "Verify by SMS").
After working this up I came to realize that the current version of the
spec could not be used to replace [XEP-0077: In-Band Registration] as it is used by [XEP-0045] and
other XEPs because I wasn't using routeable stanzas that could be sent to
components and had no way to discover support other than during stream
feature negotation.

This PR turned into an overhaul of the proposal to fix all these
problems and move the spec back out of experimental.

**Rendered:** https://blog.samwhited.com/xeps/xep-0389.html

[XEP-0401: Easy User Onboarding]: https://xmpp.org/extensions/xep-0401.html
[XEP-0077: In-Band Registration]: https://xmpp.org/extensions/xep-0077.html
[XEP-0045]: https://xmpp.org/extensions/xep-0045.html

### DIFF
--- a/xep-0389.xml
+++ b/xep-0389.xml
@@ -18,7 +18,7 @@
   </abstract>
   &LEGALNOTICE;
   <number>0389</number>
-  <status>Deferred</status>
+  <status>Experimental</status>
   <type>Standards Track</type>
   <sig>Standards</sig>
   <approver>Council</approver>
@@ -32,6 +32,18 @@
   <shortname>ibr2</shortname>
   &sam;
   <revision>
+    <version>0.3.0</version>
+    <date>2020-04-19</date>
+    <initials>ssw</initials>
+    <remark>
+      <ul>
+        <li>Move back to experimental.</li>
+        <li>Add IQ variant of registration/recovery.</li>
+        <li>Allow client to select between multiple flows.</li>
+      </ul>
+    </remark>
+  </revision>
+  <revision>
     <version>0.2.0</version>
     <date>2018-10-01</date>
     <initials>XEP Editor (jsc)</initials>
@@ -42,9 +54,7 @@
     <date>2017-03-16</date>
     <initials>XEP Editor (ssw)</initials>
     <remark>
-      <ul>
-        <li>Move to experimental.</li>
-      </ul>
+      Move to experimental.
     </remark>
   </revision>
   <revision>
@@ -67,9 +77,9 @@
 </header>
 <section1 topic='Introduction' anchor='intro'>
   <p>
-    Historically, registering with an XMPP service has been difficult.  Each
-    server either used customized out-of-band registration mechanisms such as
-    web forms which were difficult to discover, or they used &xep0077; which
+    Historically, registering with an XMPP service has been difficult.
+    Each server either used customized out-of-band registration mechanisms such
+    as web forms which were difficult to discover, or they used &xep0077; which
     could easily be abused by spammers to register large numbers of accounts and
     which allowed for only limited extensibility.
   </p>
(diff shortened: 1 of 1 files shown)


xep-0389.xml
```diff
@@ -275,38 +326,52 @@
...
-    If the client successfully completes the challenge, the server MAY return an
-    empty &lt;success/&gt; element qualified by the 'urn:xmpp:register:0'
-    namespace, at which point it may continue with the stream negotiation
-    process.
+    If the IQ based registration or recovery flow is being used and the server
+    wishes to cancel the flow, it MAY respond to any IQ from the client with the
+    cancel element and type "result".
+  </p>
+  <example caption='Server cancels rquest'><![CDATA[
+<iq type="result" id="bar">
```
lnjX, line 347:
```suggestion
<iq type='result' id='bar'>
```

xep-0389.xml
```diff
@@ -190,23 +202,62 @@
...
+<register xmlns='urn:xmpp:register:0'>
+  <flow id="1"/>
+</register>]]></example>
+  <p>
+    If the client is initiating registration or recovery after a stream has
+    already been initiated it uses the same registration element wrapped in an
+    IQ of type "set".
   </p>
+  <example caption="Client selects a recovery flow after stream negotiation"><![CDATA[
+<iq type="set" id="foo">
```
lnjX, line 252:
```suggestion
<iq type='set' id='foo'>
```

xep-0389.xml
```diff
@@ -190,23 +202,62 @@
...
-    &lt;recovery&gt; element qualified by the 'urn:xmpp:register:0' namespace.
+    replying with an element of the same name and namespace.
+    The element MUST contain a &lt;flow&gt; element that MUST have an "id"
+    attribute matching one of the flows advertised by the server.
+    For example, to select the "Verify by Phone Call" registration flow from
+    the previous example, the client would reply with:
+  </p>
+  <example caption="Client selects a recovery flow"><![CDATA[
+<register xmlns='urn:xmpp:register:0'>
+  <flow id="1"/>
```
lnjX, line 244:
```suggestion
  <flow id='1'/>
```

xep-0389.xml
```diff
@@ -190,23 +202,62 @@
...
+      <challenge>jabber:x:data</challenge>
+    </flow>
+    <flow id="1">
+      <name>Verify by Phone Call</name>
+      <challenge>jabber:x:data</challenge>
+    </flow>
   </register>
   <recovery xmlns='urn:xmpp:register:0'>
-    <challenge>jabber:x:oob</challenge>
+    <flow id="0">
```
lnjX, line 215:
```suggestion
    <flow id='0'>
```

xep-0389.xml
```diff
@@ -190,23 +202,62 @@
     <mechanism>PLAIN</mechanism>
   </mechanisms>
   <register xmlns='urn:xmpp:register:0'>
-    <challenge>jabber:x:data'</challenge>
-    <challenge>pow-example</challenge>
+    <flow id="0">
+      <name>Verify with SMS</name>
+      <challenge>jabber:x:data</challenge>
+    </flow>
+    <flow id="1">
```
lnjX, line 209:
```suggestion
    <flow id='1'>
```

xep-0389.xml
```diff
@@ -190,23 +202,62 @@
     <mechanism>PLAIN</mechanism>
   </mechanisms>
   <register xmlns='urn:xmpp:register:0'>
-    <challenge>jabber:x:data'</challenge>
-    <challenge>pow-example</challenge>
+    <flow id="0">
```
lnjX, line 205:
```suggestion
    <flow id='0'>
```

xep-0389.xml
```diff
@@ -275,38 +326,52 @@
...
     was sent to that email, the server MAY send another challenge.
   </p>
+  <example caption='Server indicates success during steam negotiation'><![CDATA[
+<success xmlns='urn:xmpp:register:0'/>]]></example>
+  <p>
+    If the IQ based flow is being used and the server wishes to indicate success
+    it sends an empty IQ response of type "result".
+  </p>
+  <example caption='Server indicates success after stream negotation'><![CDATA[
+<iq type="result" id="bar" />]]></example>
```
lnjX, line 366:
```suggestion
<iq type='result' id='bar' />]]></example>
```

xep-0389.xml
SamWhited: Doesn't matter, it's the same thing.

SamWhited:
Thought: maybe I should rework this to only use IQs like legacy IBR. For some reason I always forget that stanzas are okay during stream negotiation so long as you only send them to the server, but that was how legacy IBR works and it appears to be fine.

PR #932, opened 2020-04-19
XEP-0384: Version 0.7.0

Syndace:
Another update to the OMEMO XEP containing a lot of fixes, clarifications and general improvements.

The rendered version is available [here](https://syndace.com/xeps/xep-0384.html).


xep-0384.xml
```diff
@@ -294,11 +313,13 @@
...
           <li>Add the ciphertext to the <tt>OMEMOMessage.proto</tt> structure.</li>
-          <li>Serialize the ad and the <tt>OMEMOMessage.proto</tt> structure into a parseable byte array by concatenating ad and the serialized protobuf structure.</li>
-          <li>Calculate the HMAC-SHA-256 using the authentication key and the input material as derived in the steps above.</li>
-          <li>Put the <tt>OMEMOMessage.proto</tt> structure and the HMAC into a new <tt>OMEMOAuthenticatedMessage.proto</tt> structure.</li>
+          <li>Serialize the <tt>OMEMOMessage.proto</tt> structure into a parseable byte array. To avoid potential problems regarding non-uniqueness of the serialization, make sure to only serialize <em>once</em> and to use that exact byte sequence in the following steps.</li>
+          <li>Concatenate the ad and the <tt>OMEMOMessage.proto</tt> structure into a parseable byte array. The result builds the HMAC input material for the next step.</li>
+          <li>Calculate the HMAC-SHA-256 using the authentication key and the input material as derived in the steps above. Truncate the output of the HMAC to 16 bytes/128 bits.</li>
```
vanitasvitae, line 312: Truncate by discarding what portion of bytes?

xep-0384.xml
```diff
@@ -321,7 +345,7 @@
       <li>Use HKDF-SHA-256 to generate 80 bytes of output from the key by providing the key as HKDF input, 256 zero-bits as HKDF salt and &quot;OMEMO Payload&quot; as HKDF info.</li>
       <li>Divide the HKDF output into a 32-byte encryption key, a 32-byte authentication key and a 16 byte IV.</li>
       <li>Encrypt the plaintext using AES-256-CBC with PKCS#7 padding, using the encryption key and IV derived in the previous step.</li>
-      <li>Calculate the HMAC-SHA-256 using the authentication key and the ciphertext from the previous steps.</li>
+      <li>Calculate the HMAC-SHA-256 using the authentication key and the ciphertext from the previous steps. Truncate the output of the HMAC to 16 bytes/128 bits.</li>
```
vanitasvitae, line 341: Again, what part gets discarded?

xep-0384.xml
```diff
@@ -514,11 +530,12 @@
...
         The encrypted &content; element is encoded using base64 and placed as text content into the &payload; element.
       </p>
+      <p>A special case are <em>empty</em> messages, which are used in various places throughout the protocol purely to manage sessions and not to transfer content. With empty messages, the step of creating and encrypting the &payload; element is skipped. Instead of encrypting the key and authentication tag of the &payload; ciphertext with the Double Ratchet session, 32 zero-bytes are encrypted with the Double Ratchet session directly. The resulting OMEMOKeyExchange or OMEMOAuthenticatedMessage are put into &lt;key&gt; elements as usual, but the &payload; element is omitted altogether, so that the &lt;encrypted&gt; element only contains a &lt;header&gt;.</p>
```
vanitasvitae, line 538: Why do we use zeros here? Wouldn't that allow for potential known plaintext attacks?

xep-0384.xml
```diff
@@ -641,9 +658,12 @@
...
-  <p>There are various reasons why decryption of an OMEMOKeyExchange or an OMEMOAuthenticatedMessage could fail. One reason is if the message was received twice and already decrypted once, in this case the client MUST ignore the decryption failure and not show any warnings/errors. In all other cases of decryption failure, clients SHOULD respond by forcibly doing a new key exchange and sending a new OMEMOKeyExchange with a potentially empty SCE payload. By building a new session with the original sender this way, the invalid session of the original sender will get overwritten with this newly created, valid session. This does NOT apply to the actual SCE content. If decrypting the SCE content fails, e.g. because the HMAC does not verify, this is not a reason to forcibly initiate a new key exchange.</p>
+  <p>After receiving an OMEMOKeyExchange and successfully building a new session, the receiving device SHOULD automatically respond with an empty message to the source of the OMEMOKeyExchange. This is to notify the device that the session initiation was completed successfully and that the device can stop sending OMEMOKeyExchanges.</p>
```
vanitasvitae, line 661: "... the receiving device SHOULD automatically respond with an empty *OMEMO* message to the source of the OMEMOKeyExchange."

xep-0384.xml
Syndace: First of all, if you're not working on something super niche, you'll probably never get in touch with a cipher that is (known to be) vulnerable to known plaintext attacks. And even if AES-CBC was vulnerable, all you could find is the key for a single message, which, due to OMEMO's ratcheting, is useless except for that single message, which is known anyway :)

In summary it doesn't really matter what we encrypt here, except that _nothing_ is not an option. And because we need _something_, I felt like fixed zeros for an "empty" message are somewhat clearer/more intuitive than e.g. 32 random bytes. But that's purely subjective I guess.

xep-0384.xml
Syndace: Thanks, will clarify that excess bytes are cut off :)

xep-0384.xml
vanitasvitae: > And even if AES-CBC was vulnerable, all you could find is the key for a single message, which, due to OMEMO's ratcheting, is useless except for that single message, which is known anyway :)

Not sure about that. Since the key for the next message going the same direction is directly derived from the key an attacker would have discovered, they'd potentially gain access to more than just one message.
I'm not an expert in crypto, so I can't tell how likely it is that a known-plaintext attack will affect AES, but it also wouldn't hurt to simply encrypt 32 bytes of randomness, would it?

Syndace: No, the next key is not directly derived from the previous one. One step of the ratchet derives 32 bytes, 16 of which are used for the next derivation, the other 16 are used as the key. So having the key doesn't tell anything about the other 16 that are kept in secret for the next step.

Syndace: (sorry, it's 2x32 not 16)

vanitasvitae: Ah, great! :)

vanitasvitae (approved):

Syndace:
Ready to merge @editors :)

PR #933, opened 2020-04-20
new/reworked protoXEP Quick Response

Syndace:
With the permission of the author I have reworked the protoXEP "Simple Buttons" into a new protoXEP "Quick Response", incorporating the feedback that was given when "Simple Buttons" was rejected.

Can you please put it up for a vote again?

Edit: [rendered version](https://syndace.com/xeps/inbox/quick-response.html)


Zash:
```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

As the author of "XEP-xxxx: Simple Buttons" I am okay with Syndace
picking it up.
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEPlIRnvhTxZZ4279rre2ad7Z60ykFAl6d5+EACgkQre2ad7Z6
0ylMvQ/8C5KWvwV828HJ268vCzcwhnJBu/okonukA22h/jLyH8QKvHXAaFucKoPe
O7hKgXSpQ34WDHSdbJLbh4yBTs0IN29/lAYm6bJNMq8Q1NKklgp8dZjFd1D1ewQC
p4VaOKgOnF3IzrkIANHGnd85eVvOAGai1TvqKZ+eaWWos9q7LO9uzKBoBF8cqbH6
jZTMNHdgHI4McH19ymRHfHIIRpmpUgjwy8bP499Tf4qPqzpr4FW7ITY14/pG9DmK
RNIphCGpaIhn1hIqn3JtvdpIwJ9qpVVsTge+7IUk1lOnevFVQGhcdDA4enC/W+kb
ZAdKsP14iEM/dfBBKzULnEMSgxOS7Ln7s6jG9taRYmsBz+nfJViUfYtthRklcLyt
5GfkSVpXjt3Btb0NlVfqLB+dR1JNzz2g1f5YezB6ECBO7Hjkmu22fOm5qQsQJiKX
KCpEXPiDtekyXUSy+ZxLR7xB03oacG3vUI8sNVAks1ntP7KE9vLZsGRHEBfGNt+r
+jyprY3gAxzMXUywGixjRegFSlILrl3U2KUiZn/PxGb3vMabXi6JWz38xlTvQv66
Qh+4pJ7K/igi5RLNl1P8hSLg734MQ4urVQLPJhRrpPENPif4aGQq/Yij5aKsPYCm
bfr+MLaXwRAEalEN6qihUHvdvQ7EFadXAfFltLepoFGME1Q00LE=
=XRqU
-----END PGP SIGNATURE-----
```

inbox/quick-response.xml
```diff
@@ -0,0 +1,212 @@
...
+
+<example><![CDATA[<allowed-response xmlns="]]>&ns;<![CDATA[">
+  <translation xml:lang="en" value="yes" label="Yes!" />
+  <translation xml:lang="sv" value="ja" label="Ja!" />
+</allowed-response>]]></example>
+
+    <dl>
+      <di>
+        <dt>&lt;translation&gt;</dt>
+        <dd>One for each <tt>&lt;body&gt;</tt> and <tt>@xml:lang</tt> included in the <tt>&lt;message&gt;</tt> this response is a part of. The <tt>@value</tt> is the internationalized textual payload to put into the <tt>&lt;body&gt;</tt> of the message stanza that is sent when this response is selected. The <tt>@label</tt> is an optional internationalized textual label for this response. Clients that offer UI for quick selection of one of the allowed responses MAY refer to this response by label instead of value.</dd>
```
SamWhited, line 90: I don't think the `@attr` syntax is something we normally use in XEPs (I've never seen it anyways). We probably shouldn't introduce new syntax in this XEP.

inbox/quick-response.xml
```diff
@@ -0,0 +1,212 @@
...
+  <allowed-response xmlns="]]>&ns;<![CDATA[">
+    <translation xml:lang="en" value="yes" label="Sure!" />
+  </allowed-response>
+  <allowed-response xmlns="]]>&ns;<![CDATA[">
+    <translation xml:lang="en" value="no" label="Uuuuuuh..." />
+  </allowed-response>
+</message>]]></example>
+
+    <p>A single message MUST NOT contain multiple <tt>&lt;translation&gt;</tt> elements with the same values for both <tt>@xml:lang</tt> and <tt>@value</tt>. The same applies for the combination of <tt>@xml:lang</tt> and <tt>@label</tt>.</p>
+    <p>Clients that receive a message containing allowed responses can offer UI to quickly and conveniently select one of the allowed responses.</p>
```
SamWhited, line 134: nit: this could probably read "MAY offer a UI to quickly"…

inbox/quick-response.xml
```diff
@@ -0,0 +1,212 @@
...
+  </section2>
+
+  <section2 anchor="selecting-a-response" topic="Selecting a Response">
+    <p>When the user selects a response, their client sends a plaintext message body containing the <tt>@value</tt> as <tt>&lt;body&gt;</tt> text. The client uses the <tt>@value</tt> that corresponds to the <tt>&lt;translation&gt;</tt> of the same language as the <tt>&lt;body&gt;</tt>.</p>
```
SamWhited, line 138: I'm not actually sure what this means. It sounds like it should look at the body to pick the language to me, but the body doesn't exist yet. I assume it means the body should match the button pressed?

inbox/quick-response.xml
```diff
@@ -0,0 +1,212 @@
...
+
+<example><![CDATA[<message to="gitbot@example.com">
+  <action-selected xmlns="]]>&ns;<![CDATA[" id="merge" />
+</message>]]></example>
+  </section2>
+</section1>
+
+<section1 anchor="rules" topic="Business Rules">
+  <p>All message bodies SHOULD always list the (internationalized) allowed responses too, so that users of clients that don't support &xepname; can still know what the allowed responses are.</p>
+  <p>Bots or other entities that indicate allowed responses are free to accept variations of the response values for convenience, for example ignoring the casing or accepting abbreviations of response values.</p>
```
SamWhited, line 172: This should probably use normative language since it's in the business rules: "MAY accept variations"

inbox/quick-response.xml
```diff
@@ -0,0 +1,212 @@
...
+  <p>Actions SHOULD only be a quicker way to access a feature that could also be accessed using information in the message body. For example, a bot that notifies about a new merge request includes in its notification message body a link to the web interface where manual merging is possible. An action could offer a more convenient way to merge, without taking the route via the web interface. In summary, users of clients that don't support &xepname; SHOULD still have a way to manually trigger the action.</p>
+  <p>Clients MUST only provide quick responses for the most recently received message that contains text content.</p>
```
SamWhited, line 174: Why is this? I feel like I should be able to scroll back and respond to older messages too personally, but I haven't really thought it through.

inbox/quick-response.xml
```diff
@@ -0,0 +1,212 @@
...
+  <p>Clients MUST only provide action selection for the most recently received message that contains actions. This is slightly different than the rule for quick responses, in that the message defining the action does not have to be the most recent message with text content, but only the most recent message that contains actions.</p>
+  <p>Clients MAY decide to not offer action selection if the message containing the action is not the most recent message with text content.</p>
+  <p>It is up to the client and its specific UI requirements to decide what to do if both quick responses and actions are available, potentionally coming from different messages.</p>
```
SamWhited, line 177: I feel like this should give some suggestions using RFC 2119 language since it's in the business rules section. It's not clear to my what I would do if I were a client dev just from reading this sentence.

inbox/quick-response.xml
Syndace: Thank you so much for reviewing!

I'll only answer meta stuff, i.e. regarding formatting etc. I think discussion regarding the content of XEPs is generally discouraged in the git comments (right?).

The `@attr` was used (once :D) in Simple Buttons and I like how it clearly identifies the thing as an attribute, but if that's not used anywhere else I'll change it.

What do you mean by "preformatted text"? The `<tt>`s I'm wrapping them in?

inbox/quick-response.xml
Syndace: I hardly see the difference between text that doesn't use 2119 at all and text that precedes everything with MAY, but I'll change it (and the other notes regarding 2119 too)

inbox/quick-response.xml
Syndace: Precise English phrasing might not be my strength :D I'll try to rephrase it.

inbox/quick-response.xml
Syndace: Would you say the same if I moved this to implementation notes? Not entirely sure what goes where.

inbox/quick-response.xml
Ppjet6: This comment is rather non-editorial, maybe we can keep this out of the repo :)

Syndace: @SamWhited I'll ping you in xsf@ to discuss this when I'm done with the editorial fixes

inbox/quick-response.xml
Syndace: hmm, I don't agree that this should use normative language, maybe it shouldn't be part of the XEP at all, I'm really unsure. I'll move it to implementation notes for now and then I'll wait for further feedback.

inbox/quick-response.xml
SamWhited: > What do you mean by "preformatted text"? The <tt>s I'm wrapping them in?

yah. I like how it looks, but I don't think it's used anywhere else and don't know if it will show up well printed or anything, so it's worth thinking about using quotes or something instead.

inbox/quick-response.xml
SamWhited: I'm not in xsf@; too much traffic and notifications. Let's just discuss it here, if others want to have a more indepth discussion somewhere else they're welcome to.

inbox/quick-response.xml
SamWhited: I'm really not sure what would be best, maybe someone else will have stronger opinions :)

inbox/quick-response.xml
Syndace: I removed the `<tt>`s for elements, the `<>` already highlight elements quite well. For attribute I'm not sure, I think I'll just leave it at the `<tt>`s and see if someone else has thoughts on this.

inbox/quick-response.xml
Syndace: Okay so the reason for this rule is that we don't have clean references. Say, a bot sends a message with a list of allowed responses, 10 random off-topic messages are sent and then you type `yes`. It's obviously confusing and non-obvious that the `yes` belongs to the question the bot asked 10 messages above. Also the bot has no way to know that the other 10 off-topic messages were not attempts to answer its question. All in all, when discussing this we agreed that limiting responses to only the most recent message leads to the least confusion possible.

SamWhited: Could it not reference the message ID or something? This restriction seems like it would make the XEP almost useless for any of the use cases I was thinking (eg. things like polls).

Syndace: The point of the allowed responses is that they are exactly equivalent to a "normal" text response, no magic. Manually typing "yes" should be 100% equivalent to clicking the "yes" quick response button. You would most probably use an action to implement a poll.

SamWhited: Fair enough. Mandating client behavior here still feels odd to me, because I have a feeling this will be used for something other than bots at some point where we don't care which message the user is responding too, but I see your point and don't have a better suggestion for how to make the fallback work.

Syndace:
Render available [here](https://syndace.com/xeps/inbox/quick-response.html)

PR #934, opened 2020-04-20
XEP-0167: add rtcp-mux element

iNPUTmice:
RTCP-muxing is used in WebRTC. This element is actually used in the wild by clients like Siskin and Movim (and soon Conversations)

Based on the works by @fippo https://wiki.xmpp.org/web/User:Fippo#mapping_a.3Drtcp-mux

I think back in the day people just forgot to add this to the spec.


fippo:
I wondered whether we also included a urn:ietf: namespace but might have only done so in a disco feature. @legastero may recall.

fippo:
https://github.com/jitsi/lib-jitsi-meet/blob/86d69f54892ed8aa20d7cf59c7fca8dbbde7238d/modules/xmpp/xmpp.js#L160
-- hah!

PR #936, opened 2020-04-21
Fix some references in Password Storage ProtoXEP

SamWhited:
This fixes some broken reference text that should have no bearing on the council voting. It would be nice if it could be merged before a number gets assigned (assuming the council votes to do so, if not naturally this PR doesn't matter), but if not feel free to close this and I'll incorporate it into a v0.0.2 or similar of this document.


horazont:
Ported to the accepted XEP as 4f1ba816d4b608503e5536da6754b5425ee8510d.

PR #938, opened 2020-04-22
Move ref target attribute to root node

SamWhited:
XEPs are currently rendered without the URL when referenced from an RFC because target is on a list of formats and not on the root node. This copies the attribute to the root node.

For an example reference see https://xml2rfc.tools.ietf.org/public/rfc/bibxml2/reference.NIST.SP.800-185.xml

### DIFF
--- a/ref.xsl
+++ b/ref.xsl
@@ -35,6 +35,7 @@ THE SOFTWARE.
   <xsl:template match='/'>
     <reference>
       <xsl:attribute name='anchor'><xsl:text>XEP-</xsl:text><xsl:value-of select='/xep/header/number'/></xsl:attribute>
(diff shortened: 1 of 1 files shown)


Flowdalic:
@horazont  asked me to note what I wrote in xsf@

The refs are mostly used by https://xml2rfc.tools.ietf.org/ and can be found at  https://xmpp.org/extensions/refs/

Currently, e.g. e.g. https://xmpp.org/extensions/refs/reference.XSF.XEP-0435.xml has no target,  wheras https://xml2rfc.tools.ietf.org/public/rfc/bibxml2/reference.NIST.SP.800-185.xml has. Hence this LGTM. utACK.

PR #940, opened 2020-04-30
caps: fix complex generation example by adding missing form field type

Flowdalic:
The default form field type is text-single, but the hyptothetical
field ip_version has multiple values and hence can not be of type
text-single.


horazont:
@Flowdalic Good catch. Can I get you to file a PR for https://xmpp.org/extensions/xep-0157.html, too?

PR #949, opened 2020-05-23
XEP-0157: Add status-addresses registrar entry

Ppjet6:
https://mail.jabber.org/pipermail/standards/2020-May/037443.html


Flowdalic:
Ideally the form field's registry submission would include of data form validations, and specify the datatype of xs:anyURI for this added field.

Flowdalic:
Patch at https://github.com/Flowdalic/xeps-xsf/commit/b4a0deb99a3ad9582be5dbfbc933920a478aa4b7

Ppjet6:
I have updated the PR with @Flowdalic's changes.

xep-0157.xml
```diff
@@ -192,6 +201,15 @@
       var='security-addresses'
       type='list-multi'
       label='One or more addresses for communication related to security concerns'/>
+  <field
+      var='status-addresses'
+      type='list-multi'
+      label='One or more addresses for service status'>
+    <validate xmlns='http://jabber.org/protocol/xdata-validate'
```
horazont, line 208: @Flowdalic What is the source for being able to use validation stuff in XEP-0068 form type definitions, the schema of which is, AFAIK, normatively defined in XEP-0068 and does not include any validation stuff?

Flowdalic: It also does not disallow additional elements.

horazont: Do you have a source which makes that argument count for registries?

Flowdalic: No.

But there is also no source for extendability of extension elements in XMPP (I am aware of). Yet it is common practice.

This is in line with this spirit of extensibility of XMPP. It mimics how data form validation extend form fields on the wire protocol on the registry layer. 

I think it is very desirable to have that information there. But if you want to start a generic discussion if this the right way, which I think it clearly is, or if this belongs at all in the registry, which I think it clearly does, then we should continue this discussion on standards@.

Ppjet6:
Accepted during Council meeting on 2020-06-17.

PR #950, opened 2020-05-23
XEP-0414: Reverse SHOULD/MUST for Blake2b algorithms

Ri0n:
While Blake2b is capable to produce digests of any size from 1 to 64 bytes, It's default mode is 64 bytes (512 bits) though.
Some libraries implement just default digest size.

* OpenSSL supports blake2s-256 and blake2b-512 (no blake2b-256)
* gcrypt supports both blake2b-256 and blake2b-512
* nss supports none of blake2b
* Botan - any digest size
* Java: https://github.com/alphazero/Blake2b - any digest size
* Go-lang: https://godoc.org/golang.org/x/crypto/blake2b - both blake2b-256 and blake2b-512
* Rust: https://docs.rs/blake2/0.8.1/blake2/ - any digest size, 512 by default
* JS: https://github.com/dcposch/blakejs - any digest size, 512 by default

Also various libraries based on openssl will provide just blake2b-512.
So it looks to be a preferable choice over blake2b-256

### DIFF
--- a/xep-0414.xml
+++ b/xep-0414.xml
@@ -21,6 +21,12 @@
   <shortname>hashrecs</shortname>
   &jonaswielicki;
   <revision>
+    <version>0.4.0</version>
+    <date>2020-05-23</date>
+    <initials>rion</initials>
+    <remark>Make Blake2b-256 SHOULD and Blake2b 512 MUST</remark>
+  </revision>
+  <revision>
     <version>0.3.0</version>
     <date>2019-08-19</date>
     <initials>jsc</initials>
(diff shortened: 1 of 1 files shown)


xep-0414.xml
```diff
@@ -20,6 +20,12 @@
   <supersededby/>
   <shortname>hashrecs</shortname>
   &jonaswielicki;
+  <revision>
+    <version>0.3.1</version>
```
lnjX, line 24: I think this should be `0.4.0` since it's not an editorial change.

Ri0n:
I don't really know about versioning but updated :)

horazont:
@Ri0n can you please merge your commits together so the change of version number won’t be confusing?

Ri0n:
seems to be too late :)
If the PRs to this repo never assume merging with squashing I'll take it into account for my next PRs.

horazont:
@Ri0n Yeah, I figured out that I can do it myself.

In general, I don’t like auto-squashed commits. If you do it yourself with proper commit messages and boundaries, that’s great though.

PR #955, opened 2020-06-01
DNM: XEP-0393: updates based on LC feedback

SamWhited:
## DO NOT MERGE

This updates XEP-0393 based on LC feedback. In particular it attempts to further clarify how accessibility might be handled (without imposing requirements which will depend on the system in which the spec is being implemented), mention security concerns people had around Markdown parsers, and add an `unstyled` hint.

### DIFF
--- a/xep-0393.xml
+++ b/xep-0393.xml
@@ -27,6 +27,18 @@
   <shortname>styling</shortname>
   &sam;
   <revision>
+    <version>0.3.0</version>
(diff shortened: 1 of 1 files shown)


SamWhited:
Sorry, I should have made this as a draft PR. GitHub doesn't let you change back, apparently though. Please hold off on merging this until discussion on the mailing list has finished up around using word joiners instead.

PR #956, opened 2020-06-02
XEP-0393: updates based on LC feedback

SamWhited:
Resubmitting with registrar updates, a disco feature, and the zero width space technique.

### DIFF
--- a/xep-0393.xml
+++ b/xep-0393.xml
@@ -27,6 +27,18 @@
   <shortname>styling</shortname>
   &sam;
   <revision>
+    <version>0.3.0</version>
+    <date>2020-06-02</date>
+    <initials>ssw</initials>
+    <remark>
+      <p>
+        Add ability to disable styling, further clarify accessibility
+        considerations, and mention that Styling is not Markdown in the security
+        considerations section.
+      </p>
+    </remark>
+  </revision>
+  <revision>
     <version>0.2.2</version>
     <date>2020-05-20</date>
     <initials>ssw</initials>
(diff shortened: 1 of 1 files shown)


xep-0393.xml
```diff
@@ -255,6 +267,19 @@
     </di>
   </dl>
 </section1>
+<section1 topic='Discovering Support' anchor='disco'>
+  <p>
+    Clients that support Extensible IBR MUST advertise the fact by including a
```
horazont, line 272: Extensible IBR? I suppose this was meant to say "Message Styling"?

xep-0393.xml
```diff
@@ -467,24 +525,67 @@
...
   </p>
   <p>
-    Styled text may also be rendered poorly by screen readers.
+    Styled text may be rendered poorly by screen readers.
     When applying formatting it may be desirable to include directives to
-    exclude styling directives from being read.
+    exclude styling directives from being read, or to add markers to the final
+    output that have semantic meaning for screen readers.
+    For example, in an web based client an emphasis span might be converted to
+    an HTML &lt;emph&gt; element.
```
horazont, line 533: The HTML element which I think you’re thinking of is ``<em/>``, not ``<emph/>``.

xep-0393.xml
SamWhited: Oops, copy/pasta. Fixed. Thanks!

xep-0393.xml
SamWhited: Ah thanks, I don't do a lot of HTML but wanted an example people would be familiar with. Should have familiarized myself first. Fixed!

PR #959, opened 2020-06-08
XEP-0156: reorganize stating XRD/JRD requirements

Flowdalic:
The reference to RFC 6120 was incorrect, what this really meant is RFC
6415. But instead of simply s/RFC 6120/RFC 6415/ here, I decided to
reorganize stating the requirements of XRD and JRD a little.

### DIFF
--- a/xep-0156.xml
+++ b/xep-0156.xml
@@ -200,6 +200,7 @@ _xmppconnect IN TXT "_xmpp-client-websocket=wss://web.example.com:443/ws"
   <section2 topic='Business Rules' anchor='httpbizrules'>
     <p>The following business rules apply:</p>
     <ol start='1'>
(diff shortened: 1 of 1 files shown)


xep-0156.xml
```diff
@@ -200,6 +200,7 @@ _xmppconnect IN TXT "_xmpp-client-websocket=wss://web.example.com:443/ws"
   <section2 topic='Business Rules' anchor='httpbizrules'>
     <p>The following business rules apply:</p>
     <ol start='1'>
+      <li>Services implementing this XEP MUST use the XRD and SHOULD additionally provide the JRD format (both formats are specified in &rfc6415;).</li> <!---->
```
horazont, line 203: "``<!---->``" Huh?

horazont: Also maybe `s/use/offer/`?

PR #961, opened 2020-06-10
XEP-0030: Specify that the disco#info feature may not be explicitly set

Flowdalic:


xep-0030.xml
```diff
@@ -283,7 +283,7 @@
...
+      <p>The target entity then MUST either return an IQ result, or return an error (see the <link url="#errors">Error Conditions</link> section of this document). The result MUST contain a &lt;query/&gt; element qualified by the 'http://jabber.org/protocol/disco#info' namespace, which in turn contains one or more &lt;identity/&gt; elements and one or more &lt;feature/&gt; elements. Note: Every entity MUST have at least one identity; however, an entity is not required to return a result and MAY return an error, most likely &feature; or &unavailable;, although other error conditions may be appropriate. Furthermore, entities are NOT REQUIRED to announce the 'http://jabber.org/protocol/disco#info' feature explicitly, as its support is implicitly signalled by returning an non-error IQ to disco#info requests.</p>
```
horazont, line 286: @Flowdalic Why still require at least one ``feature`` element if disco#info is not required?

horazont: Also, ``NOT REQUIRED`` is not part of RFC 2119.

horazont:
This was vetoed by Council in [a vote started on 2020-06-17](https://logs.xmpp.org/council/2020-06-17#2020-06-17-1fb8f50138c320d1).

Summary:

* dwd:

    >  That said, I'm -1 for the reasons jonas’ raiuses on the PR.

* Jonas:

    > -1, unless the author can find real-world issues which are solved by this PR. Specifically, if we don’t require disco#info and then stop requiring at-lesat-one element, it might break strict validators (the existence of which I could endorse given that this is a Final document and at-least-one is a MUST) for no good reason.

* Zash:

    > -1, agree with jonas’ & dwd 

* daniel:

    >  yes I agree with jonas’ here as well

* Ge0rG ([voted on-list here](https://mail.jabber.org/pipermail/standards/2020-June/037594.html)):

    > As stated time and again, neither removing the mandatory 0030 feature, nor adding the 0030 feature to all disco examples in all XEPs is The Right Thing.

PR #962, opened 2020-06-14
Support experiments on GitLab.com

horazont:
This PR configures a GitLab CI pipeline which features the following:

- Building of an nginx image with the XEPs as static files, in all formats.
- Incremental builds on the main branch and incremental builds for MRs based on the last main build.
- Automatic archiving of changed XEPs to the attic
- Automatic announcement to the mailing lists
- Attach rendered versions of changed documents to MRs (Example: https://gitlab.com/xsf/xeps/-/merge_requests/3 click on "exposed artifacts")

This is an experiment. There have been no decisions about a potential move to GitLab.com made yet. We are investigating the use of gitlab.com over github.com for improved CI features.

There will be official communication about this from iteam and editors once gitlab.com actually seems a viable alternative.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,96 @@
+stages:
+- build
+- export
+
+"build@main":
+  image: registry.gitlab.com/xsf/docker-images/xep-buildspace/image:0.1.0
+  stage: build
+  script:
+  - python3 tools/ci-restore-timestamps.py
+  - make html inbox-html inbox-xml pdf xeplist refs xml
+  - bash tools/ci-prune-build.sh
+  rules:
+  - if: '$CI_COMMIT_REF_NAME =~ /^main$/'
+    when: always
+  - when: never
+  cache:
+    key: build-cache
+    paths:
+    - build/
+  artifacts:
+    paths:
+    - build/
+    expire_in: '1 day'
+  resource_group: xep-build
+
+"pack@main":
+  image: docker:19.03.11
+  stage: export
+  services:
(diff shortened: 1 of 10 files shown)


lnjX:
About the gitlab.com migration: Could (anyone with access) please also import the full repository, including the history of issues and pull requests to gitlab? This is easily possible using the github import on gitlab. This way you could link to old issues and pull/merge requests.

horazont:
@lnjX I prefer to discuss this on the mailing list, but we have a choice between "Provide a transition period where you can contribute on both GitHub and GitLab" and "Do a full import".

Let me explain.

#### Do a full import, no transition

This case is clear; we archive this repo, we import it on gitlab, everything’s clear.

#### Do a full import, start transition period

Then we have two repositories where you can contribute via issues and pull requests, both of which share the same history until approximately PR number 965, after which they diverge. This seems oddly confusing to me.

#### Start clean on GitLab, import later

In this case, the initial period is nice, but if we ever turn off GitHub, we would have to overwrite all issue/PR history we have created on GitLab.

Or do you know of a way to import PRs *into* an existing repository?

lnjX:
I guess the easiest way would be to avoid a transition period. If you allow issues and PRs on both sides this will probably get messy. I personally would consider making only some experiments on gitlab (so not allowing production issues/PRs) and then doing a full import.

> Or do you know of a way to import PRs into an existing repository?

You could certainly use the GitLab and GitHub APIs to do that, but that would probably require writing some additional code – probably not what we currently want to do.

PR #963, opened 2020-06-16
XEP-0178: Clarify SASL-EXTERNAL specification when s2s auth fails

woj-tek:
A while back it was discussed that XEP-0178 (SASL-EXTERNAL) for s2s was kinda misleading - it says that server should close connection if authentication fails but it seems that "everyone" (at least Prosody[0] and ejabberd) actually fallbacks to dialback in that case.

This PR changes the specification to reflect status quo.

[0] https://issues.prosody.im/1006


horazont:
@dwd will start a thread about this on the standards list.

PR #969, opened 2020-07-16
XEP-0045 v1.33.0

mwild1:
Per discussion in xsf@ about the UX implications of showing user disconnects as kicks.

*Note: This is a new attempt at #926 - there is only one word changed (as Council requested).*

### DIFF
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -46,6 +46,12 @@
   <registry/>
   &stpeter;
   <revision>
+    <version>1.33.0</version>
+    <date>2020-04-15</date>
+    <initials>mw</initials>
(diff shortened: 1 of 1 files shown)


jcbrand:
@mwild1 Example 44. in XEP-0045 explicitly mentions that both 307 and 333 need to be sent by the server, so you probably need to update that as well.

https://xmpp.org/extensions/xep-0045.html#example-44

mwild1:
@jcbrand the usage in example 44 is a slightly different case. It is responding directly to a client attempting to join the MUC, and isn't broadcast to other occupants.

It's not (in my mind) what 333 was originally intended for, but it's unrelated to the UX issue that this PR is aiming to fix. I tried to keep the changes minimal, given that this is a Draft document.

PR #970, opened 2020-07-17
XEP-0004: fix typo s#&gt;required/&lt;#&lt;required/&gt;

Flowdalic:


horazont:
Holding this one back to batch-merge with either #971 or #972.

PR #971, opened 2020-07-17
XEP-0004: Clarify field type omission for 'submit' and 'result' form field types

Flowdalic:
The specification was a bit involved about that topic. Hopefully this
text improves the readability.

Note that I actually tend to believe it would be better that we only allow omission for 'submit' forms, instead of doing what this PR proposes. But given that we have already a few xep examples that show 'result' forms without types (on non text-single fields), I fear that ship has sailed. I leave it to the wisdom of council to decide how to move forward.

Background is mostly that ejabberd and prosody return xep157 'result' forms with form field types, and Openfire does not. Now the questions that arise include:
- Is Openfire allowed to do that according to the spec?
- Should we recommend implementations to add field types to fields in 'result' forms?
- Should we require implementations to add field types to fields in 'result' forms?


horazont:
Tagged as Needs Council because it competes with #972. Next time, please mention that in the description of both PRs, since this PR as-is would be editorial.

horazont:
This proposal was [vetoed by the Council on 2020-07-22](https://logs.xmpp.org/council/2020-07-22#2020-07-22-b1d9e21b08484448).

- Jonas: essentially [So while the PR reduces the amount of text, it introduces undefined behaviour  for @type="error" [sic!], which I don’t endorse.](https://mail.jabber.org/pipermail/standards/2020-August/037664.html).

Other members followed that rationale.

As was discovered after the meeting (thanks @Flowdalic), that rationale was based on false premises. So another discussion of the PR took place on [2020-08-12](https://logs.xmpp.org/council/2020-08-12#2020-08-12-602e41eaf4969a3d). It was then decided to move ahead with #975 (which bases on this #971), which was discussed on [2020-08-19](https://logs.xmpp.org/council/2020-08-19#2020-08-19-5d38de65405ae15d) and since accepted.

PR #972, opened 2020-07-17
XEP-0004: Clarify that 'result' forms must have explicit field types

Flowdalic:
The specification was a bit involved and unclear if 'result' forms
must come with explicit form field type annotations.

This is the counterproposal to #971.


horazont:
I find your choice of the word "Clarify" disturbing when changing a MAY (and a SHOULD) to a MUST.

xep-0004.xml
```diff
@@ -223,7 +223,7 @@
...
+    <p>If the &lt;field/&gt; element type is anything other than "fixed" (see below), it MUST possess a 'var' attribute that uniquely identifies the field in the context of the form (if it is "fixed", it MAY possess a 'var' attribute). The &lt;field/&gt; element MAY possess a 'label' attribute that defines a human-readable name for the field. For data forms of type "form" and "result", each &lt:field/&gt; MUST possess a 'type' attribute that defines the data "type" of the field data, unless the type is "text-single", in which case the explicit type declaration MAY be omitted (the default is "text-single"). For data forms of type "submit", inclusion of the 'type' attribute is OPTIONAL, since the form-processing entity is assumed to understand the data types associated with forms that it processes.</p>
```
Zash, line 226: > `&lt:`
colon instead of semicolon

horazont:
This proposal was [vetoed by the Council on 2020-07-22](https://logs.xmpp.org/council/2020-07-22#2020-07-22-b1d9e21b08484448):

- Jonas: I am -1 on PR#972, because it changes a strongly-worded RFC 2119 business rule without sufficient rationale and while evidence of behaviour, which would then be non-compliant, exists.
- Daniel: not significant enough to change the normative language of a final xep

PR #973, opened 2020-07-29
XEP-0313: Correct MAM UID "not-found" note

rufferson:
perhaps missed during copy/paste from similar note further below


mwild1:
Looks good, thanks!

PR #976, opened 2020-08-16
XEP-0060: Add missing querytype 'retrieve' action and 'item' key to registry submission

vanitasvitae:
As discussed [in this mailing list thread](https://mail.jabber.org/pipermail/standards/2020-August/037679.html) the PubSub specification is missing registry entries for the 'retrieve' action and the 'item' key.
This PR adds those to the query type registry submission.

PR #977, opened 2020-08-16
XEP-0060: Fix minor typo in form example

lnjX:


horazont:
I removed the revision block because it conflicts with #976 and I want to merge them in one batch.

PR #979, opened 2020-08-29
XEP-0384 clarify server side requirements

rufferson:
As clarified on the list PEP service must _support_ persistance, not must _persist_ (unconditionally).
Also fix a typo in max value for max_items.

### DIFF
--- a/xep-0384.xml
+++ b/xep-0384.xml
@@ -651,9 +651,9 @@
   <section2 topic='Server side requirements' anchor='server-side'>
     <p>While OMEMO uses a Pubsub Service (&xep0060;) on the user’s account it has more requirments than those defined in &xep0163;. The requirements are:</p>
     <ul>
-      <li>The pubsub service MUST persist node items.</li>
+      <li>The pubsub service MUST support persistent node items via 'pubsub#persist_items' node configuration and 'pubsub#persist_items' as a publish option.</li>
       <li>The pubsub service MUST support publishing options as defined in <link url='https://xmpp.org/extensions/xep-0060.html#publisher-publish-options'><cite>XEP-0060</cite> §7.1.5</link>.</li>
-      <li>The pubsub service MUST support 'max' as a value for the 'pubsub#persist_items' node configuration.</li>
+      <li>The pubsub service MUST support 'max' as a value for the 'pubsub#max_items' node configuration and support 'pubsub#max_items' as a publish option.</li>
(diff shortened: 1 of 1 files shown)


vanitasvitae:
FYI: his PR partially intersects with #923

xep-0384.xml
```diff
@@ -651,9 +651,9 @@
   <section2 topic='Server side requirements' anchor='server-side'>
     <p>While OMEMO uses a Pubsub Service (&xep0060;) on the user’s account it has more requirments than those defined in &xep0163;. The requirements are:</p>
     <ul>
-      <li>The pubsub service MUST persist node items.</li>
+      <li>The pubsub service MUST support persistent node items via 'pubsub#persist_items' node configuration and 'pubsub#persist_items' as a publish option.</li>
```
lovetox, line 654: Either you support publish options that means for all configuration features you support, or you dont

Servers announcing that they support publish options, but in reality don’t for all options is a bug on the server, and should not leak into this XEP

So the line

`The pubsub service MUST support publishing options as defined in <link url='https://xmpp.org/extensions/xep-0060.html#publisher-publish-options'><cite>XEP-0060</cite> §7.1.5</link>`

says everything that is important.

Also there is no way for clients to check if a server only implements publish options for some special options

lovetox: So i would not change this line

And for node configuration i would add a second line that says that pubsub#config-node needs to be supported

rufferson: While I agree with  the statement "Servers announcing that they support publish options, but in reality don’t for all options is a bug on the server, and should not leak into this XEP" I've added it just because of the similar line below for 'open' access-model. Or can there be a case where server supports open AM, publish options, but not open AM in publish options?

lovetox: i didnt see that, its the same story also for open access there is a feature announce by the server, pubsub#access-open

What the server needs to support is

pubsub#node-config
pubsub#persist-items
pubsub#max-items
pubsub#access-open
pubsub#publish-options

rufferson: yup ok, let me rephrase those then

mar-v-in: I think one reason why it was phrased this way was that some time ago there was uncertainty what persisted items in XEP-0060 exactly means. At least one major server implementation was using persisted items in the sense that they can be requested at a later point (as described in the XEP) but are not persisted to disk, which meant they may get removed occasionally without further notice on server restart.

In older OMEMO versions, clients were made efforts to handle such servers. This note to me reads that no such efforts need to be made for current OMEMO version.

PR #986, opened 2020-09-24
XEP-0277: pubsub#type MUST be set in bare PubSub

Ppjet6:
Not entirely sure it's valid as is (breaking change?). I'd appreciate council guidance.

### DIFF
--- a/xep-0277.xml
+++ b/xep-0277.xml
@@ -35,6 +35,12 @@
     <uri>https://valeriansaliou.name/</uri>
   </author>
   <revision>
+    <version>0.7.0</version>
+    <date>2020-09-24</date>
+    <initials>pep</initials>
+    <remark>Require pubsub#type to be set in bare PubSub.</remark>
+  </revision>
+  <revision>
     <version>0.6.3</version>
     <date>2019-03-26</date>
     <initials>pep</initials>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Seems like a breaking change. But it's also experimental so it doesn’t matter. And I personally think it seems like a change we should make. Probably needs author approval rather than council?

horazont:
@stpeter You as an author, please comment and/or ping other authors who may know more. Specifically, Council was unsure about whether this should be a MUST/SHOULD.

stpeter:
Hmm, it seems to me that the payload type is actually `http://www.w3.org/2005/Atom` not `urn:xmpp:microblog:0` (which is only the node name).

Ppjet6:
Yes the payload type is actually `http://www.w3.org/2005/Atom` but rules applied to that payload are those of microblog.

Maybe I'm just assuming something on `pubsub#type` that isn't true, it's not that well defined in 0060 after all. The only normative text there is about it is “Payload type”. We can also find in examples:
- “Specify the type of payload data to be provided at this node”
- “The type of node data, usually specified by the namespace of the payload (if any)”

 There is no such thing as “just” pubsub + atom, is what I'm trying to point out here, and I'd like to have a place to express this.

Ppjet6:
With this said, I'd consider microblog a payload type anyway, it's not like Atom was a serialization format, there are rules associated with it. Microblog is mandating behaviour in the node (where I'm setting `pubsub#type`) and in each individual atom payload (required Atom XML elements, etc.)

stpeter:
Hmm. We might need to clarify what pubsub#type means, then. I agree that the processing rules for data coming from `urn:xmpp:microblog:0` nodes might be special, but the payload format is indeed `http://www.w3.org/2005/Atom`. It seems to me that the processing rules depend on the "node type" not the "payload type" in this instance. (And yes, "type" is one of the most overloaded terms in computing!)

Ppjet6:
On 2020/09/30, Peter Saint-Andre wrote:
> Hmm. We might need to clarify what pubsub#type means, then. I agree that the processing rules for data coming from `urn:xmpp:microblog:0` nodes might be special, but the payload format is indeed `http://www.w3.org/2005/Atom`. It seems to me that the processing rules depend on the "node type" not the "payload type" in this instance. (And yes, "type" is one of the most overloaded terms in computing!)

Thinking about it, (and that's how I justified this to myself originally), I'm
enclined to say `http://www.w3.org/2005/Atom` is not the payload type here but
`urn:xmpp:microblog:0` is (or whatever NS used to designate microblog).

I don't think "payload format" is defined to designate the serialization
format, (which would be XML and not Atom), and I'd rather have microblog say
nodes contain Atom items.

What would be the point of having Atom defined as a payload type exactly?

Tools wanting to read Atom in a somewhat generic fashion can still go through
nodes and see that there are Atom items in there (shaped respecting the
microblog rules, which the tool might know about or not).

-- 
Maxime “pep” Buquet

stpeter:
Well, pubsub#type is defined (to the extent it's defined) as "The type of node data, usually specified by the namespace of the payload (if any)". Here the payload is Atom, right? If we want to define processing rules for data coming from microblog nodes, we can do that in the spec without overloading or changing the meaning of the pubsub#type field. Or so it seems to me.

Ppjet6:
On 2020/09/30, Peter Saint-Andre wrote:
> Here the payload is Atom, right?

I'm saying I wouldn't define the payload as Atom, no.

From xsf@ yesterday:
> Say I publish my new spec also using pubsub + atom with different but very specific rules about how to handle Atom stuff that is also different from how the Atom spec defines things, I can, right? It's my spec [and I'm telling others about it]
> Now I don't want microblog clients coming on my stuff and doing non-compliant things (as per my spec)
> And I guess they also wouldn't want that from me

-- 
Maxime “pep” Buquet

horazont:
@Ppjet6 Apparently, there is disagreement about this. Please post a summary to the standards@ list for further discussion.

PR #988, opened 2020-09-24
XEP-0060: Add integer-or-max datatype to use with Data Forms Validation

Ppjet6:
I'd like to have council's opinion on how to handle this.

Seems like the datatype requires a prefix, where to define it? where do I put things so that they end up in https://xmpp.org/registrar/xdv-prefixes.html ? In 0060 as well?

Should this type be its own spec? And then referred to and given meaning in 0060?

[Rendered version](https://ppjet.bouah.net/xeps/xep-0060-integer-or-max.html)

### DIFF
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -50,6 +50,12 @@
   &stpeter;
   &ralphm;
   <revision>
+    <version>1.20.0</version>
+    <date>2020-09-24</date>
+    <initials>pep</initials>
+    <remark><p>Add integer-or-max datatype to use with Data Forms Validation.</p></remark>
+  </revision>
+  <revision>
     <version>1.19.1</version>
     <date>2021-03-04</date>
     <initials>mw</initials>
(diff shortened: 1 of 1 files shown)


Ppjet6:
Standards thread: https://mail.jabber.org/pipermail/standards/2020-September/037753.html

horazont:
I looked at this from a XML schema perspective, and I think there are formal issues here.

I think the bare minimum would be to re-write this as:

```xml
<?xml version="1.0"?>
<xs:schema xmlns:xs='http://www.w3.org/2001/XMLSchema' targetNamespace='urn:foo:bar'>
  <xs:simpleType name='integer-or-max'>
    <xs:union>
      <xs:simpleType>
        <xs:restriction base='xs:integer'/>
      </xs:simpleType>
      <xs:simpleType>
        <xs:restriction base='xs:string'>
          <xs:enumeration value='max'/>
        </xs:restriction>
      </xs:simpleType>
    </xs:union>
  </xs:simpleType>
</xs:schema>
```

(with a proper namespace of course).

The `xs:` is, at this point, mostly convention I think. However, the `targetNamespace` is required to import the Schema into another Schema -- which is the only way to cross-reference Schemas across documents. The `targetNamespace` needs to match the `namespace` on the `xs:import` directive and it must not be empty.

If you put the above in `/tmp/foo.xsd`, you could put this down as `bar.xsd`:

```xml
<?xml version="1.0"?>
<xs:schema xmlns:xs='http://www.w3.org/2001/XMLSchema'
           xmlns:foo='urn:foo:bar'>
  <xs:import namespace="urn:foo:bar" schemaLocation="file:///tmp/foo.xsd"/>
  <xs:element name='size' type='foo:integer-or-max'/>
</xs:schema>
```

And then you could validate this XML file (`foo.xml`):

```xml
<size>max</size>
```

using xmllint like this:

```console
$ xmllint --schema bar.xsd foo.xml
<?xml version="1.0"?>
<size>max</size>
foo.xml validates
```

horazont:
fixe.

horazont:
Merged as b7fe8d8ebcda066babe8e5555bb6f13e8fd1e271, but github doesn't know :)

PR #993, opened 2020-10-31
XEP-0438: update to match draft-ietf-kitten-password-storage-01

SamWhited:
Updates to match [draft-ietf-kitten-password-storage-01](https://datatracker.ietf.org/doc/draft-ietf-kitten-password-storage/).


xep-0438.xml
```diff
@@ -31,11 +32,17 @@
   <supersededby/>
   <shortname>passwords</shortname>
   &sam;
+  <revision>
+    <version>0.1.2</version>
```
horazont, line 36: This goes beyond an editorial update. Please bump the minor version number instead of just the patch level or omit the revision block altogether  to let the editors do it.

SamWhited: There is no protocol in this spec; the whole thing is editorial.

horazont: The thing is Informational, but that still allows for distinguishing between editorial updates and things touching normative language or being substantial changes.

SamWhited: I don't really want to deal with it; feel free to update it if it matters.

horazont: Rule of thumb: if you’d want it to be announced on the ML, bump the minor or higher level. Otherwise, bump patch only.

I’ll force-push to your branch to fix it.

PR #998, opened 2020-11-06
XEP-0420: Release version 0.4.0

melvo:
* Use 'envelope' and 'content' consistently by renaming elements
* Update namespace to 'urn:xmpp:sce:1'

@vanitasvitae Here is the outcome of our discussion.

### DIFF
--- a/xep-0420.xml
+++ b/xep-0420.xml
@@ -2,7 +2,6 @@
 <!DOCTYPE xep SYSTEM 'xep.dtd' [
   <!ENTITY content "&lt;content/&gt;">
   <!ENTITY envelope "&lt;envelope/&gt;">
-  <!ENTITY payload "&lt;payload/&gt;">
   <!ENTITY time "&lt;time/&gt;">
   <!ENTITY rpad "&lt;rpad/&gt;">
   <!ENTITY to "&lt;to/&gt;">
@@ -31,6 +30,15 @@
   <shortname>SCE</shortname>
   &paulschaub;
   <revision>
+    <version>0.4.0</version>
+    <date>2021-04-07</date>
+    <initials>melvo</initials>
+    <remark>
+      <p>Use 'envelope' and 'content' consistently by renaming elements</p>
+      <p>Update namespace to <tt>urn:xmpp:sce:1</tt></p>
+    </remark>
+  </revision>
+  <revision>
     <version>0.3.2</version>
     <date>2021-03-04</date>
     <initials>mw</initials>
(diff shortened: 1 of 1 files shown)


vanitasvitae:
LGTM, but please *do not* merge right away. I want to hold this off until 2 weeks from now (November 20th) to see if we can accumulate more changes under the namespace bump.

vanitasvitae:
@horazont and other editors:
I'm currently in a dilemma. I really like the change @melvo proposes here, as it makes the XEP easier to comprehend (especially as its harder to mix up the payload element of encryption XEPs with the then-called content element, but I'm not sure if this minor change deserves a version bump.
The change in this PR would also affect other XEPs (OMEMO for example). Those would need to be updated as well, resulting in additional namespace bumps.
As XEP-420 has not yet been implemented in too many places, one could consider merging the changes without bumping the namespace, however that would cause chaos as well.

What would you say would be the best thing to do in this situation? Namespace bump yes or no?
Writing this I think I came to the conclusion of going with a namespace and simply updating dependent XEPs later on, but you editors have the last word :)

Flowdalic:
> Those would need to be updated as well, resulting in additional namespace bumps.

The change in this PR would also affect other XEPs (OMEMO for example). They don't strictly have to be updated, they could refer to the old namespace. However if XEPs depending on this XEP, i.e. XEP-0420, ever underwent a namespace bump, then they should also bump the namespaces of all XEPs they depend on.

This PR introduces a breaking change of the wire protocol and hence requires a namespace bump.

vanitasvitae:
So in that case I'd say the change is good to go as is :+1:

vanitasvitae:
This PR is good to go :+1:

horazont:
>  The change in this PR would also affect other XEPs (OMEMO for example). Those would need to be updated as well, resulting in additional namespace bumps.

Make no mistake: The other specs (or their implementors) *are* affected. They have to deal with maintaining multiple versions of the elements which are used there.

vanitasvitae:
> Make no mistake: The other specs (or their implementors) are affected.
True, but there is a clearly defined way to deal with that :)

melvo:
As I understand the comments, the namespace should be changed. This pull request changes it from `urn:xmpp:sce:0` to `urn:xmpp:sce:1`.

@horazont Are there any other issues which have to be solved in order to merge this?

vanitasvitae:
@horazont is there anything we can do to increase the chance of this change getting merged?

vanitasvitae:
Ping @horazont

melvo:
Again: @horazont Are there any issues which have to be solved in order to merge this?

horazont:
@melvo Sorry, I somehow managed to miss all your past pings. There is a lot of noise in my inbox sometimes.

Right now, there is a conflict with the master branch which would have to be resolved. Relabelling so that I don’t miss it.

horazont:
@melvo Could you take care of the conflict?

melvo:
@horazont Thanks. The conflict is solved.

PR #1003, opened 2020-11-12
XEP-0434: Release version 0.2.0

melvo:
Improve explanations, descriptions and examples, introduce new attribute and complete all sections:

* Remove link to encryption protocol namespaces
* Shorten and improve introduction
* Use emphasizing text formatting instead of quotation marks
* Add new section for explaining the core properties of trust messages
* Add examples comparing trust messages to public key certificates
* Improve description of trust message structure
* Introduce 'usage' attribute for 'trust-message' element
* Focus on XEP-0420 and adjust examples accordingly
* Complete sections 'IANA Considerations', 'XMPP Registrar Considerations' and 'XML Schema'

### DIFF
--- a/xep-0434.xml
+++ b/xep-0434.xml
@@ -1,13 +1,16 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <!DOCTYPE xep SYSTEM 'xep.dtd' [
 	<!ENTITY % ents SYSTEM 'xep.ent'>
-	<!ENTITY encryption-protocols "<span class='ref'><link url='https://xmpp.org/extensions/xep-0380.html#table-1'>encryption protocols</link></span><note>Explicit Message Encryption - Encryption Protocols &lt;<link url='https://xmpp.org/extensions/xep-0380.html#table-1'>https://xmpp.org/extensions/xep-0380.html#table-1</link>&gt;.</note>">
+	<!ENTITY ns "urn:xmpp:trust-messages:0">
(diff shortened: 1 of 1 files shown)


xep-0434.xml
```diff
@@ -26,48 +29,57 @@
 	</dependencies>
 	<supersedes/>
 	<supersededby/>
-	<shortname>NOT_YET_ASSIGNED</shortname>
+	<shortname>None</shortname>
```
mdosch, line 32: Is there any specific reason to change this from `NOT_YET_ASSIGNED` to `None`?

melvo: I am not familiar with the process of assigning a short name. Therefore, I thought I should enter `None` since the XEP has no short name. Could you tell me how it should be done?

Flowdalic: NOT_YET_ASSINGED is more or less the canonical String XEPs carry if there is not yet a shortname assigned. You can enter your preferred shortname here, usually the editor will check if the shortname is available and, if so, simply accept it.

horazont: @melvo Yeah, please stick with `NOT_YET_ASSIGNED`. It is better to have a single "not assigned" value then a bunch of different ones. Easier to automate with that.

melvo: Thanks for your explanations! I added a short name. If that is already in use, I would change it to `NOT_YET_ASSIGNED`.

PR #1008, opened 2020-11-21
XEP-0047: fix incorrect max sequence number

SamWhited:
Previously the text said that the max sequence number was 65535 (max
unsigned 16 bit integer), but also said this was (1<<15)-1, which is
actually max signed 16 bit integer or 32767.

This spec is final and therefore will need council review but per [XEP-0001](https://xmpp.org/extensions/xep-0001.html#states-Final) can still be modified in small ways. I think this fits within the scope of "backwards compatible" modifications since it's effectively just a typo fix.


Flowdalic:
I'd like to note that the schema also declares 'seq' to be of type xs:unsignedShort, which is a 16-bit unsigned integer. All in all, this is just an editorial change.

PR #1014, opened 2020-11-27
XEP-0176: Improve compatibility with WebRTC clients

linkmauve:
Ping @iNPUTmice on this one.

An alternative would be to describe a way to make up 'id' and 'network' when working with WebRTC.

Foundation is actually described as “[a] sequence of up to 32 characters” [in RFC 8445](https://tools.ietf.org/html/rfc8445#section-5.3), but it is still described as a number elsewhere so not sure which one would be correct, maybe we’d like to take it up to the IETF?

### DIFF
--- a/xep-0176.xml
+++ b/xep-0176.xml
@@ -34,6 +34,17 @@
   &seanegan;
   &robmcqueen;
   <revision>
+    <version>1.1</version>
+    <date>2020-11-27</date>
+    <initials>egp</initials>
+    <remark>
+      <ul>
+        <li>Make the 'foundation' attribute a string instead of an unsignedByte, in accordance with &rfc8445;.</li>
+        <li>Make the 'network' attribute optional, and add a mapping to SDP.</li>
+      </ul>
+    </remark>
+  </revision>
+  <revision>
     <version>1.0</version>
     <date>2009-06-10</date>
     <initials>psa</initials>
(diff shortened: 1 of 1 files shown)


fippo:
foundation is a string, that is fixed in 0371.
network should be optional indeed (which needs to be fixed in 0371 too). One can still map it since the candidate syntax is extensible (https://tools.ietf.org/html/rfc5245#section-15.1 extension-att-name etc or https://tools.ietf.org/html/draft-ietf-mmusic-ice-sip-sdp-24#section-4.1)

Maybe @stpeter remembers where the long list of notes we had in 2017 is... unlikely though ;-)

linkmauve:
What would be the plan for migrating existing clients to XEP-0371 btw?

linkmauve:
Hi, I’ve fixed your comments, I’ll make another PR for making network optional in 0371 too.

fippo (approved):

linkmauve:
@fippo, speaking of which, what is “a character” in SDP?  Does it map to an Unicode character, in which case for UTF-8 “up to 32 characters” would mean “up to 128 bytes”?  I couldn’t find any mention of encoding in the RFCs.

fippo:
https://tools.ietf.org/html/rfc4566#section-5 and the BNF is the closest I've seen to a definition.
Technically unicode is allowed in some fields but some browsers [reject that](https://searchfox.org/mozilla-central/source/dom/media/PeerConnection.jsm#1040)

PR #1018, opened 2020-12-07
XEP-0372: Determine substring indexing convention

jcbrand:
The `begin` index is inclusive and the `end` index is exclusive.

This is the so-called Dijkstra convention that's used for subsequence indexing in many programming languages.
More info here: https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html

Standards list discussion: https://mail.jabber.org/pipermail/standards/2020-December/thread.html#37954

CC @Kev


Kev (approved):

Kev:
Thanks JC

PR #1020, opened 2020-12-09
XEP-0372: Use codepoints for references

Kev:


jcbrand:
Should this perhaps refer the user to https://xmpp.org/extensions/xep-0426.html?

Kev:
It could, but I'd rather not for the moment.

horazont:
Fun fact: Given that this used the term `character` without further context, it would’ve been clear reading the XML 1.0 standard that these have always been code points. Thanks.

horazont (approved):

PR #1024, opened 2020-12-30
XEP-0104: Fix examples by adding missing 'http:' prefix to child elements

vanitasvitae:


horazont:
I’m going to process this as editorial and adapt the revision block accordingly.

PR #1029, opened 2021-01-19
prettify: add link to source and license information

mathieui:
Source can be built from the referenced repo using "grunt uglify"


mathieui:
I forgot to add that apart from the obvious "we need to give attribution & licensing information", this pull request updates the library and fixes a long-standing issue of capitalized letters not highlighted in XML tags.

PR #1031, opened 2021-01-22
Update 'SOS' protoxep

mathieui:
- Address the issue of the tmp namespace
- Fixup abstract
- Restructure to use subsections
- Clarify usage of future date times
- Clarify the pubsub node a bit further
- Add some business rules & security considerations


horazont:
I’m porting this into the accepted version and github will not be able to see that -> closing this PR.

PR #1032, opened 2021-01-27
XEP-0004: Data Forms clarifications

Flowdalic:


mdosch:
To editors: The new revision is not bumping the minor but the patch version.

xep-0004.xml
```diff
@@ -317,7 +317,7 @@
...
+    <p>Each of these &lt;item/&gt; elements MUST contain one or more &lt;field/&gt; children. The &lt;reported/&gt; element defines the data format for the result items by specifying the fields to be expected for each item; for this reason, the &lt;field/&gt; elements SHOULD possess a 'type' attribute and 'label' attribute in addition to the 'var' attribute, and SHOULD NOT contain a &lt;value/&gt; element. Each &lt;item/&gt; element defines one item in the result set, and MUST contain each field specified in the &lt;reported/&gt; element (although the XML character data of the &lt;value/&gt; element MAY be null).</p>
```
ge0rg: I'm pretty sure that "Each of these elements" means both the `<reported>` and all `<item>` elements, so this correction actually changes the XEP's "MUST" semantics.

xep-0004.xml
```diff
@@ -295,6 +306,8 @@
       <li>One and only &lt;reported/&gt; element, which can be understood as a "table header" describing the data to follow.</li>
```
SamWhited, line 306: while you're touching this section and creating a diff anyways, maybe add the missing "one" here so it can go in the same version bump? (in case this gets accepted)

horazont:
Merged as 2cdc01a20f8b5555cc90396f4d0e5688c87d29a4, I messed something up which is why github doesn’t notice that.

PR #1036, opened 2021-02-03
Draft: Compliance Suite 2022

sonnyp:
It is early, but before I forgot about "Compliance Suite 2022: Make Connection Mechanism Discovery required for Web Core Client" which is the only way we can get decent interoperability on the Web.

Follow-up of the conversation surrounding https://github.com/xsf/xeps/pull/1033


inbox/cs-2022.xml
```diff
@@ -0,0 +1,582 @@
...
+          <td align='center'>&yes;&onlyone;</td>
+          <td align='center'>&yes;&component;</td>
+          <td align='center'>&yes;&onlyone;</td>
+          <td>&rfc7395;, &xep0206; (See also: &xep0124;)</td>
+        </tr>
+        <tr>
+          <td><strong>Connection Mechanism Discovery</strong></td>
+          <td align='center'>&no;</td>
+          <td align='center'>&yes;</td>
+	  <td align='center'>N/A</td>
```
sonnyp, line 249: Regarding `Connection Mechanism Discovery` - why is it `N/A` for Advanced Server but `no` for Core Server? 

See https://xmpp.org/extensions/xep-0443.html#web

Should there also be a column "deployments" were Connection Mechanism Discovery is required?

horazont:
@sonnyp This is still marked as draft. What is the status? Does this need to be a PR or can it be (temporarily) closed to not clutter up the PR view any further? I don’t know of any mailing list discussion related to this.

sonnyp:
@horazont if I remember correctly I marked this as draft because of the following issue on CI:

> anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

Instead of closing I would like to move this forward. What would be the way forward?

PR #1038, opened 2021-02-11
XEP-0452: Fix reference to Stanza Forwarding

vanitasvitae:
The XEP was erroneously referencing to XEP-0279 (Server IP Check) instead of XEP-0297: Stanza Forwarding.

### DIFF
--- a/xep-0452.xml
+++ b/xep-0452.xml
@@ -31,6 +31,12 @@
   </author>
   &mwild;
   <revision>
+    <version>0.2.1</version>
+    <date>2021-02-12</date>
+    <initials>ps</initials>
+    <remark>Fix reference to XEP-0297: Stanza Forwarding</remark>
+  </revision>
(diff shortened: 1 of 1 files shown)


xep-0452.xml
```diff
@@ -30,6 +30,12 @@
     <jid>soul@blastersklan.com</jid>
   </author>
   &mwild;
+  <revision>
+    <version>0.2.1</version>
+    <date>2021-02-12</version>
```
horazont, line 35: ``mismatched tag: line 35, column 22``

horazont (changes requested):
Please fix the XML errors.

xep-0452.xml
vanitasvitae: Oh, those have slipped through.

horazont (approved):

PR #1041, opened 2021-02-25
Update mix namespace for XEP-0406 and XEP-0408

mathieui:


Flowdalic:
Is the motivation only homogenization, or are those values just wrong?

mathieui:
Homogenization, but I could argue that the examples are technically wrong since 0369 is linked every time, and the text says that the replies should be conform to 0369 (and they do not since the namespace bump).

PR #1043, opened 2021-03-04
Cross-document editorial adjustments for inclusive language

mwild1:
Further reading and some rationale for these changes can be found at:
https://tools.ietf.org/html/draft-knodel-terminology-04

All changes are intended to be editorial in nature, not break existing
wire protocols, and not alter the meanings of any text.

### DIFF
--- a/xep-0353.xml
+++ b/xep-0353.xml
@@ -24,6 +24,12 @@
   &fippo;
   &stpeter;
   <revision>
+    <version>0.3.1</version>
+    <date>2021-03-04</date>
+    <initials>mw</initials>
+    <remark><p>Cross-document editorial adjustments for inclusive language.</p></remark>
+  </revision>
+  <revision>
     <version>0.3</version>
     <date>2017-09-11</date>
     <initials>XEP Editor (jwi)</initials>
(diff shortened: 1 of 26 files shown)


xep-0420.xml
```diff
@@ -395,7 +401,7 @@
 </section1>
 
 <section1 topic='XMPP Registrar Considerations' anchor='registrar'>
-  <p>TODO: Maybe the Registrar should handle a blacklist of elements that are allowed as child elements of the &content; element?</p>
+  <p>TODO: Maybe the Registrar should handle a list of elements that are allowed as child elements of the &content; element?</p>
```
Kev, line 404: I think this flips the logic doesn't it?

xep-0353.xml
```diff
@@ -270,7 +276,7 @@
   <p>Thanks to Lance Stout for his feedback.</p>
 </section1>
 <section1 topic='Security Considerations' anchor='security'>
-  <p>Because exchanging messages with other entities is effectively is a presence leak, an XMPP client that implements the receiving side of this specification MUST disable sending of proceed messages by default and MUST enable the feature only as a result of explicit user confirmation. Such confirmation can be provided per request, by whitelisting requests received from Jingle initiators in the responder's contact list, or through some other suitable means as long as sending proceed messages does not occur by default.</p>
+  <p>Because exchanging messages with other entities is effectively is a presence leak, an XMPP client that implements the receiving side of this specification MUST disable sending of proceed messages by default and MUST enable the feature only as a result of explicit user confirmation. Such confirmation can be provided per request, by only allowing requests received from Jingle initiators in the responder's contact list, or through some other suitable means as long as sending proceed messages does not occur by default.</p>
```
Kev, line 279: I think this subtly changes the meaning - to me the whitelist text implied that it was "automatically allowing" rather than "only allowing".

xep-0420.xml
mwild1: It does indeed, thanks for the catch! Pushed a fix.

xep-0353.xml
mwild1: Good question. I had originally written simply "allowing", but actually got the opposite impression, and that "whitelist" was equivalent to "allow only" rather than "[may] allow".

On re-reading it you're probably right, it was fine. The sentence as a whole is pretty clear of its intention either way. I've switched "only" to "automatically".

horazont (approved):

PR #1044, opened 2021-03-05
xep-0294: add mapping for a=extmap-allow-mixed

fippo:
@JonathanLennox @legastero can you review please?

meta issues:
1/ how do we handle IETF specs being updated like the 5285->8285 change?
2/ we still have the general issue of not having an equivalent to the session part in jingle so I had to write some text about mapping session level attributes

This is somewhat relevant since Chrome 89 started doing extmap-allow-mixed (after a long grace period during which old chrome versions rejected SDP with this) and did so at a session level.


JonathanLennox:
RFC 8285 obsoletes 5285 - it's the new version of the relevant spec.  So rather than saying "RFC 5285 and RFC 8285", I'd just say "RFC 8285", possibly with a note like "(formerly RFC 5285)" in some appropriate place.

Only 8285 should be listed in the abstract and the dependencies.

fippo:
right but that opens the question what clients that only implement 5285 do and whether updating the reference requires a namespace bump. I'm happy to conclude "no bump requires" and add a note to the changelog

JonathanLennox:
8285 was designed to be backward-compatible with 5285 - its main purpose was to define a=extmap-allow-mixed, and endpoints that don't understand that parameter can ignore it, not send it, and continue to behave in 5285-compatible ways.  See section 11 of 8285.  So yeah, "no bump required" seems right to me, as long as nothing implementing XEP-0294 1.0 will choke on seeing a new element in the `urn:xmpp:jingle:apps:rtp:rtp-hdrext:0` namespace.

fippo:
Updated to just mention 8285. Thank you @JonathanLennox!

fippo:
(also needs schema update, this has a schema even)

PR #1046, opened 2021-03-17
XEP-0280: take over XEP maintenance

ge0rg:
As voted in Council today, https://logs.xmpp.org/council/2021-03-17#2021-03-17-8aae2c6dbeddfaf5


mdosch:
See also https://mail.jabber.org/pipermail/standards/2021-March/038234.html

PR #1047, opened 2021-03-31
Relegate message retraction to xep-0424 instead of using custom one.

woj-tek:
As per, rather small, thread `[XEP-0407] - Message retraction question` on Standards I changed MIX specification (XEP-0407) and replaced custom retraction mechanism with reference to XEP-0424

### DIFF
--- a/xep-0407.xml
+++ b/xep-0407.xml
@@ -30,6 +30,7 @@
     <spec>XEP-0313</spec>
     <spec>XEP-0369</spec>
     <spec>XEP-0372</spec>
+    <spec>XEP-0424</spec>
   </dependencies>
   <supersedes/>
   <supersededby/>
@@ -172,22 +173,8 @@
 
     <section1 topic="Retracting a Message" anchor="usecase-retract">
       <p>
-        A MIX channel MAY support message retraction, where the sender of a messages or an authorized administrator deletes a message.  A MIX channel MAY limit the time frame in which a message is allowed to be retracted, for example to prevent retraction of very old messages.   When a messages is retracted the original message MAY be replaced by a tombstone.  Message retraction is done by sending a special message that identifies the original message.   This mechanism allows the retraction to be distributed on the same path as the original message so that all participating servers and clients MAY honour the retraction.  The protocol to request retraction does this by adding to a message  a &lt;retract&gt; element qualified by the 'urn:xmpp:mix:misc:0' namespace.  A retract messages MUST NOT have a body.   The &lt;retract&gt; element MUST include an 'id' attribute that holds the MAM-ID of the original  message.  The message sender will need to look up the MAM-ID.   The MAM-ID is the convenient message identification for message recipients.  A message and its retraction shown in the following example.
+        A MIX channel MAY support message retraction, where the sender of a messages or an authorized administrator deletes a message.  A MIX channel MAY limit the time frame in which a message is allowed to be retracted, for example to prevent retraction of very old messages.   When a messages is retracted the original message MAY be replaced by a tombstone.  Message retraction is done by means defined in &xep0424;.
       </p>
-      <example caption="User Retracts a Message"><![CDATA[
-
-<message from='hag66@shakespeare.example/UUID-a1j/7533'
-         to='coven@mix.shakespeare.example'
-         id='abcde'>
-  <body> A Message I did not mean to send </body>
-</message>
-
-<message from='hag66@shakespeare.example/UUID-a1j/7533'
-         to='coven@mix.shakespeare.example'
-         id='92vax143g'>
-  <retract id='77E07BB0-55CF-4BD4-890E-3F7C0E686BBD' xmlns='urn:xmpp:mix:misc:0'/>
-</message>
-]]></example>
       <p>
         The MIX channel will allow a user to retract a message sent by the user if the 'Allow User Message Retraction' option is configured.    The MIX channel will allow an administrative user to retract any message if the user is in the group specified by the 'Administrator Message Retraction Rights' option.
       </p>
@@ -198,23 +185,8 @@
         Two approaches to message retraction can be used.  In the first approach, the retracted message is simply removed.   This is appropriate where retraction is provided as a user service and the user has rights to remove messages sent from the record.
       </p>
       <p>
-        The second approach is to leave a tombstone, which if taken MUST be done in the following manner.  It is recommended to use a tombstone, as simply deleting the message may cause confusion with MAM queries.   Use of a tombstone is appropriate where it is desired to leave a record of the message that was redacted.
-        With this approach, the original message &lt;body&gt; is removed and replaced with a tombstone using the &lt;retracted&gt; element qualified by the 'urn:xmpp:mix:misc:0' namespace that shows the JID of user performing the retraction and the time of the retraction.
+        The second approach is to leave a tombstone.  It is recommended to use a tombstone, as simply deleting the message may cause confusion with MAM queries.   Use of a tombstone is appropriate where it is desired to leave a record of the message that was redacted.
       </p>
-      <example caption="Retracted message tombstone in a MAM result"><![CDATA[
-<message id='aeb213' to='juliet@capulet.example/UUID-e3r/9264'>
-  <result xmlns='urn:xmpp:mam:2' queryid='f27' id='28482-98726-73623'>
-    <forwarded xmlns='urn:xmpp:forward:0'>
-      <delay xmlns='urn:xmpp:delay' stamp='2010-07-10T23:08:25Z'/>
-      <message xmlns='jabber:client' from="hag66@shakespeare.example"
-                 to="macbeth@shakespeare.example">
-        <retracted xmlns='urn:xmpp:mix:misc:0' by='hag66@shakespeare.example'
-                 time='2010-07-10T23:08:25Z'/>
-      </message>
-    </forwarded>
-  </result>
-</message>
-]]></example>
     </section1>
 
     <section1 topic='Telling another User about a Channel' anchor='usecase-user-tell'>


horazont:
Thread: https://mail.jabber.org/pipermail/standards/2021-March/038256.html
Contains author approval.

Kev:
Can I have a chance to review as a whole when I’m back at work next week before it’s merged/published please? This is slightly more involved than it appears at first glance and I’d rather think it through first rather than potentially reverting it later.

/K

horazont:
@Kev Yes, no problem. I will tag it as Needs Author. Please report back when you’re done.

Kev:
Sorry for the delay, thanks for your patience. I've not been able to think of a reason moving to 424 in principle wouldn't work, but I think there might be bits missing from this PR:
* Under the previous text it was hopefully obvious how to construct the ID (from the MAM service of the MIX - which it needs to be), while 424 usually works on the origin-id, which wouldn't be appropriate here.
* The old protocol stored a note of who had retracted the item in the retraction, which I think is missing in 424.
* I think with the removal of the examples (which arguably shouldn't be the only place anyway, I realise), addressing of the retraction was obvious, which might be less true with the examples gone.
* It's probably worth keeping an example in place even while refering to 424.
* If delegating to 424, it probably matters which version of 424 needs to be supported.

horazont:
@Kev Thanks for finding the time to provide feedback!

@woj-tek Can you please take a look at that feedback and let me know how you want to proceed?

woj-tek:
@horazont I'll try to address @Kev 's comments and refine this PR.

woj-tek:
I just pushed commit that should address @Kev 's concerns:

> Under the previous text it was hopefully obvious how to construct the ID (from the MAM service of the MIX - which it needs to be), while 424 usually works on the origin-id, which wouldn't be appropriate here.

Somewhat of a OT and rant, but `origin-id` is just awful. Having server generated `stanza-id` would fix a lot of issues…

I added explicit comment about using `stanza-id` here, though I'm not sure it doesn't violates/contradicts 424: "by referring to its Unique and Stable Stanza IDs (XEP-0359) [5] **origin ID**."

> The old protocol stored a note of who had retracted the item in the retraction, which I think is missing in 424.

Shouldn't all messages have from? So even if message is retracted it's meta (to/from) remains intact? (though, it's missing from 424's examples - candidate for another PR?)

> I think with the removal of the examples (which arguably shouldn't be the only place anyway, I realise), addressing of the retraction was obvious, which might be less true with the examples gone.
> It's probably worth keeping an example in place even while refering to 424.

Restored and adjusted examples.

> If delegating to 424, it probably matters which version of 424 needs to be supported.

Included particular features - is that ok?

horazont:
@woj-tek Thanks for adapting your work! I pinged @Kev, hopefully he’ll find the time for another round soon.

xep-0407.xml
```diff
@@ -198,19 +204,19 @@
...
 <message id='aeb213' to='juliet@capulet.example/UUID-e3r/9264'>
   <result xmlns='urn:xmpp:mam:2' queryid='f27' id='28482-98726-73623'>
     <forwarded xmlns='urn:xmpp:forward:0'>
       <delay xmlns='urn:xmpp:delay' stamp='2010-07-10T23:08:25Z'/>
       <message xmlns='jabber:client' from="hag66@shakespeare.example"
                  to="macbeth@shakespeare.example">
-        <retracted xmlns='urn:xmpp:mix:misc:0' by='hag66@shakespeare.example'
-                 time='2010-07-10T23:08:25Z'/>
-      </message>
+        <retracted xmlns='urn:xmpp:message-retract:0' stamp='2010-07-10T23:08:25Z'>
```
Kev, line 217: The id of the MAM result needs to be the id the mix channel originally assigned to the retracted stanza.

xep-0407.xml
```diff
@@ -198,19 +204,19 @@
...
+          <origin-id xmlns='urn:xmpp:sid:0' id='a568706e-867b-49f1-a31b-e428ba9cece1'/>
```
Kev, line 218: This is missing the note of who retracted the message (also a comment from my original round ;))

xep-0407.xml
```diff
@@ -198,19 +204,19 @@
...
       </p>
       <p>
-        The second approach is to leave a tombstone, which if taken MUST be done in the following manner.  It is recommended to use a tombstone, as simply deleting the message may cause confusion with MAM queries.   Use of a tombstone is appropriate where it is desired to leave a record of the message that was redacted.
-        With this approach, the original message &lt;body&gt; is removed and replaced with a tombstone using the &lt;retracted&gt; element qualified by the 'urn:xmpp:mix:misc:0' namespace that shows the JID of user performing the retraction and the time of the retraction.
+        The second approach is to leave a tombstone.  It is recommended to use a tombstone, as simply deleting the message may cause confusion with MAM queries.   Use of a tombstone is appropriate where it is desired to leave a record of the message that was redacted. The tombstones are also defined in &xep0424;. A service which supports tombstones MUST advertise the 'urn:xmpp:message-retract:0#tombstone' feature in its Service Discovery responses.
```
Kev, line 207: I still think we need to be explicit which version of 424 is required.

Kev (commented):
Sorry for the ludicrous delay in replying. I have no excuse.

>>     Under the previous text it was hopefully obvious how to construct the ID (from the MAM service of the MIX - which it needs to be), while 424 usually works on the origin-id, which wouldn't be appropriate here.
> 
> Somewhat of a OT and rant, but origin-id is just awful. Having server generated stanza-id would fix a lot of issues…

Yes, we *have* to use the MIX channel generated stanza-id here, not the origin-id (or any other stanza-id)

> I added explicit comment about using stanza-id here, though I'm not sure it doesn't violates/contradicts 424: "by referring to its Unique and Stable Stanza IDs (XEP-0359) [5] origin ID."

This switches to using stanza-id, but it's still using a stanza-id stamped by the original sender, not by the MIX channel, unless I'm badly misreading it.

>>     The old protocol stored a note of who had retracted the item in the retraction, which I think is missing in 424.
> 
> Shouldn't all messages have from? So even if message is retracted it's meta (to/from) remains intact? (though, it's missing from 424's examples - candidate for another PR?)

Whether the message has a from is different from storing who it was who issued the retraction, no?

>>     I think with the removal of the examples (which arguably shouldn't be the only place anyway, I realise), addressing of the retraction was obvious, which might be less true with the examples gone.
>     It's probably worth keeping an example in place even while refering to 424.
> 
> Restored and adjusted examples.

Thanks.
 
>>     If delegating to 424, it probably matters which version of 424 needs to be supported.
> 
> Included particular features - is that ok?

I don't think that's enough, is it? If someone implements MIX they need to know which version of 424 to read to implement it.

xep-0407.xml
```diff
@@ -172,20 +174,24 @@
...
          id='abcde'>
   <body> A Message I did not mean to send </body>
+  <stanza-id by='hag66@shakespeare.example' id='a568706e-867b-49f1-a31b-e428ba9cece1' xmlns='urn:xmpp:sid:0'/>
 </message>
 
 <message from='hag66@shakespeare.example/UUID-a1j/7533'
          to='coven@mix.shakespeare.example'
          id='92vax143g'>
-  <retract id='77E07BB0-55CF-4BD4-890E-3F7C0E686BBD' xmlns='urn:xmpp:mix:misc:0'/>
+  <apply-to id="a568706e-867b-49f1-a31b-e428ba9cece1" xmlns="urn:xmpp:fasten:0">
```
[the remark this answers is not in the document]
Flowdalic: I think this could be solved if apply-to would not only refer to the id, but also to the 'by' value of the stanza-id element. For example
```
<apply-to
  xmlns="urn:xmpp:fasten:0"
  id="a568706e-867b-49f1-a31b-e428ba9cece1"
  by='hag66@shakespeare.example'>
```

See also my 2019 mail to @standards: https://mail.jabber.org/pipermail/standards/2019-September/036435.html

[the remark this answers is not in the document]
Kev: I don't think stamping the origin of the id actually matters very much in this case (it very much does in others), because in this case only one id is going to be usable - the one from the MIX itself, so the stamper can be implicit.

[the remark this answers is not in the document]
Flowdalic: Maybe, but I wonder how many implementations will get the implicit value wrong and open a security hole due that. I am usually a fan of avoiding unnecessary bytes on the wire and implicit values. But this smells a little bit like the carbon 'from' security hole that many implementations had.

I think at least, it should be clearly spelled out that the message is fastened to a message containing a stanza-id element where the 'id' matches the 'id' value from apply-to, and the 'by' value is the bare JID of the MIX channel (right?).

woj-tek:
Hi, apologies for delay on my part - this got buried in the notifications.

I improved the example so it should be clearer now which ID should be used. On a sidenode - I think that xep-0369 should add dependency on xep-0359 as well.

> Whether the message has a from is different from storing who it was who issued the retraction, no?

Correct. I haven't add any information regarding who made the retraction as... it seems that xep-0424 doesn't have that bit? And it should be included in 0424 first?

> I don't think that's enough, is it? If someone implements MIX they need to know which version of 424 to read to implement it.

My previous change already included:
>>> MUST advertise it's support by advertising  'urn:xmpp:message-retract:0' feature

so it did include version? Or am I missing something?

xep-0407.xml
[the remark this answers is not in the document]
woj-tek: I'm with @Kev here - the example was just bad and used wrong ID. Adding "by" wouldn't change that here.

xep-0407.xml
woj-tek: Should be fixed by correcting first example.

xep-0407.xml
woj-tek: Should be fixed by correcting first example.

PR #1050, opened 2021-04-06
XEP-0434: Release version 0.4.0

melvo:
Add new section, use more precise sentences, apply consistent formatting:

* Add new section for Trust Message URIs
* Use 'that' instead of 'which' for restrictive clauses
* Apply consistent formatting for paragraphs and revision history

PR #1051, opened 2021-04-06
XEP-0450: Release version 0.2.0

melvo:
Add usage of Trust Message URIs, use more precise sentences, apply consistent formatting:

* Add usage of Trust Message URIs for initial authentications
* Use 'that' instead of 'which' for restrictive clauses
* Apply consistent formatting for paragraphs and revision history


horazont:
Same issue as #1050 AFAICT.

melvo:
@horazont I changed the date.

PR #1052, opened 2021-04-07
xep.xsl: produce BibLaTeX entry

Flowdalic:
I am tired of seeing all those badly or misreferenced XEPs out
there. This produces a high-quality BibLaTeX entry that follows the
recommendations of the biblatex Package and
draft-carpenter-rfc-citation-recs-01 § 5.2 [1].

It also uses proper 'date' attribution, from the first date in the XEP
history to the date of the latest revision entry.

1: https://tools.ietf.org/html/draft-carpenter-rfc-citation-recs-01#section-5.2

### DIFF
--- a/xep.xsl
+++ b/xep.xsl
@@ -481,6 +481,40 @@ content: "XEP-<xsl:value-of select='/xep/header/number'/>";
             <xsl:apply-templates select='/xep/header/revision'/>
           </ol>
         <p>END</p>
+        <h3 id='appendix-biblatex'>Appendix I: Bib(La)TeX Entry</h3>
+        <xsl:variable name="lowercase" select="'abcdefghijklmnopqrstuvwxyz'" />
+        <xsl:variable name="uppercase" select="'ABCDEFGHIJKLMNOPQRSTUVWXYZ'" />
+        <xsl:variable name="lowercased-surname-firstauthor">
+          <xsl:value-of select='translate(/xep/header/author[position()=1]/surname, $uppercase, $lowercase)'/>
+        </xsl:variable>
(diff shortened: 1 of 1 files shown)


xep.xsl
```diff
@@ -481,6 +481,40 @@ content: "XEP-<xsl:value-of select='/xep/header/number'/>";
...
+            </xsl:otherwise>
+          </xsl:choose>
+        </xsl:variable>
+        <pre>
+<!-- If we had XSL 2.0 we could use year-from-date() and lower-case() below. -->
+@report{<xsl:value-of select='$lowercased-surname-firstauthor'/><xsl:value-of select='$year-of-first-history-entry'/><xsl:value-of select='$bibtex-citekey-postfix'/>,
+  title = {<xsl:value-of select='/xep/header/title'/>},
+  version = {<xsl:value-of select='/xep/header/revision[position()=1]/version'/>},
+  type = {XEP},
+  number = <xsl:value-of select='/xep/header/number'/>
```
vanitasvitae, line 511: It may be worth considering to replace this with `number = {<xsl:value-of select='/xep/header/number'/>}`,
so that leading zeros are preserved (0027 vs 27).

Flowdalic: Stylistic issue. I personally see no need to preserve the leading zeros. But no strong opinion.

vanitasvitae: Yeah me neither.

horazont: Since neither of you has strong opinions, I in fact do :). The leading numbers are used *everywhere*, in all references, so they should also be present in the BibLaTeX entries. I’ll add this myself so you don’t have to become active.

PR #1059, opened 2021-05-04
WIP: XEP-0373: Public key and metadata nodes: Require 'open' access model

vanitasvitae:
With 'open' as access model, entities without presence subscription are still able to establish encrypted communication channels and stay up to date with key material.


xep-0373.xml
```diff
@@ -337,6 +345,11 @@
     public key using Base64. The id of the node MUST be "urn:xmpp:openpgp:0:public-keys:" followed
     by the fingerprint string of the OpenPGP public-key contained in the data node.</p>
 
+    <p>It is REQUIRED to set the access model of the public key data node to ‘open’ to give
```
Flowdalic, line 348: REQUIRED feels to strong here. There may be valid use cases where a different access model is sensible, e.g. a whitelist based one. I think I would go with with RECOMMENDED, maybe even like "In absence of a use-case specific access model, it is RECOMMENDED to use the 'open' access model, to give…".

xep-0373.xml
```diff
@@ -372,6 +395,9 @@
     OpenPGP key fingerprint to the metadata node, the publisher MUST ensure that the public key is available
     at the corresponding data node.</p>
 
+    <p>Just like with the public key data node, the access model of the metadata node is also REQUIRED to be
```
Flowdalic, line 398: As above.

Flowdalic (changes requested):
LGTM. And since you asked in the MUC: I don't think that this requires a namespace bump. Basically, if something didn't work prior this change, then it would potentially work if entities apply this change. But not the other way around. So no bump necessary.

vanitasvitae:
Updated to say RECOMMENDED instead of REQUIRED.

dwd:
LGTM2. I am in agreement with Flow that no namespace bump is needed here; there is no effect on interoperability without one (and a namespace bump would simply break interoperability, as that's what they're for).

Issue #1065, opened 2021-06-02
Typo in XEP-0124


stpeter:
Someone reported to me a typo in XEP-124 (BOSH). In section "7.2 Session Creation Response", the document says: 

> "'to' -- This attribute communicates the identity of the backend server to which the client is attempting to connect."          

This should be FROM, not "TO". In fact, the XML examples are correct. The text is wrong.

PR #1066, opened 2021-06-03
XEP-0124: Fix incorrect attribute name in text

mwild1:
Confirmed that Prosody, ejabberd and Openfire all already use the correct attribute name.

Fixes xsf/xeps#1065


horazont:
Approved by council.

PR #1070, opened 2021-06-23
XEP-0001: add Archived status

SamWhited:
It was thrown out that an "Archived" status might be useful for XEPs like "SOAP over XMPP" that are still the "correct" way to implement the technology, but also they aren't likely to get implementations or ever be useful to anyone again. This PR is to start more discussion on that.

### DIFF
--- a/xep-0001.xml
+++ b/xep-0001.xml
@@ -23,6 +23,12 @@
   &dcridland;
   &ralphm;
   <revision>
+  <revision>
+    <version>1.25.0</version>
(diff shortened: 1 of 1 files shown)


SamWhited:
Ping board.

SamWhited:
Rebased. Ping @xsf/board again.

ralphm:
This was discussed at today's Board meeting. There doesn't appear to be appetite for introducing this new status. It appears to be very close to `Deprecated`. Closing.

PR #1086, opened 2021-07-10
XEP-0383: add listing of JIDs

SamWhited:
Rewrite parts of 0383 to include listing JIDs and to improve the security considerations and registrar considerations sections.


horazont:
(bumped version number to minor because of non-editorial changes, moved back from Deferred to Experimental)

PR #1087, opened 2021-07-12
Expire old XEPs

SamWhited:
It appears that XEPs aren't expiring properly. I wrote a quick script to expire all the ones that are older than 12 months in experimental.


horazont:
Did you know about `tools/deferrals.py`? ;) I just constantly forget to run it.

horazont:
I’m gonna use deferrals.py because it can create commits for the individual XEPs, which I find desirable. Thanks for the reminder.

PR #1088, opened 2021-07-12
XEP-0369 fix date format

SamWhited:
Every other date (including the month in this revision) uses a two digit day. Fix this so that parsers can be more strict.


horazont:
Good catch!

PR #1089, opened 2021-07-12
Move expired last calls back to experimental

SamWhited:
These XEPs are still showing up as under last call even though in some cases it expired quite a while ago. I didn't know if last calls counted towards the deferral date or not so I went to experimental even though some of them could be deferred if it's only revisions. It makes sense to me that LC counted though.


horazont:
@SamWhited Thanks for bringing those up and just in time! I am going to do '280 and '313 in a separate commit on main right away (council votes from 2021-04-07). The other two have slipped through the cracks in holiday season and need to be voted on, so I’ll put them on the agenda for tomorrow.

SamWhited:
These appear to have all been moved forward.

PR #1090, opened 2021-07-14
Carbons stream feature

SamWhited:
We should consider just adding a stream feature and calling it a day with carbons. A more streamlined solution that gets rid of carbons all together can always happen if bind2 ever becomes a thing.


horazont:
@SamWhited Is there a list thread about this? If not, please create one as per the [current process](https://mail.jabber.org/pipermail/standards/2021-February/038185.html).

horazont:
Discussed in https://mail.jabber.org/pipermail/standards/2021-July/038425.html.

PR #1093, opened 2021-07-20
XEP-0283: port moved2.xml into this document

horazont:
As discussed with @mwild1 and Council.

@mwild1: Please check, though the diff between moved2.xml and this one is really minimal.


horazont:
merged via 08baaba5b2a3838eb5bb178d250f3644b6f4d60f, but github doesn't know.

PR #1099, opened 2021-08-23
XEP-0313: add implementation considerations

SamWhited:
Currently this only explains the difference between the before/after
parameters of RSM and the before-id/after-id parameters of mam#extended.

### DIFF
--- a/xep-0313.xml
+++ b/xep-0313.xml
@@ -30,6 +30,14 @@
   &mwild;
   &ksmith;
   <revision>
+    <version>0.7.4</version>
+    <date>2021-08-23</date>
+    <initials>ssw</initials>
+    <remark>
+      <p>Add implementation note about before/after and before-id/after-id</p>
+    </remark>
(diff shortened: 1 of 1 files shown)


xep-0313.xml
```diff
@@ -403,7 +411,11 @@
     <section3 topic='Limiting results by id' anchor='query-limit-id'>
       <p>If the client has already seen some messages, it may choose to restrict its query to
       before and/or after messages it already knows about. This may be done through the 'before-id'
-      and 'after-id' fields.</p>
+      and 'after-id' fields.
+        In some cases 'before-id' and 'after-id' are the same as using RSM's
+        'before' and 'after parameters. For more information, see the
```
wurstsalat3000, line 416: Nitpick: 'after'

SamWhited: Fixed, thanks!

mwild1 (approved):

PR #1100, opened 2021-08-24
XEP-0001: change "Draft" to "Stable"

SamWhited:
Lots of discussion around this recently. Let's just do it.


horazont:
voted on by board last thursday: https://logs.xmpp.org/xsf/2021-08-26#2021-08-26-ab5cf8729dcfe858

PR #1102, opened 2021-08-25
XEP-0372: Document how to reference non-body elements

jcbrand:
Documents how the 'anchor' attribute can be used to disambiguate which text-containing element is being referred to.

I brought this up in the standards list here: https://mail.jabber.org/pipermail/standards/2021-August/038504.html

@Kev Please let me know if you agree with this approach.

### DIFF
--- a/xep-0372.xml
+++ b/xep-0372.xml
@@ -26,6 +26,12 @@
   <shortname>Refs</shortname>
   &ksmithisode;
   <revision>
+    <version>0.6.0</version>
+    <date>2021-08-25</date>
+    <initials>jcb</initials>
+    <remark>Document how to reference non-body elements</remark>
+  </revision>
+  <revision>
     <version>0.5.0</version>
     <date>2020-12-09</date>
     <initials>kis</initials>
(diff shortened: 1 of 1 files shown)


horazont:
@Kev Please post your response to the mailing list thread instead of here to keep the technical discussions off github… I also replied there, @jcbrand.

xep-0372.xml
```diff
@@ -126,6 +123,48 @@
...
+  <message type='headline'
+          id='a8hef4-99hct232h0'
+          from='verona@channels.shakespeare.lit'>
+    <subject id="subject">Enter SAMPSON and GREGORY, of the house Capulet, armed with swords and bucklers</subject>
+    <body>Gregory, o' my word, we'll not carry coals</body>
+    <reference xmlns='urn:xmpp:reference:0'
+              begin='6'
+              end='13'
+              type='mention'
+              anchor='subject'
```
Kev, line 142: On the mailing list your example used fragment URIs so anchor='#subject' - was the change here deliberate?

PR #1114, opened 2021-10-04
XEP-0450: Release version 0.4.0

melvo:
Update to XEP-0434 version 0.6.0 and XEP-0384 version 0.8.0:

* Use Base64-encoded key identifiers in examples
* Update TM's namespace to `urn:xmpp:tm:1`
* Update OMEMO's namespace to `urn:xmpp:omemo:2`


melvo:
This should be merged after #1113.

PR #1121, opened 2021-10-27
XEP-0459: Incorporate feedback from last call

sonnyp:
Incorporate feedback from last call.

https://mail.jabber.org/pipermail/standards/2021-October/038628.html


sonnyp:
There is a mistake in the changelog, it's core client/server that were renamed to client/server, not advanced client/server.

PR #1126, opened 2021-11-18
XEP-0060: Release version 1.23.0

melvo:
Advertise support, specify handling of partial configuration forms and fix typo:

* Advertise support for publishing items
* Specify how submitted partial configuration forms must be processed
* Replace 'allow' with 'allows'

### DIFF
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -50,6 +50,19 @@
   &stpeter;
   &ralphm;
   <revision>
+    <version>1.23.0</version>
+    <date>2021-11-18</date>
+    <initials>melvo</initials>
+    <remark>
+      <p>Advertise support, specify handling of partial configuration forms and fix typo:</p>
+      <ul>
+        <li>Advertise support for publishing items</li>
+        <li>Specify how submitted partial configuration forms must be processed</li>
+        <li>Replace 'allow' with 'allows'</li>
+      </ul>
+    </remark>
+  </revision>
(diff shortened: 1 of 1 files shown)


xep-0060.xml
```diff
@@ -3718,7 +3731,7 @@ And by opposing end them?
       </section4>
     </section3>
     <section3 topic='Form Submission' anchor='owner-configure-submit'>
-      <p>After receiving the configuration form, the owner SHOULD submit a completed configuration form.</p>
+      <p>After receiving the configuration form, the owner SHOULD submit a completed configuration form. The submitted configuration form MAY contain a subset of possible configuration options. In that case, the service MUST only change the submitted configuration options.</p>
```
horazont, line 3734: This is a normative change.

horazont (changes requested):
This change as-is has been rejected by the XMPP council. The reasoning is that the normative change highlighted above does not belong in XEP-0060; it should, instead, be moved to XEP-0004 or possibly even better an add-on XEP to '0004, as '0004 is in Final state.

The changes except the one noted above are editorial and can be merged. I would like to keep this PR on the record for future archeologists so I would highly appreciate it if you could cherry-pick the other changes into a fresh PR.

I can also do that, though that'll take until next week until I get around to it.

horazont:
Closing, because this PR should not be changed anymore for archiving purposes.

PR #1129, opened 2021-12-01
XEP-0459: Replace deprecated XEP-0411 with XEP-0402

linkmauve:
This is the way forward, as we deprecated the previous conversion method.

This keeps XEP-0048 stored in XEP-0049 as the preferred bookmarks storage in the absence of the `bookmarks2#compat` feature.


horazont:
Passed council vote on 2021-12-08.

PR #1130, opened 2021-12-25
XEP-0363: Size in bytes

Ppjet6:


Ppjet6:
Pulled in this PR in #1130. Closing.

PR #1135, opened 2021-12-28
XEP-0363: Add a sentence about multiple same headers and case insensitivity

linkmauve:
Thanks pep. for finding these issues!


Ppjet6:
I took these changes in #1140. This can be closed.

PR #1140, opened 2022-01-04
XEP-0363: notes

Ppjet6:
- Size in bytes
- Headers MUST be included in the PUT request
- Clients SHOULD NOT interpret headers
- Servers may want to sign headers
- Multiple same headers and case insensitive headers allowed

In different commits, if that helps. I think these should be fairly straightforward and not breaking common usage. Just that they weren't explicited in the XEP. Shall I dare say this is a clarification PR.


xep-0363.xml
```diff
@@ -32,6 +32,20 @@
...
     <jid>daniel@gultsch.de</jid>
   </author>
+  <revision>
+    <version>1.0.1</version>
+    <date>2021-12-28</date>
+    <initials>egp, mb</initials>
+    <remark>
+      <ul>
+        <li>Filename size in bytes.</li>
+        <li>Headers MUST be included in the PUT request.</li>
```
horazont, line 42: This is not an editorial change -> needs a minor version bump (and will run by council).

horazont (changes requested):

PR #1141, opened 2022-01-04
XEP-0353: editorial changes for #1128

tmolitor-stud-tu:
@jcbrand This should be merged on top of my other big PR #1128 and contains the editorial changes of the xml schema.

PR #1142, opened 2022-01-04
XEP-0353: add disco (based upon PR #1141)

tmolitor-stud-tu:
@jcbrand This is based upon my second PR (#1141) and adds a new revision (0.4.1)

This adds the disco feature suggested by Daniel Gultsch on the standards@ list.

### DIFF
--- a/xep-0353.xml
+++ b/xep-0353.xml
@@ -100,6 +100,29 @@
     <li>Allow all devices of the involved users to display start and end timestamps of the call</li>
   </ul>
 </section1>
+<section1 topic='Discovering support' anchor='disco'>
+  <p>If a client implements Jingle Message Initiation, it MUST specify the 'urn:xmpp:jingle-message:1' feature in its service discovery information features as specified in &xep0030; and the Entity Capabilities profile specified in &xep0115;.</p>
+    <example caption='Romeo&apos;s client requests information about a chat partner&apos;s client'><![CDATA[
+<iq type='get'
+    from='romeo@montague.example/orchard'
+    to='juliet@capulet.example/phone'
+    id='disco#info-1'>
+  <query xmlns='http://jabber.org/protocol/disco#info'/>
+</iq>]]></example>
+    <example caption='Juliet&apos;s client advertises support for Jingle Message Initiation'><![CDATA[
+<iq type='result'
+    to='romeo@montague.example/orchard'
+    from='juliet@capulet.example/phone'
+    id='disco#info-1'>
+  <query xmlns='http://jabber.org/protocol/disco#info'>
+...
+    <feature var='urn:xmpp:jingle-message:1'/>
+...
+  </query>
+</iq>]]></example>
+  <p>Because of &xep0280;, &xep0313; and &xep0357; it is unknown whether some or all receiving devices support this extension. Nevertheless determining support can help to identify online clients not supporting this spec and allows a client to initiate a direct jingle session with such a client in addition to the Jingle Message Initiation.</p>
+  <p>A client supporting this spec MUST always send &lt;propose/&gt; elements etc. even if it can not determine if the receiver supports this spec.</p>
+</section1>
 <section1 topic='Use Cases' anchor='usecases'>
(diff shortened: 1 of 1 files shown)


tmolitor-stud-tu:
@fippo @stpeter Please review this

fippo:
sorry for the delay!

I don't think this is the way to go, the lack of a disco is intentional.
In terms of UX you would probably timeout the attempt to use M-I and then proceed with falling back to the established behavior of discovering clients support 0166/0167 but that does not belong into this XEP and shouldn't use the disco feature.

tmolitor-stud-tu:
The lack of disco already made (as far as I know) Conversations and others implement a disco even though it is not specified in the xep.
What intentions does not having a disco fulfill?

@iNPUTmice could you clarify the usecases of having a disco a bit? First hand implementers experience could help a lot here.

tmolitor-stud-tu:
As far as I know, both, Dino and Conversations implement a disco and I will implement one for Monal, too.

That makes this spec another spec that does not reflect reality, which is sad.

tmolitor-stud-tu:
@fippo Why is the lack of disco intentional? Please elaborate!

fippo:
The main use-cases of 0353 is sending a message to a bare JID in order for it to be forked to all available resources.
This means you'd disco against the bare jid which typically (in the case of an IM server) won't support Jingle.

If you discover JMI support what do you gain? You'd still send to the bare JID and can't use the information returned from the information discovered from any available resource (and you can not make decisions about devices which appear to be offline but would get a push notification)

mwild1:
Related issue: https://github.com/snikket-im/snikket-server/issues/86

Based on the number of "why can't I see a call button?" support questions I see almost daily from a range of clients, I am inclined to lean towards less discovery and more graceful failure handling when JMI responds with an error or not at all.

That said, with the advent of long-lived device identifiers (new SASL2/bind2), I also consider this a candidate for a feature we could eventually advertise on the account rather than on individual devices (i.e. when the server knows a JMI-capable client has an active push registration).

tmolitor-stud-tu:
@mwild1 Yes, I don't like Conversation's behavior here, too.

What I want to accomplish with Monal is: show a warning that the call might not reach the receiver if not at least one device can be found having that disco feature and don't show that warning at all, if at least one device can be found.
I'm not going to hide the call button like Conversations does.

@fippo For that reason I'd appreciate to have a disco. I'm even going one step further: I will add such disco to Monal even if not specified in the XEP.
But I'd like XEPs to reflect reality. And having at least Monal, Dino and Conversations publish such disco features makes me think that this should be added to the XEP as well.

**And yes, surely with a big fat warning not to prevent calls if the disco can't be found because of temporarily offline devices that could still be woken up via push etc..**

iNPUTmice:
>  I am inclined to lean towards less discovery and more graceful failure handling when JMI responds with an error or not at all.

On paper I agree with you. However JMI will usually not respond with an error. It doesn’t even respond with an acknowledgment that *some device has started ringing*. So if you have a perfectly good device online right now that supports direct call initialization but not JMI your timeout for going from JMI to *let’s do direct IQ* has to be fairly high, unnecessarily prolonging the time the device (that only supports direct init) starts ringing.

That’s what the disco feature in Siskin, Dino and Conversations is for.

In a world where JMI was a quasi requirement for jingle calls I would agree with you that always sending JMI is the right move. But this wasn’t the case in 2020 when Conversations implemented calls and with the namespace bump in JMI and competing XEPs that are dedicated for call initialization I’m not sure we will get there in the short term.


I think it's also somewhat ironic that Disco is a thing in XMPP and the one time when Conversations uses that to actually show or hide UI everyone hates it. In general I don’t think disco is the wrong move. We just need to figure out how to make disco work with offline devices (account disco?) instead of finding work arounds for each XEP that may want to rely on disco.

tmolitor-stud-tu:
I think this can be closed now.

Many clients (at least Siskin, Dino, Conversations and Monal) will do disco.
Either for timely fallback to direct call initialization or, like Monal, to at least show a note to the user that the JMI call might go unnoticed on the receivers end.

The XEP just does not reflect this reality, but that's fine by me.

PR #1155, opened 2022-01-30
XEP-0353: Jingle Message Initiation -> Call Invite Message

mar-v-in:
- Restrict to calls only, removing non-call Jingle functionality (as discussed in Council)
- Support multiple and non-Jingle session establishment methods (derived from [Call Invites ProtoXEP](https://xmpp.org/extensions/inbox/call-invites.html))
- Adjust title and namespace
- Define Jingle-independent set of conditions for `<reason>`
- Add disco (from #1142)
- Add remark about random IDs (from #1142)

Rendered: https://larma.de/xeps/xep-0353.html
Diff: https://larma.de/xeps/xep-0353-diff.html

### DIFF
--- a/xep-0353.xml
+++ b/xep-0353.xml
@@ -100,6 +100,29 @@
     <li>Allow all devices of the involved users to display start and end timestamps of the call</li>
   </ul>
 </section1>
+<section1 topic='Discovering support' anchor='disco'>
+  <p>If a client implements Jingle Message Initiation, it MUST specify the 'urn:xmpp:jingle-message:1' feature in its service discovery information features as specified in &xep0030; and the Entity Capabilities profile specified in &xep0115;.</p>
+    <example caption='Romeo&apos;s client requests information about a chat partner&apos;s client'><![CDATA[
+<iq type='get'
+    from='romeo@montague.example/orchard'
+    to='juliet@capulet.example/phone'
+    id='disco#info-1'>
+  <query xmlns='http://jabber.org/protocol/disco#info'/>
+</iq>]]></example>
+    <example caption='Juliet&apos;s client advertises support for Jingle Message Initiation'><![CDATA[
+<iq type='result'
+    to='romeo@montague.example/orchard'
+    from='juliet@capulet.example/phone'
+    id='disco#info-1'>
+  <query xmlns='http://jabber.org/protocol/disco#info'>
+...
+    <feature var='urn:xmpp:jingle-message:1'/>
+...
+  </query>
+</iq>]]></example>
+  <p>Because of &xep0280;, &xep0313; and &xep0357; it is unknown whether some or all receiving devices support this extension. Nevertheless determining support can help to identify online clients not supporting this spec and allows a client to initiate a direct jingle session with such a client in addition to the Jingle Message Initiation.</p>
+  <p>A client supporting this spec MUST always send &lt;propose/&gt; elements etc. even if it can not determine if the receiver supports this spec.</p>
+</section1>
 <section1 topic='Use Cases' anchor='usecases'>
   <p>All &MESSAGE; stanzas exchanged by this protocol MUST be of type="chat" and contain &xep0334; &lt;store/&gt; hints.</p>
   <section2 topic='Indicating Intent to Start a Session' anchor='intent'>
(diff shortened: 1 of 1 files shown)


mar-v-in:
@fippo @stpeter Can you review this, please?

xep-0353.xml
```diff
@@ -237,48 +431,123 @@
...
-    <p>The &lt;finish/&gt; element MUST contain a &lt;reason/&gt; element as defined in &xep0166; section 7.4. This SHOULD use a condition of &lt;success/&gt;, but implementations MAY use other conditions if deemed more appropriate (see <link url="#security">Security Considerations</link> below for details and rationale).</p>
-    <example caption="Initiator Sends Finish Message"><![CDATA[
+    <p>
+      This protocol in conjunction with &xep0280; and &xep0313; or appropriate
+      groupchat protocols (like &xep0045;) allows all devices of all involved
```
fippo, line 436: I don't see how this solves a problem in MUC where the full JIDs are known.

xep-0353.xml
```diff
@@ -237,48 +431,123 @@
...
+  <finish xmlns='urn:xmpp:call-message:1' id='a73sjjvkla37jfea'>
+    <reason><success/></reason>
   </finish>
   <store xmlns="urn:xmpp:hints"/>
 </message>
 ]]></example>
   </section2>
+  <section2 topic="Reason element" anchor="reason">
+    <p>
+      The structure of the &lt;reason&gt; element is as follows.
```
fippo, line 484: what value does this provide over the jingle reason?

xep-0353.xml
```diff
@@ -237,48 +431,123 @@
...
+      to-attribute is set to the lower session ID and a &lt;reason/&gt; element
+      carrying the condition &lt;tie-break/&gt;.
+    </p>
       <example caption="Tie break in propose state"><![CDATA[
 <!-- lower session ID -->
 <message from='romeo@montague.example/orchard'
          to='juliet@capulet.example'
          type='chat'>
-  <propose xmlns='urn:xmpp:jingle-message:1' id='a73sjjvkla37jfea'>
-    <description xmlns='urn:xmpp:jingle:apps:rtp:1' media='audio'/>
```
fippo, line 281: the intent of putting the media into was to allow a phone to reject a video call when receiving a push message. That use-case remains valid.

fippo (commented):
a couple of questions inline.

Overall I do not think bringing groupchat into this is a good idea.

xep-0353.xml
mar-v-in: The information of the media being used in the call is still available via the `audio` and `video` attributes on the `<propose>` (audio defaulting to true, video defaulting to false, as described in the section "Invite to a Call").

Also, the clients can still negotiate to not accept the video content and only accept the audio content when doing the Jingle connection establishment.

xep-0353.xml
mar-v-in: - Jingle is only one establishment method to be used with call invites
- Most of Jingle conditions cannot be applied to call invites.
- Jingle's reason element does not have a `<tie-break/>` condition.

xep-0353.xml
mar-v-in: I don't understand what you mean here. Synchronizing the call state via messages works independently of full JIDs being used in MUCs or not. MUCs do however help with synchronization by providing message history (although nowadays we typically use MAM there as well) and forwarding the messages to all other participating resources in the room (which include your own other devices).

xep-0353.xml
```diff
@@ -126,37 +249,87 @@
...
+    add an attribute 'passive' with value 'true' to the &lt;accept&gt; element.
+    As this typically implies degraded user experience, this should be avoided
+    when possible.
+  </p>
   <example caption="One of Responder's Resources Accepts the Call"><![CDATA[
 <message from='juliet@capulet.example/phone'
          to='romeo@montague.example'
          type='chat'>
-  <accept xmlns='urn:xmpp:jingle-message:1' id='a73sjjvkla37jfea'/>
+  <accept xmlns='urn:xmpp:call-message:1' id='a73sjjvkla37jfea'>
```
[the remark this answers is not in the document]
mar-v-in: The `id` of `<accept>`  is referring to the `id` of a `<propose>`, the `sid` of the `<jingle>` element in `<accept>` is indeed redundant, as it is the reflected element from the `<propose>` message.

mar-v-in:
@fippo I guess inviting a group of people to a call is a feature we want, no? Do you think this should always happen with a bunch of direct messages instead of a groupchat message?

Kev:
Seems to have conflicts.

mar-v-in:
Larger parts of the functionality I aimed to get added to 0353 through this PR (mostly related to group calls and invites to *external* calls) is now provided through 0482. To avoid further confusion, I'm closing this one. If anyone feels like they still want to get parts of this into 0353, you're always open to take it.

PR #1158, opened 2022-02-10
Remove _xmppconnect DNS method from XEP-0156 and add warnings

moparisthebest:
https://mail.jabber.org/pipermail/standards/2022-February/038759.html

### DIFF
--- a/xep-0156.xml
+++ b/xep-0156.xml
@@ -7,7 +7,7 @@
 <xep>
 <header>
   <title>Discovering Alternative XMPP Connection Methods</title>
-  <abstract>This document defines an XMPP Extension Protocol for discovering alternative methods of connecting to an XMPP server using two ways: (1) DNS TXT Resource Record format; and (2) Web Host Metadata Link format.</abstract>
+  <abstract>This document defines an XMPP Extension Protocol for discovering alternative methods of connecting to an XMPP server via Web Host Metadata Link format.</abstract>
   &LEGALNOTICE;
   <number>0156</number>
   <status>Draft</status>
@@ -25,6 +25,12 @@
   &hildjj;
   &stpeter;
   &lance;
+  <revision>
+    <version>1.4.0</version>
+    <date>2022-02-10</date>
+    <initials>tjb</initials>
+    <remark><p>Remove DNS _xmppconnect method due to security vulnerability.</p></remark>
+  </revision>
   <revision>
     <version>1.3.1</version>
     <date>2020-07-07</date>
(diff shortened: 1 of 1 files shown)


xep-0156.xml
```diff
@@ -154,12 +160,9 @@
   </revision>
 </header>
 <section1 topic='Introduction' anchor='intro'>
-  <p>Although &xmppcore; specifies the use of TCP as the method of connecting to an XMPP server, alternative connection methods exist, including the &xep0124; method (for which &xep0206; is the XMPP profile) and the websocket subprotocol specified in &rfc7395;. For some of these methods, it is necessary to discover further parameters before connecting, such as the HTTP URL of an alternative connection manager. Without ways to auto-discover alternative connection methods, the relevant information would need to be provided manually by a human user (which is cumbersome and error-prone) or hard-coded into XMPP software applications (which is brittle and not interoperable).</p>
-  <p>This document defines two ways to encapsulate information about alternative connection methods for auto-discovery:</p>
-  <ol>
-    <li>DNS TXT resource records</li>
-    <li>Link entries in a server's "host-meta" file</li>
-  </ol>
+    <p>The proper and secure way to look up websocket/bosh connections is defined in &rfc7395; Section 4, which is to grab the host-meta over HTTPS, clients/servers MUST only use this method.</p>
```
sonnyp, line 163: RFC 7395 doesn't define bosh lookups

What about we don't obsolete this XEP but remove `_xmppconnect TXT records` ?

sonnyp: Nevermind - happy to remove BOSH and XEP-0156 from XEP-0459: XMPP Compliance Suites 2022

https://mail.jabber.org/pipermail/standards/2022-February/038766.html

moparisthebest: I went ahead and just removed the DNS method, let me know how that looks

xep-0156.xml
```diff
@@ -212,8 +179,9 @@ _xmppconnect IN TXT "_xmpp-client-websocket=wss://web.example.com:443/ws"
   <section2 topic='Business Rules' anchor='httpbizrules'>
     <p>The following business rules apply:</p>
     <ol start='1'>
+      <li>host-meta files MUST be fetched only over HTTPS, and MUST ignore any connection URLs that are not secure, eg not 'https://' or 'wss://', this provides secure delegation, meaning you SHOULD send SNI matching the host of the URL from the connection URL and validate that the certificate is valid for that host *or* the the XMPP domain</li>
```
mar-v-in, line 182: Instead of having (negated) positive examples, I think it would be better to make explicit negative examples: "e.g. 'http://' or 'ws://'", otherwise uneducated users may decide they consider http without TLS secure enough.

nits:
- "and **clients** MUST ignore any..."
- "e.g." instead of "eg"
- Add full stop before "this provides secure delegation"
- "the the"
- Add full stop at end of item

moparisthebest: Thanks for the feedback, I've re-worded it and fixed the other issues you identified, but I really want to keep the "allow these" wording vs "do not allow these" as it encourages implementations to implement a list of allowed protocols which is more secure than a list of disallowed ones.

mar-v-in: Yes, this way is even better, given that as of now, https and wss are effectively the only schemes that should be supported, as bosh and websockets are the only alternative connection methods you can discover with the XEP.

Only other nit:
The markdownish `*or*` doesn't really fit here. We have `<em>or</em>` or `<strong>or</strong>` for this. :)

PR #1168, opened 2022-02-18
XEP-0115: mention the pre-image attacks in the XEP

SamWhited:
This XEP currently does not mention several [known vulnerabilities](https://mail.jabber.org/pipermail/security/2009-July/000812.html). The security considerations section even suggests that the algorithm protects against this kind of attack.

I am not sure what the best path forward here is, but if it had mentioned it I would have at least considered this when writing my initial implementation, even if it's still not visible enough. Other suggestions welcome, and I'd love to see the council discuss this again.

### DIFF
--- a/xep-0115.xml
+++ b/xep-0115.xml
@@ -602,7 +602,7 @@
       &ltwarning;
     </section2>
     <section2 topic='Caps Poisoning' anchor='security-poisoning'>
-      <p>Adherence to the method defined in the <link url='#ver'>Verification String</link> section of this document for both generation and processing of the 'ver' attribute helps to guard against poisoning of entity capabilities information by malicious or improperly implemented entities.</p>
+      <p>Adherence to the method defined in the <link url='#ver'>Verification String</link> section of this document for processing of the 'ver' attribute is known to be vulnerable to certain cache poisoning attacks that can not be fixed in a backwards compatible manner <note><link url="https://mail.jabber.org/pipermail/security/2009-July/000812.html">[Security] Trivial preimage attack against the entity capabilities protocol</link>.</note>.</p>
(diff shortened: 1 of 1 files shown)


xep-0115.xml
```diff
@@ -602,7 +602,7 @@
       &ltwarning;
     </section2>
     <section2 topic='Caps Poisoning' anchor='security-poisoning'>
-      <p>Adherence to the method defined in the <link url='#ver'>Verification String</link> section of this document for both generation and processing of the 'ver' attribute helps to guard against poisoning of entity capabilities information by malicious or improperly implemented entities.</p>
+      <p>Adherence to the method defined in the<link url='#ver'>Verification String</link> section of this document for processing of the 'ver' attribute is known to be vulnerable to certain cache poisoning attacks that can not be fixed in a backwards compatible manner <note><link url="https://mail.jabber.org/pipermail/security/2009-July/000812.html">[Security] Trivial preimage attack against the entity capabilities protocol</link>.</note>.</p>
```
linkmauve, line 605: You removed a space before “Verification String”.

SamWhited: Oops, fixed, thanks.

mar-v-in:
Rather than writing "things are bad", I'd prefer seeing us giving detailed explanations on (heuristic) mitigations for the known vulnerabilities, so that the original statement becomes mostly true again (maybe adjust it to note that care has to be taken about using heuristic mitigations).

Regarding mitigations, I did some writeup here: https://gist.github.com/mar-v-in/8ee8259d5e5a185ba8009479be336253

Of course the best solution is to migrate to caps 2.0, but that is something that takes more time than adding a few mitigations to the few clients that care.

SamWhited:
I generally agree with this, but I'm not sure that I want to spend a lot of time adding content to this old XEP (which is presumably why no one else added them), I figured that at least mentioning them and linking to the email was a quicker way to just get something done. I will consider sitting down and writing more if I can find the time today, but either way it might be good to go ahead and get this out there so that the security considerations at least match the note on security earlier in the text.

mar-v-in:
Not saying I'm against the change as proposed at this point 🙂

PR #1177, opened 2022-04-16
Ephemeral messages -- Take 2

Ppjet6:
Reopening the discussion about ephemeral messages.

Rendered version: https://bouah.net/specs/ephemeral-messages.html

Should I try to contact the original author to take over the name? I added a 0.0.2 version to keep track of the original document.

Technically this spec isn't complete and I'm hoping to get feedback by submitting it, even though I would have preferred to get it before.


horazont:
@Ppjet6 Even though I don't think there will be an issue, please confer with the original author that the re-submission is acceptable.

Ppjet6:
Email sent.

PR #1180, opened 2022-05-15
XEP-0308: Correct `disco#info` example JIDs

deuill:
This commit corrects the JIDs provided in the query/response examples for XEP-0308,
specifically: the request example did not provide a valid `to` attribute, and the response
example did not have matching JIDs for the `from` and `to` attributes.


deuill:
This is my first contribution to XEPs so I apologize if I've missed any steps -- not sure if this warrants a version bump, and what the process for that might be.

PR #1188, opened 2022-07-17
XEP-0428: Fallback Indication v0.2.0

mar-v-in:
This incorporates most of the ideas of the ["Compatibility Fallback" ProtoXEP](https://xmpp.org/extensions/inbox/compatibility-fallback.html) as discussed in the council chat room back then.

- Add 'for' attribute such that entities can discover what the fallback is for.
- Allow to specify that only one of `<subject/>` or `<body/>` is meant as a fallback.
- Allow to specify the part of respective text that is meant as fallback where applicable.
- Don't use encryption example, which should use XEP-0380 instead.

### DIFF
--- a/xep-0428.xml
+++ b/xep-0428.xml
@@ -85,7 +85,7 @@
   </section2>
   <section2 topic="Fallback Indicator">
     <p>The fallback indicator is an element <tt>&lt;fallback/></tt> qualified by the <tt>&ns;</tt> namespace. It has an attribute
-      <tt>for</tt> that indidcates the specification that the fallback is meant to replace. This is typically the primary namespace
+      <tt>for</tt> that indicates the specification that the fallback is meant to replace. This is typically the primary namespace
       of the respective specification, but may be specified otherwise. The <tt>&lt;fallback/></tt> element may have one or multiple
       <tt>&lt;body/></tt> or <tt>&lt;subject/></tt> child elements, that indicate the part of the message, that is a fallback. Both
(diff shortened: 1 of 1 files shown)


mar-v-in:
@dwd

wurstsalat3000:
@mar-v-in congratulations on taking co-authorship on XEP-0428! I'd love to see this move forward, especially in conjunction with XEP-0461: Message Replies 👍

singpolyma:
Is `end` intended to be inclusive or exclusive?  By the example provided it seems exclusive, but the normative text should indicate one way or the other IMHO.

mar-v-in:
@singpolyma I'll add that to XEP-0426 instead (as it should be a general convention and not specific to this XEP).

mar-v-in:
This can be merged.

xep-0428.xml
```diff
@@ -67,25 +84,41 @@
       of <tt>&ns;</tt>. Note that lack of support will result in the desired fallback behaviour.</p>
   </section2>
   <section2 topic="Fallback Indicator">
-    <p>The fallback indicator is an element <tt>&lt;fallback/></tt> qualified by the <tt>&ns;</tt> namespace. It has no
-      attributes, content, or child elements.</p>
+    <p>The fallback indicator is an element <tt>&lt;fallback/></tt> qualified by the <tt>&ns;</tt> namespace. It has an attribute
+      <tt>for</tt> that indidcates the specification that the fallback is meant to replace. This is typically the primary namespace
```
bodqhrohro, line 88: indidcates → indicates

PR #1190, opened 2022-07-17
XEP-0448: Encryption for stateless file sharing v0.2.0

mar-v-in:
- Replace the ProtoXEP reference with a reference to the published XEP.
- Add urn:xmpp:ciphers:aes-256-cbc-pkcs7:0 (same as used in XEP-0384).

### DIFF
--- a/xep-0448.xml
+++ b/xep-0448.xml
@@ -1,9 +1,5 @@
 <?xml version='1.0' encoding='UTF-8'?>
-<!--
-Note to editor: Remove xep-sfs entity declared below and change all references from &xep-sfs; to respective &xepXXXX; to refeence sfs when moving to experimental.
--->
 <!DOCTYPE xep SYSTEM 'xep.dtd' [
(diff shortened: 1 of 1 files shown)


xep-0448.xml
```diff
@@ -146,6 +151,12 @@ Note to editor: Remove xep-sfs entity declared below and change all references f
       <td>Key: 256, IV: 96</td>
       <td>GCM/NoPadding</td>
     </tr>
+    <tr>
+      <td>urn:xmpp:ciphers:aes-256-cbc-pkcs7:0</td>
```
lnjX, line 155: Question: Do we need versioning for this namespace (`:0`)? :)

mar-v-in: I sticked with it when copying from [XEP-0391](https://xmpp.org/extensions/xep-0391.html#ciphers). We probably want to factor this out of both XEPs before they go stable and then should revisit the choice of having a version (it might make sense if we wanted to change things like IV size, but that's not even possible for many ciphers) - it certainly doesn't hurt for now.

lnjX: Ah okay, I wasn't aware of that it's already used elsewhere :) Then this is probably totally fine :)

PR #1194, opened 2022-07-25
XEP-0004: clarify rules for multi-item forms

SamWhited:
After a discussion on the mailing list a few days ago no conclusion was reached other than "multi-item forms are confusing". I believe this was likely the original intent of the section, and it's likely the simplest solution for now so I've gone ahead and made changes to disallow elements other than "reported" and "items" in multi-item forms. Users who want to mix and match should write a new XEP defining a new field type such as "table" that can contain other fields.

### DIFF
--- a/xep-0004.xml
+++ b/xep-0004.xml
@@ -28,6 +28,12 @@
     &temas;
     &stpeter;
     <revision>
+      <version>2.13.1</version>
+      <date>2022-07-25</date>
+      <initials>ssw</initials>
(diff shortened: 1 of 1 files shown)


singpolyma:
While I would prefer if we specified the opposite of this, I do understand that many have thought this is what the XEP already said so having something made clear is certainly better than leaving it unspecified.

singpolyma:
Oh wait... the version that got merged implies that even `title` isn't allowed with reported/items?  How does that make sense?

PR #1196, opened 2022-08-05
xep.xsl: Show new status 'Stable' instead of 'Draft' in appendix

melvo:


melvo:
@horazont I created this PR to follow the current approach covering each case of `Draft`'s usage in the code. But we could have cleaner code if we simply replaced the status `Draft` in each XEP file by  `Stable` and used that word directly (as it was before this PR). Would that have any disadvantage?

horazont:
@melvo It would also require updating existing downstream tooling, which may be somewhat out of our control.

melvo:
@horazont I see. Thanks for clarification!

PR #1197, opened 2022-08-06
XEP-0272: Release version 0.1.2

melvo:
Fix typos / grammar / examples and use 'Muji' instead of 'muji' consistently

### DIFF
--- a/xep-0272.xml
+++ b/xep-0272.xml
@@ -38,6 +38,12 @@
     <jid>dafydd.harries@collabora.co.uk</jid>
   </author>
   <revision>
+    <version>0.1.2</version>
+    <date>2022-08-06</date>
+    <initials>melvo</initials>
+    <remark>Fix typos / grammar / examples and use 'Muji' instead of 'muji' consistently</remark>
+  </revision>
+  <revision>
     <version>0.1.1</version>
     <date>2018-11-03</date>
     <initials>pep</initials>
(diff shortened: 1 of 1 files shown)


xep-0272.xml
```diff
@@ -24,7 +24,7 @@
   </dependencies>
   <supersedes/>
   <supersededby/>
-  <shortname>muji</shortname>
+  <shortname>Muji</shortname>
```
mar-v-in, line 27: Don't we typically use lowercase shortnames?

melvo: I saw that at on https://xmpp.org/extensions/xep-0384.html#appendix-docinfo and https://xmpp.org/extensions/xep-0369.html#appendix-docinfo. But https://xmpp.org/extensions/xep-0313.html#appendix-docinfo has indeed lowercase letters. @horazont Is that documented somewhere?

horazont: We indeed typically use lowercase names, I think, but that's not documented anywhere. Could you please fix that?

melvo: OK, thanks. I thought that the short name would simply be the name that is used instead of the complete name *Multiparty Jingle*. That would be *Muji* (and not *muji*) as the title says. But if that is not the case, in which context is the short name used then?

mar-v-in: I just checked through the list of all shortnames that we have and all of them use either uppercase or lowercase letters, combined with numerics and/or hyphens, except for two:
- XEP-0106 has shortname `jid\20escaping` - I'd consider this exception acceptable ;)
- XEP-0372 has shortname `Refs` - even worse than what we're discussing here IMO

That said, I don't think there is any real use of the shortname so it also doesn't really matter. I'd rather like to have a "XEP URN", which should be the same as the "primary" XML namespace used in the XEP. Often enough that namespace already is "urn:xmpp:$shortname:$version". But that's a whole different topic.

PR #1200, opened 2022-08-21
XEP-0001: Release version 1.25.0

melvo:
Specify XEP versioning

### DIFF
--- a/xep-0001.xml
+++ b/xep-0001.xml
@@ -23,6 +23,12 @@
   &dcridland;
   &ralphm;
   <revision>
+    <version>1.25.0</version>
+    <date>2022-08-21</date>
+    <initials>melvo</initials>
+    <remark><p>Specify XEP versioning</p></remark>
+  </revision>
+  <revision>
     <version>1.24.0</version>
     <date>2021-08-24</date>
     <initials>ssw</initials>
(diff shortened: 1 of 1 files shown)


melvo:
@wurstsalat3000 @Echolon @Zash @Kev Here is the outcome of our discussion.

horazont:
Is there a mailing list discussion on this? Because I have words to say on that and I'd like to put them in our standard venue instead of on github.

If there is a discussion, please link it here.

Kev:
> Is there a mailing list discussion on this? Because I have words to say on that and I'd like to put them in our standard venue instead of on github.

No. There was a question in one of the MUCs in which I said (roughly) that major was for advancement, patch was for Editorial-only changes, and minor was for other changes. Which I believe to be correct, but I don't believe it needs this change to confirm that (the major bit is already documented elsewhere, and IIRC when we introduced patches for editorial changes a few years ago we decided that this was implicit because of the definition of maj.min).

melvo:
@horazont The reason for specifying that topic was that @wurstsalat3000 wanted to create a list of XEPs a client, library or server implements and show a hint if the implemented version is not up-to-date. But such a hint should not be shown for different patch versions since they are only meant for editorial changes. The problem was that it was not clear and we could not find any document that specified it. I will sent a link to this PR to the corresponding mailing list.

horazont:
I'm not sure what to do with this, the [list discussion](https://mail.jabber.org/pipermail/standards/2022-August/038940.html) was inconclusive to me.

While I agree that it'd be great to have a versioning scheme written down, as I mentioned on-list, that's not going to be fixed for existing documents and as others noted, that's not the only interoperability criterium there is, so it might not even solve the issue at hand.

Please reopen if you think differently or drive the mailing list discussion forward to a clear consensus.

PR #1208, opened 2022-09-18
Add author entity for tmolitor

tmolitor-stud-tu:
Like the title says.


tmolitor-stud-tu:
Sorry, wrong branch :D

PR #1214, opened 2022-10-06
XEP-0388: Rework whole spec, namespace bump

tmolitor-stud-tu:
This is a collaboration between @mwild1 and me. Most parts have already been discussed with the original author @dwd, too.

In short, this adds:
- Namespace bump
- Make SCRAM hashes (and other SASL mechanism families) upgradable
- Add dependency on XEP-0440 for channel-binding agility
- Add new `<inline/>` element as extension point to inline stream resumption and bind2 (and possibly others)
- Clarify of `<continue/>` and tasks in general
- Update security considerations and business rules
- Add optional `<user-agent/>` element

### DIFF
--- a/xep-0388.xml
+++ b/xep-0388.xml
@@ -12,14 +12,34 @@
   <number>0388</number>
   <status>Deferred</status>
   <type>Standards Track</type>
+  <approver>Council</approver>
   <sig>Standards</sig>
   <dependencies>
     <spec>XMPP Core</spec>
+    <spec>rfc5802</spec>
+    <spec>XEP-0440</spec>
   </dependencies>
   <supersedes/>
   <supersededby/>
   <shortname>sasl2</shortname>
   &dcridland;
+  &tmolitor;
+  &mwild;
+  <revision>
+    <version>0.4.0</version>
+    <date>2022-11-16</date>
+    <initials>tm</initials>
+    <remark>
+      <ul>
+        <li>Bump namespace</li>
+        <li>Add reference to &xep0440; and &rfc5802;</li>
+        <li>Update security considerations and business rules</li>
+        <li>Clarify &lt;continue/> and tasks</li>
+        <li>Add expansion point to inline stream resumption and BIND2 (and possibly others)</li>
+        <li>Add optional &lt;user-agent/> element</li>
+      </ul>
+    </remark>
+  </revision>
   <revision>
     <version>0.3.0</version>
     <date>2018-10-01</date>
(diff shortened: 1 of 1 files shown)


horazont:
@dwd Ack please?

tmolitor-stud-tu:
For reference, here is the rendered version: https://dyn.eightysoft.de/final/xep-0388.html

xep-0388.xml
```diff
@@ -195,71 +300,187 @@
 </continue>
 ]]></example>
       <p>After the final octet of the first &lt;continue> element, any SASL security layer negotiated in the preceding exchange SHALL be immediately in effect.</p>
-      <p>Clients respond with a &lt;next/> element, which has a single mandatory attribute of "task", containing the selected task name, and contains an OPTIONAL base64 encoded initial response contained in an &lt;initial-response> element.</p>
+      <p>Clients respond with a &lt;next/> element, which has a single mandatory attribute of "task", containing the selected task name, and MAY contain any other element defined in a future protocol defining this task.</p>
+      <p>The concrete elements exchanged for each task after the &lt;next/> have to be defined in a future protocol defining a concrete task. Each task MUST end either by the server sending a &lt;failure/> element, if the task failed, a &lt;continue/> element, if the task was completed successfully and the server requests the client to perform a new task, or a &lt;success/> element, indicating that the task was completed successfully and no further tasks are needed.</p>
```
dwd, line 304: I designed tasks to be syntactically aligned with SASL mechanisms, partly in the hope that other protocols might adopt them, and partly because it allowed for an easily pluggable system on the endpoints (like SASL mechanisms themselves), and partly because it meant that there are only SASSL2 top level elements in flow during the entirety of the SASL2 exchange, making a few bits of housekeeping simpler.

The fact you've elected to switch to XML is in itself fine, but it does mean that the exchange now introduces arbitrary elements, and the SASL2 code itself needs to be more aware of what tasks may be in flight.

I'm not averse to using XML here for tasks; I don't honestly have much hope for (for example) LDAP or IMAP using them. But I do think that means you'll need other framing to handle the now-discarded properties.

xep-0388.xml
```diff
@@ -195,71 +300,187 @@
...
+  <p>Upon receiving the &lt;next/> element for the upgrade, the server provides the elements and data needed for the client to calculate the requested data. The concrete elements and exchanges needed for the upgrade are specific to individual tasks. This document currently defines tasks for upgrading credentials between different members of the SCRAM family of mechanisms. Other tasks may be documented in future versions of this document, or in other documents.</p>
+  <p>Upgrade task names SHOULD have a prefix of "UPGR-" followed by the SASL mechanism name to upgrade to, and if multiple mechanisms differ only in their support for channel-binding (e.g. SCRAM's -PLUS variants), implementations MUST use only the names of variants without channel-binding for the task names, because mechanism upgrades are independent of any channel-binding. Finally, upgrade tasks MUST NOT transmit plaintext passwords (or any reversible encoding of them) if the SASL mechanism to upgrade allows this to be avoided.</p>
```
dwd, line 328: Every time you impose a restriction on the string syntax of an opaque identifier, a layer-violation fairy dies.

Less prosaically, either a client knows how to upgrade to a mechanism or else it doesn't, and either way the "UPGR-" doesn't really help.

xep-0388.xml
```diff
@@ -195,71 +300,187 @@
...
+<data xmlns='urn:xmpp:bloop2:example'>
+  <data>U28sIG5leHQgRk9TREVNIC0gMjAxOCwgdGhhdCBpcy4uLg==</data>
+  <description>BLOOP2 is cool!</description>
+</data>
+
+<!-- The Client now responds -->
+<upgrade xmlns='urn:xmpp:bloop2:example'>
+  <response>Li4uSSdsbCBidXkgYSBiZWVyIGZvciB0aGUgZmlyc3QgcGVyc29uIHdoby4uLg==</response>
+  <finalize>42</finalize>
+</upgrade>
```
dwd, line 354: See above comments on switching from base64 to XML. Even a <task xmlns='urn:xmpp:sasl:2'/> wrapper around these would simplify a lot of things.

xep-0388.xml
```diff
@@ -195,71 +300,187 @@
...
+-->
+<parameters xmlns='urn:xmpp:sasl:2'>
+  <salt>A_SXCRXQ6sek8bf_Z</salt>
+  <iterations>4096</iterations>
+</parameters>
+
+<!-- The client responds with the base64 encoded SaltedPassword. -->
+<hash xmlns='urn:xmpp:sasl:2'>
+  BzOnw3Pc5H4bOLlPZ/8JAy6wnTpH05aH21KW2+Xfpaw=
+</hash>
```
dwd, line 392: See above - all worthy things, but baking a particular SASL mechanism into the SASL profile seems entirely counter to the principles of SASL.

xep-0388.xml
```diff
@@ -195,71 +300,187 @@
...
+</section1>
+
+<section1 topic='Security Considerations' anchor='security'>
+  <p>Relative to the SASL profile documented in RFC 6120, this introduces more data unprotected by any security layer negotiated by SASL itself.</p>
+  <p>While no actual exchanges are introduced that are unprotected, the nature of this exchange might allow for (for example) a resource binding extension to be introduced.</p>
+  <p>SASL security layers are sparingly used in the field, however, so this is thought to be a theoretical, rather than practical, concern.</p>
+  <p>As explained in § 13.8.3 of &rfc6120;, servers and clients SHOULD NOT support SASL PLAIN unless it is required by the authentication backend. Usage of PLAIN allows for trivial downgrade atacks by Man-In-The-Middle attackers circumventing any possible channel-binding or mutual authentication and even allowing the attacker to gain access to the cleartext password. If possible, PLAIN should be disabled by default on servers and clients and only be enabled when configured by the admin respective client user (if at all).</p>
```
dwd, line 413: I hate restating things from other specifications, but I accept this is a personal quibble. And in any case, you can defer to the RFC for the weakening of the SHOULD NOT:

As explained in ... servers and clients SHOULD NOT support SASL PLAIN.

xep-0388.xml
```diff
@@ -195,71 +300,187 @@
...
+  <p>This protocol allows the client to provide a hint of its identity prior to authentication. This value MUST NOT be trusted to contain the real identity of the connecting client. Varying pre-authentication responses (such as supported SASL mechanisms) based on this value may allow unauthorized parties to discover whether an account exists, and even guess at what software might be used with that account (e.g. by observing whether the client has upgraded to specific mechanisms). Minimizing differences between the default response for unregistered JIDs and registered JIDs, as well as rate-limiting (to prevent enumeration attempts), are possible countermeasures. If the server does not know the client-identity it is RECOMMENDED to randomize the list of provided SASL mechanisms to reduce these risks even further.</p>
```
dwd, line 414: No it doesn't provide or allow for that - the only hint is provided by facilities in RFC 6120 itself. "This protocol makes use of the hitn contained in the stream "from" attribute ..."?

xep-0388.xml
```diff
@@ -195,71 +300,187 @@
...
+  <p>SASL2 MUST only be used by Clients or offered by Servers after TLS negotiation, see § 5.2 and § 5.3.4 of &rfc6120;.</p>
```
dwd, line 415: I'd actually specify this as a mitigation to the first three points.

xep-0388.xml
```diff
@@ -195,71 +300,187 @@
...
+  <p>Clients MUST use channel-binding, if available, when requesting an upgrade to make sure no MITM can eavesdrop that hash and subsequently use it for authentication. Note that a client can always choose to not upgrade SASL mechanisms if it can not use channel-binding or the connection is otherwise deemed not secure enough.</p>
```
dwd, line 416: Again, we're somewhat into MTI territory .

xep-0388.xml
```diff
@@ -195,71 +300,187 @@
...
+  <p>Clients MUST NOT send, and servers MUST NOT process, authentication requests included in the TLS 0-RTT ("early data") extension. Such data is vulnerable to replay attacks, and without adequate protection this could allow an attacker to disrupt established sessions or cause other side-effects depending on the inline features negotiated. This rule may be overridden by later specifications, provided they define appropriate protocols to mitigate these issues. Examples of these attacks and mitigations are discussed in Section 8 of &rfc8446;.</p>
```
dwd, line 417: I would make this a SHOULD NOT. I think you list enough explanation and potential mitigations to cover it.

xep-0388.xml
```diff
@@ -85,105 +105,190 @@
...
     <example caption="An authentication request"><![CDATA[
-<authenticate xmlns='urn:xmpp:sasl:1' mechanism="BLURDLYBLOOP">
-  <initial-response>Tm9ib2R5IGV2ZXIgZGVjb2RlcyB0aGUgZXhhbXBsZXMu</initial-response>
+<authenticate xmlns='urn:xmpp:sasl:2' mechanism='SCRAM-SHA-1-PLUS' upgrade='UPGR-SCRAM-SHA-256'>
+  <!-- Base64 of: 'p=tls-exporter,,n=user,r=12C4CD5C-E38E-4A98-8F6D-15C38F51CCC6' -->
+  <initial-response>cD10bHMtZXhwb3J0ZXIsLG49dXNlcixyPTEyQzRDRDVDLUUzOEUtNEE5OC04RjZELTE1QzM4RjUxQ0NDNg==</initial-response>
+  <user-agent id='d4565fa7-4d72-4749-b3d3-740edbf87770'>
+    <software>AwesomeXMPP</software>
+    <device>Kiva's Phone</device>
+  </user-agent>
```
dwd, line 181: OK, I could be persuaded that <user-agent/> itself, with the id, is a good idea. I can see the utility in it as a useful baseline.

But the way it's written, <software/> and <device/> seem to be both mandatory, and I think that's definitely wrong.

dwd (commented):
There's a *lot* of changes to digest here, beyond even those we discussed (and I reserve my right to change my mind even when we did discuss!).

I don't think this is sensible to publish yet, and certainly not without extensible discussion on the list.

Flowdalic:
> For reference, here is the rendered version: https://dyn.eightysoft.de/final/xep-0388.html

It would probably help to have a diff from the current version. Especially for an upcoming list disccusion. I usually [create HTML diffs for XEPs](https://github.com/Flowdalic/xeps/blob/e31cf7c009f3a30558bda947d787e61e40ecfe39/xep-template/Makefile#L66-L70) using [htmldiff](https://github.com/cygri/htmldiff).

mwild1:
I generally agree with @dwd's points. The main thing I'll defend is having `<user-agent>` in here.

Honestly, I spent the past 5 years somehow thinking SASL2 *already* had a client identifier in it. I guess I made that up somewhere along the way, but I do think it makes sense. It's not the end of the world if it's in an separate XEP, but it's a solution to so many problems, and necessary for some SASL mechanisms (e.g. `HT-*`) because they don't provide any other key via which the server can determine which credentials it should be comparing against (maybe that's something to fix in those mechanisms, but there are other use cases too, including just debugging and diagnostic purposes).

As for XEP-0198 integration, the intent is not to define it in this spec, merely to say it's possible. It's covered by PR #1215 updating XEP-0198 itself. I prefer that PR over XEP-0397, as it reuses more of the existing protocol and fulfils more use cases without involving any extra stuff (e.g. token auth mechanisms).

Anyway, this should all be on the mailing list. If someone sparks a thread, I'll be happy to ^C^V everything there. If nobody does, I'll start one soon.

xep-0388.xml
iNPUTmice: Obligatory comment that they are other methods to secure a stream than TLS.

xep-0388.xml
```diff
@@ -15,11 +15,31 @@
...
+    <date>2022-10-06</date>
+    <initials>tm</initials>
+    <remark>
+      <ul>
+        <li>Bump namespace</li>
+        <li>Make SCRAM hashes (and other SASL mechanism families) upgradable</li>
+        <li>Add dependency on &xep0440; and &rfc5802;</li>
+        <li>Update security considerations and business rules</li>
+        <li>Clarify &lt;continue/> and tasks</li>
+        <li>Add expansion point to inline stream resumption and BIND2 (and possibly others)</li>
```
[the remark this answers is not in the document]
iNPUTmice: To be clear `<inline/>` is merely a place where SASL 2 extensions can announce themselves (instead of being direct stream features).
But I don’t have any strong feelings on whether stream feature or inline stream feature is the cleaner approach. One signals that it depends on SASL2 the other one would allow use to use Bind 2 over Sasl 3.

tmolitor-stud-tu:
> There's a _lot_ of changes to digest here, beyond even those we discussed (and I reserve my right to change my mind even when we did discuss!).

Sure! I did not intend to imply that our discussionwas something like a pre-approval. I just wanted to note that you are informed that I was working on this XEP. Of course you can change your mind any time.

> I don't think this is sensible to publish yet, and certainly not without extensible discussion on the list.

I think the way forward if to address some of your issues we don't need to discuss about and leave the rest for an open discussion on the list. Does this seem sensible to you?

_Quick warning: I'm really short at time currently. Several days may pass until I'm able to answer._

xep-0388.xml
```diff
@@ -15,11 +15,31 @@
...
+  &mwild;
+  <revision>
+    <version>0.4.0</version>
+    <date>2022-10-06</date>
+    <initials>tm</initials>
+    <remark>
+      <ul>
+        <li>Bump namespace</li>
+        <li>Make SCRAM hashes (and other SASL mechanism families) upgradable</li>
+        <li>Add dependency on &xep0440; and &rfc5802;</li>
```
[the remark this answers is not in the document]
tmolitor-stud-tu: "dependency" is meant as "refers to". I'll change that to say `Add reference to &xep0440; and &rfc5802;`

xep-0388.xml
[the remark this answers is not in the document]
tmolitor-stud-tu: SASL3 could well define an extension point, too. I don't think it's neccessary to remove the `<inline/>` just because of an hypothetical SASL3.

xep-0388.xml
```diff
@@ -15,11 +15,31 @@
...
+  &tmolitor;
+  &mwild;
+  <revision>
+    <version>0.4.0</version>
+    <date>2022-10-06</date>
+    <initials>tm</initials>
+    <remark>
+      <ul>
+        <li>Bump namespace</li>
+        <li>Make SCRAM hashes (and other SASL mechanism families) upgradable</li>
```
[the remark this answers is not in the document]
tmolitor-stud-tu: Well, yes and no. The upgrade tasks themselves probably still belong into SASL2. SCRAM updates on the other hand could be in an extra XEP as well.

I merely added them to SASL2 because it was easier and the SCRAM update part was relatively small compared to the rest of the SASL2 XEP.

xep-0388.xml
```diff
@@ -85,105 +105,190 @@
...
+    <p>The &lt;authentication/> element MAY contain a list of &lt;upgrade/> elements each containing a string value of a possible SASL mechanism the client can perform an upgrade task for prefixed with "UPGR-". For more information, see <link url="#upgrade">SASL mechanism upgrade tasks</link>.</p>
     <p>The feature so advertised, and its child content, SHOULD be stable for the given stream to and from attributes and encryption state, and therefore MAY be cached by clients for later connections.</p>
-    <p>The Service Name used by XMPP is unchanged from RFC 6120.</p>
+    <p>Note that SASL2 is impossible for clients to initiate without at least one mechanism being available, and therefore MUST NOT be offered if not at least one mechanism can be provided.</p>
+    <p>The Service Name used by XMPP is unchanged from &rfc6120;.</p>
+    <p>All servers and clients supporting channel-binding MUST implement &xep0440;.</p>
```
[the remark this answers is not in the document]
tmolitor-stud-tu: The "MUST implement channel-binding" elsewhere could well be a bug and left over from an earlier iteration of the XEP.

A separate XEP for all the MTI stuff is a good idea, but I think the "if channel-binding, XEP-0440 is a MUST" should still be in SASL2.

xep-0388.xml
```diff
@@ -85,105 +105,190 @@
...
+  <authorization-identifier>user@example.org</authorization-identifier>
+</success>
+]]></example>
+      <example caption="Successful authentication with BLURDYBLOOP"><![CDATA[
+<success xmlns='urn:xmpp:sasl:2'>
   <additional-data>
     ip/AeIOfZXKBV+fW2smE0GUB3I//nnrrLCYkt0Vj
   </additional-data>
-  <authorization-identifier>juliet@montague.example/Balcony/a987dsh9a87sdh</authorization-identifier>
+  <authorization-identifier>juliet@montague.example</authorization-identifier>
```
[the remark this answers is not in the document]
tmolitor-stud-tu: That's a bug.

xep-0388.xml
```diff
@@ -85,105 +105,190 @@
...
       <p>Any security layer negotiated SHALL take effect after the ">" octet of the closing tag (ie, immediately after "&lt;/success>"), if it has not already taken effect at a &lt;continue> - see <link url='#continue'>Continue</link> below.</p>
-      <p>The &lt;success> element is immediately followed by a &lt;features> element containing the applicable stream features of the newly authenticated stream. Note that no stream restart occurs.</p>
+      <p>The &lt;success> element is immediately followed by a &lt;features> element in the "http://etherx.jabber.org/streams" namespace containing the applicable stream features of the newly authenticated stream. Note that no stream restart occurs.</p>
     </section3>
-    <section3 topic="Failure">
+    <section3 topic="Failure" anchor="failure">
       <p>A &lt;failure/> element is used by the server to terminate the authentication attempt. It MAY contain application-specific error codes, and MAY contain a textual error. It MUST contain one of the SASL error codes from RFC 6120 Section 6.5.</p>
+      <p>It MAY contain an &lt;additional-data/> element, containing additional data from the SASL mechanism that has just failed.</p>
```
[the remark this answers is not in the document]
tmolitor-stud-tu: Oh, I didn't know this.
I thought all these different error conditions defined in SCRAM could be of real value.

I'll remove this sentence.

xep-0388.xml
```diff
@@ -85,105 +105,190 @@
...
   <aborted xmlns='urn:ietf:params:xml:ns:xmpp-sasl'/>
   <optional-application-specific xmlns='urn:something:else'/>
   <text>This is a terrible example.</text>
 </failure>
 ]]></example>
     </section3>
     <section3 topic="Continue" anchor="continue">
-      <p>A &lt;continue/> element is used to indicate that while the SASL exchange was successful, it is insufficient to allow authentication at this time.</p>
-      <p>This can be used to indicate that the Client needs to perform a Second Factor Authentication ("2FA"), or is required to change password.</p>
+      <p>A &lt;continue/> element is used to indicate that while the SASL exchange was successful, it is insufficient to allow authentication at this time. This can be used to indicate that the Client needs to perform a Second Factor Authentication ("2FA"), needs to perform an SCRAM upgrade task as requested by the client, or is required to change password.</p>
```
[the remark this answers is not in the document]
tmolitor-stud-tu: Hmm, I'll clarify this.

xep-0388.xml
tmolitor-stud-tu: I wanted to make sure these tasks could eventually be used as SASL mechanism without introducing name clashes. I used `UPGR-` instead of `UPGRADE-` to keep the name below the 20 chars limit of SASL.
(`UPGR-` and `-PLUS` have the same size, and `-PLUS` won't be used for upgrades.

But I don't have a strong opinion about that.

xep-0388.xml
tmolitor-stud-tu: Well, as I said above I thought defining the SCRAM upgrade task was small addition and didn't want to introduce yet another new XEP, but of course we could move that to a new XEP.

xep-0388.xml
tmolitor-stud-tu: I've changed that to defer to the RFC :)

xep-0388.xml
tmolitor-stud-tu: changed that, too, thanks!

xep-0388.xml
tmolitor-stud-tu: @iNPUTmice RFC 6120 contains the same wording for SASL1.

That's the reason I wanted to state the same for SASL2 as well.

@dwd why would mandatory TLS be a mitigation for these 3 points?

xep-0388.xml
tmolitor-stud-tu: ah sorry, I can change that to a SHOULD. That's the MTI for channel-binding you are referring above, right?

xep-0388.xml
tmolitor-stud-tu: hmmm, any thoughts @mwild1 ?

xep-0388.xml
tmolitor-stud-tu: see @mwild1's response on list

xep-0388.xml
```diff
@@ -15,11 +15,31 @@
...
+    <initials>tm</initials>
+    <remark>
+      <ul>
+        <li>Bump namespace</li>
+        <li>Make SCRAM hashes (and other SASL mechanism families) upgradable</li>
+        <li>Add dependency on &xep0440; and &rfc5802;</li>
+        <li>Update security considerations and business rules</li>
+        <li>Clarify &lt;continue/> and tasks</li>
+        <li>Add expansion point to inline stream resumption and BIND2 (and possibly others)</li>
+        <li>Add optional &lt;user-agent/> element</li>
```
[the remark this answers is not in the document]
tmolitor-stud-tu: @mwild1 that one is for you, I guess.

xep-0388.xml
tmolitor-stud-tu: You can mark this as resolved if it was the MTI you referred to above.

xep-0388.xml
mwild1: I've no strong feelings, and don't object to this change. I'm curious why you would relax a 'MUST NOT' that relates to a security concern, but sure... maybe there are closed trusted environments that want to do TLS 0-RTT without negotiation or replay protection. I would likewise hope that the explanation is enough to tell developers not to do this under any scenario where there might be a risk.

FWIW the TLS spec itself [says](https://www.rfc-editor.org/rfc/rfc8446#section-8):

> Therefore, in normal operation, clients will not know which, if any, of these mechanisms servers actually implement and hence MUST only send early data which they deem safe to be replayed.

xep-0388.xml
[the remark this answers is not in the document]
mwild1: My response to this is in [this comment](https://github.com/xsf/xeps/pull/1214#issuecomment-1279753150) and [this mailing list post](https://mail.jabber.org/pipermail/standards/2022-October/038998.html).

xep-0388.xml
tmolitor-stud-tu: I'm moving the SCRAM part of upgrade tasks into a new ProtoXEP.

xep-0388.xml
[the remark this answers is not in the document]
tmolitor-stud-tu: I've split out the SCRAM upgrade task specification into a new ProtoXEP.

xep-0388.xml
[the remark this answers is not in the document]
tmolitor-stud-tu: I think this is already discussed on-list. I'll mark that as resolved here for now.

tmolitor-stud-tu:
I marked everything addressed in my last commits as "solved" to redirect the focus to unresolved things.
Feel free to "unresolve" things if you see the need for additional changes and/or start a discussion about this on-list.

tmolitor-stud-tu:
lgtm

current rendered version: https://dyn.eightysoft.de/final/xep-0388.html
current diff: https://dyn.eightysoft.de/final/xep-0388-diff.html

@mwild1 lgtm, too?

xep-0388.xml
dwd: You're presumably implying that if TLS is negotiated, the risk due to the first three points is mitigated. I'd state that explicitly.

xep-0388.xml
dwd: It's more that if you're using a multi-trip mechanism like SCRAM, then 0-RTT is fine for sending the SCRAM initial response, and it's no more or less secure. An attacker can at best repeat the client nonce, but that's hardly an issue. The problem is more likely to be that channel binding is trickier.

tmolitor-stud-tu:
> It's more that if you're using a multi-trip mechanism like SCRAM, then 0-RTT is fine for sending the SCRAM initial response, and it's no more or less secure. An attacker can at best repeat the client nonce, but that's hardly an issue. The problem is more likely to be that channel binding is trickier.

We just did not want to go down this rabbit hole and rather leave the 0-RTT stuff to be weakened and defined in more detail by other/future XEPs:
It might work for SCRAM, but there might be other/future SASL mechanisms that would be more vulnerable if 0-RTT was used. Going the "default deny" route rather than the "default allow" seems like a sensible thing to do here.

tmolitor-stud-tu:
> You're presumably implying that if TLS is negotiated, the risk due to the first three points is mitigated. I'd state that explicitly.

I still don't understand what "3 points" you are referring to.

xep-0388.xml
```diff
@@ -94,98 +113,179 @@
 </section1>
 
 <section1 topic='Overview' anchor="overview">
-  <section2 topic="Discovering Support" anchor="feature">
-    <p>Servers capable of SASL2 offer a stream feature of &lt;mechanisms/>, qualified by the "urn:xmpp:sasl:1" namespace. This in turn contains one or more &lt;mechanism/> elements in the same namespace, and potentially other elements (for example, the &lt;hostname/> element defined within XEP-0233).</p>
-    <p>Note that SASL2 is impossible for clients to initiate without at least one mechanism being available, and therefore MUST NOT be offered.</p>
+  <section2 topic="SASL mechanism negotiation" anchor="feature">
+    <p>Clients capable of SASL2 MUST send the desired bare JID they want to authenticate for in the "from" attribute of the stream-header unless they don't know it (e.g. when using the GSS-API SASL mechanism etc.) according to section 4.7.1 of &rfc6120;. If the "from" attribute is not empty and the domain used in this attribute does not match the domain used in the "to" attribute the server SHOULD respond with an invalid-from error as defined in §4.9.3.9 of &rfc6120;. Providing the bare JID in the "from" attribute saves one round-trip.</p>
```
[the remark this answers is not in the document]
tmolitor-stud-tu: Adding a sentence that TLS SHOULD be in place when specifying the from-attribute is a good idea.

The from-attribute is a MUST, because SASL2 is designed to not work properly without it for most authentication methods: the server can not specify in it's stream features which methods it supports unless it knows which user wants to authenticate.

That's okay only, if the server only offers mechanisms like SASL EXTERNAL etc.

That means the `if the "from" attribute is missing then this may incur an additional round-trip` is an explanation why the protocol was designed the way it is rather than an explanation for the MUST.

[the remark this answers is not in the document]
tmolitor-stud-tu: Thinking about this again, I think this whole paragraph should be moved to the upgrade tasks ProtoXEP.
Without upgrade tasks (and the resulting dependence of the mechanism list on the user in the from attribute), giving the from attribute won't even save an additional round-trip: SASL2 behaves exactly like SASL1 here.

dwd:
> > It's more that if you're using a multi-trip mechanism like SCRAM, then 0-RTT is fine for sending the SCRAM initial response, and it's no more or less secure. An attacker can at best repeat the client nonce, but that's hardly an issue. The problem is more likely to be that channel binding is trickier.
> 
> We just did not want to go down this rabbit hole and rather leave the 0-RTT stuff to be weakened and defined in more detail by other/future XEPs: It might work for SCRAM, but there might be other/future SASL mechanisms that would be more vulnerable if 0-RTT was used. Going the "default deny" route rather than the "default allow" seems like a sensible thing to do here.

You're not doing "default deny", you're doing "deny".

dwd:
> > You're presumably implying that if TLS is negotiated, the risk due to the first three points is mitigated. I'd state that explicitly.
> 
> I still don't understand what "3 points" you are referring to.

The first three points made in this section, about additional exposure and so on.

xep-0388.xml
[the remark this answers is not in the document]
dwd: But... All you're doing is raising the SHOULD in RFC6120 §4.7.1 to MUST. If you restate everything else people will think any omissions are intentional.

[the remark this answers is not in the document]
tmolitor-stud-tu: Done

tmolitor-stud-tu:
> > > It's more that if you're using a multi-trip mechanism like SCRAM, then 0-RTT is fine for sending the SCRAM initial response, and it's no more or less secure. An attacker can at best repeat the client nonce, but that's hardly an issue. The problem is more likely to be that channel binding is trickier.
> > 
> > 
> > We just did not want to go down this rabbit hole and rather leave the 0-RTT stuff to be weakened and defined in more detail by other/future XEPs: It might work for SCRAM, but there might be other/future SASL mechanisms that would be more vulnerable if 0-RTT was used. Going the "default deny" route rather than the "default allow" seems like a sensible thing to do here.
> 
> You're not doing "default deny", you're doing "deny".

But that "deny" could be overwritten by a later xep, no?

xep-0388.xml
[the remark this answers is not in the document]
tmolitor-stud-tu: I've moved that MUST to the protoXEP and rephrased it. It should be clearer now: https://dyn.eightysoft.de/final/xep-scram-upgrade.html#protocol
Or do you think it's still not clear?

tmolitor-stud-tu:
> > > You're presumably implying that if TLS is negotiated, the risk due to the first three points is mitigated. I'd state that explicitly.
> > 
> > 
> > I still don't understand what "3 points" you are referring to.
> 
> The first three points made in this section, about additional exposure and so on.

Ah, I understand.
Fixed here: https://github.com/xsf/xeps/pull/1214/commits/18fb29343be4c8ada47bb23450d2c08ce2e46afe

xep-0388.xml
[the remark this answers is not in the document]
tmolitor-stud-tu: @horazont over here

dwd (approved):

Neustradamus:
To follow

tmolitor-stud-tu:
@Kev this isn't "needs author" anymore, Dave already confirmed this on standards@

@dwd can you confirm this here again?

tmolitor-stud-tu:
@dwd I'm sorry, I can't find your mail at standards@ confirming this can be merged. Am i hallucinating?

tmolitor-stud-tu:
@Kev This is not NeedsAuthor anymore, but ReadyToMerge!

See here: https://mail.jabber.org/pipermail/standards/2023-January/039133.html

guusdk:
@tmolitor-stud-tu I believe you can make the XEP validation happy if you reverse the order of lines 15 and 16. `<sig>` should go before `<approver>`, not the other way around.

tmolitor-stud-tu:
> Update: I was allowed to and have pushed a commit with this change to your branch.
Perfect, thanks!

guusdk:
XEP state is currently 'deferred'. Changes are _not_ purely editorial. As per https://github.com/xsf/xeps/blob/master/docs/TRIAGING.md this requires a move to 'experimental' state.

guusdk:
In an attempt to move this forward (and after consultation in editors@) I've added a commit that changes the status from the XEP from Deferred to Experimental.

guusdk:
> not doing what I asked

I blame poor instructions :)

Should be fixed now.

PR #1217, opened 2022-10-07
XEP-0386: New revision v0.4.0

mwild1:
This is an extensive update that goes along with the new [XEP-0388 (SASL2) updates](https://github.com/xsf/xeps/pull/1214).

Quoting from the revision change block:

- Various changes, made in parallel with working client and server implementation experience, and   
SASL2 updates.                                                                                     
- More tightly define the integration with XEP-0388 and several session feature XEPs: XEP-0198,     
XEP-0280, XEP-0352.
- Replace the custom latest-id element with the new metadata element from XEP-0313, which also      
provides richer information.
- Drop unread tracking, as this is a deep topic not directly related to resource binding. Instead   
the details of integration with other extensions have been better defined and demonstrated, to allow such  
functionality when it is fully defined and exists.
- Adjust proposed namespace on aesthetic grounds and consistency with SASL2's approach. As this     
protocol may become part of the new preferred connection flow for a long time to come, it makes no sense   
to include the redundant and potentially confusing '2' when there is no conflict without it. Similarly,    
the '.0' has been dropped from the XEP's title, as it isn't really a version number.
- Allow the client some influence over the resulting resource identifier. Encouraged the server to  
ensure the final identifier is unique and stable, following what is common practice on the network
today.

The new revision has been run by @Kev for review.

### DIFF
--- a/xep-0386.xml
+++ b/xep-0386.xml
@@ -6,7 +6,7 @@
 <?xml-stylesheet type='text/xsl' href='xep.xsl'?>
 <xep>
 <header>
-  <title>Bind 2.0</title>
+  <title>Bind 2</title>
   <abstract>This specification provides a single-request replacement for several activities an XMPP client needs to do at startup.</abstract>
   &LEGALNOTICE;
   <number>0386</number>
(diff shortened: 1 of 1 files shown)


mwild1:
Rendered and HTML diff: https://preview.xmpp.net/b2869c4f/xep-0386.html

Neustradamus:
To follow

xep-0386.xml
```diff
@@ -24,6 +24,21 @@
   <supersededby/>
```
lnjX, line 24: I think the SASL 2 / XEP-0388 dependency is still missing here.

PR #1220, opened 2022-10-11
Add license for bibliographic data

### LINKED ISSUE #1219: State license of bibliographic data
If we want our bibliographic data to be consumed by third parties, see, e.g., https://github.com/ietf-tools/bibxml-service/issues/302#issuecomment-1274686153. then it should have a license disclaimer.

Flowdalic:
We wheref asked [1] to state that the XEP bibliographic data is openly available and free of charge if we want it to be consumed by third parties.

The license text is only a suggestion. Feel free to improve, criticize or bikeshed. ;)

Fixes #1219.

1: https://github.com/ietf-tools/bibxml-service/issues/302#issuecomment-1274686153


stpeter:
We should either just point to the XSF's IPR policy https://xmpp.org/about/xsf/ipr-policy/ or create a LICENSE file that incorporates the licensing bits from that policy (i.e., modified MIT license).

Flowdalic:
I've considered our IPR, but as far as I can tell, it only applies to the XEPs itself and not to the bibliography data of the XEPs. Of course, the later is derived from the former, so an argument can be made that the IPR may apply to the bibliography data too. I have a minor tendency to just put the bibliography data under CC BY-ND, but not strong opinion. I wonder if this should be potentially decided by board (or council?, or the editor?)?

stpeter:
IANAL, but it seems most straightforward to use the same license for both bibliographic metadata and the underlying specifications. However, because the metadata is public information and does not include much IPR (only titles, authors, dates, etc. - not protocol descriptions, schemas, code snippets, etc.) personally I would be comfortable with CC0 or CC-BY. (I don't see why we would care to limit the ability to create derivatives). In any case, this is a business/legal decision and should be decided by the Board of the XSF.

Flowdalic:
> I don't see why we would care to limit the ability to create derivative works

That was just me following @ronaldtse's [suggestion](https://github.com/ietf-tools/bibxml-service/issues/302#issuecomment-1274738752):

> I would also suggest using a license that does not allow forking/content changes -- because this is authoritative bibliographic data.

ralphm:
I'm with Peter. Let's please use the same license for all the things.

ralphm:
Also, I don't think that this kind of data is copyrightable to begin with.

ronaldtse:
Hello everyone!

The bibliographic data is technically already licensed in the same way as the specifications they describe, because bibliographic data is basically an excerpt of the underlying specification.

Please note that the [XSF IPR policy](https://xmpp.org/about/xsf/ipr-policy/) states the following:

> Permission is hereby granted, free of charge, to any person obtaining a copy of this specification (the “Specification”), to make use of the Specification without restriction, including without limitation the rights to implement the Specification in a software program, deploy the Specification in a network service, and copy, modify, merge, publish, translate, distribute, sublicense, or sell copies of the Specification, and to permit persons to whom the Specification is furnished to do so, subject to the condition that the foregoing copyright notice and this permission notice shall be included in all copies or substantial portions of the Specification. Unless separate permission is granted, modified works that are redistributed shall not contain misleading information regarding the authors, title, number, or publisher of the Specification, and shall not claim endorsement of the modified works by the authors, any organization or project to which the authors belong, or the XMPP Standards Foundation.

In particular this sentence:
> Specification is furnished to do so, subject to the condition that the foregoing copyright notice and this permission notice shall be included in all copies or substantial portions of the Specification.

It depends on the XSF to define whether the bibliographic data constitutes "copies or substantial portions of the Specification". (possibly not, but it is not clearly defined). Someone may ask if "fair use" covers it, but jurisdictions outside the US have different interpretation of what fair use constitutes; it's better to be specific.

As @stpeter mentioned:
> However, because the metadata is public information and does not include much IPR (only titles, authors, dates, etc. - not protocol descriptions, schemas, code snippets, etc.) personally I would be comfortable with CC0 or CC-BY. 

I suppose it would be easier for people who redistribute or cite XEPs if the bibliographic data is distributed under one of these licenses. Thank you in advance!

Flowdalic:
Switched to CC0.

ronaldtse:
Thank you @Flowdalic !

ralphm (approved):

horazont:
I reverted this in #1221 because it broke the docker based build, see details there.

PR #1234, opened 2022-11-03
Shtml removal

Ppjet6:
I replaced a few shtml links with their corresponding page. I've split what I could in various commits as I wasn't sure what would be needed and/or if everything was going to pass as is. Feel free to squash parts or all of it.


xep-0019.xml
```diff
@@ -19,6 +19,12 @@
   <supersededby/>
   <shortname>N/A</shortname>
   &stpeter;
+  <revision>
+    <version>1.1</version>
```
horazont, line 23: Why did you create a revision block for this XEP only? And why did you use a minor level bump instead of a patch level bump?

xep-0019.xml
```diff
@@ -72,6 +78,6 @@
...
+  <p>There may be value in bringing back specialized SIGs in the future when the Jabber/XMPP community becomes larger. However, at this time I urge that we face the facts and proactively implement the solution I have outlined in this document. <note>Lest there be any concern that disbanding the SIGs is outside the power or purview of the XMPP Council, I note that Section 8.2 of the Bylaws of the XMPP Standards Foundation states in part that "The XMPP Council or the Members of the Corporation may, by resolution, ... terminate a Special Interest Group at any time for any reason." (An electronic copy of the Bylaws may be found at <link url="https://xmpp.org/about/xsf/bylaws/">http://www.jabber.org/bylaws.html</link>.)</note></p>
```
linkmauve, line 81: The link and its text are still not synchronised here, in addition the jabber.org text points to a 404.

Ppjet6:
It looks like a previous version of the PR had been merged, without the fix on versions.
4d763ee88506242da67554df41456bea985ebfd1
56c6de9da4eee1b57b112e5bdeaa0a830a033ac1

This fixes versions and also Link Mauve's suggestion.

Kev:
What is the right thing to do here?

PR #1242, opened 2022-11-05
fixes #1235: Script that validates XEP-0001 conformance

### LINKED ISSUE #1235: Create tool to validate standalone XEP-0001 conformance
## Summary

While a lot of things are already covered by `xep.dtd`, some more complex
requirements from XEP-0001 need to be checked manually currently. This is
error-prone and should be automated as far as feasible.

## Work Proposal

- Create a command-line utility which takes a single XEP filename as input (stdin won't work, see below)
- Validate the input file against the following criteria

    - Optional: Check DTD conformance against `xep.dtd`
    - If the filename matches `xep-[0-9]{4}.xml`: check that the number in the filename equals `/xep/header/number` (XPath)
    - If the filename does not match `xep-[0-9]{4}.xml`, check:

        - That the `/xep/header/status` (XPath) is ProtoXEP
        - That the `/xep/header/number` (XPath) is literally `XXXX`
        - That the name does not start with `xep-[0-9]`

    - Check that `/xep/header/status` (XPath) is a defined status (see `tools/xeplib.py` for an enum)
    - Check that `/xep/header/type` (XPath) is defined in XEP-0001

        - If the XEP number is less than 400, also accept some legacy values. To find which, see which you encounter in the XEP numbers below 400 :-).

    - Check that `/xep/header/approver` (XPath) is either Board or Council
    - Check that the version numbers in the revision blocks are descending (from top to bottom in the document)
    - If the approver (see above) is Board, enforce that `/xep/header/type` is not `Standards Track`.
    - Check that it uses the `xep.xsl` XML stylesheet.
    - Check that it includes the correct legal notice (either by checking for the entity reference, or by checking the content)

guusdk:
This adds a simple bash script that checks if a provided file (filename to be provided as first argument) conforms to XEP-0001.

Expected to be executed from the directory that holds all xep XML files. Requires one argument: the file name of the xep to be validated, eg:

    $ tools/validate-xep0001-conformance.sh xep-0010.xml

exit status will be non-zero upon validation failure.

requires: bash, xmllint

### DIFF
--- a/tools/validate-xep0001-conformance.sh
+++ b/tools/validate-xep0001-conformance.sh
@@ -31,6 +31,7 @@ header_number=$(xmllint --xpath '/xep/header/number/text()' --nowarning --dtdval
 header_status=$(xmllint --xpath '/xep/header/status/text()' --nowarning --dtdvalid "$xep_dtd" "$1")
 header_type=$(xmllint --xpath '/xep/header/type/text()' --nowarning --dtdvalid "$xep_dtd" "$1")
 header_approver=$(xmllint --xpath '/xep/header/approver/text()' --nowarning --dtdvalid "$xep_dtd" "$1")
+processing_instructions=$(xmllint --xpath '/processing-instruction("xml-stylesheet")' --nowarning --dtdvalid "$xep_dtd" "$1")
 
 if echo "$file_name" | grep -Eq "^xep-[0-9]{4}.xml$"
 then
@@ -42,7 +43,7 @@ then
     echo "[PASS] The number in the filename ('$file_name') equals XPATH value /xep/header/number/text() ('$header_number')."
   else
     echo "[FAIL] The number in the filename ('$file_name') does not equals XPATH value /xep/header/number/text() ('$header_number') (but should)."
-    validation_result=1;
+    validation_result=1
   fi
 else
   echo "[INFO] The filename ('$file_name') does not match 'xep-[0-9]{4}.xml'."
@@ -53,7 +54,7 @@ else
     echo "[PASS] XPATH value /xep/header/status/text() ('$header_status') equals 'ProtoXEP'."
   else
     echo "[FAIL] XPATH value /xep/header/status/text() ('$header_status') does not equal 'ProtoXEP' (but should)."
-    validation_result=1;
+    validation_result=1
   fi
 
   # 3.2 That the /xep/header/number/text() (XPath) is literally XXXX
@@ -62,14 +63,14 @@ else
     echo "[PASS] XPATH value /xep/header/number/text() ('$header_status') equals 'XXXX'."
   else
     echo "[FAIL] XPATH value /xep/header/nimber/text() ('$header_status') does not equal 'XXXX' (but should)."
-    validation_result=1;
+    validation_result=1
   fi
 
   # 3.3 That the name does not start with xep-[0-9]
   if echo "$file_name" | grep -Eq "^xep-[0-9].*$"
   then
     echo "[FAIL] The filename ('$file_name') starts with 'xep-[0-9]' (but should not)."
-    validation_result=1;
+    validation_result=1
   else
     echo "[PASS] The filename ('$file_name') does not start with 'xep-[0-9]'."
   fi
@@ -82,7 +83,7 @@ case $header_status in
     ;;
   *)
     echo "[FAIL] XPATH value /xep/header/status/text() ('$header_status') does not equals a defined status (but should)."
-    validation_result=1;
+    validation_result=1
     ;;
 esac
 
@@ -97,7 +98,7 @@ case $header_approver in
     ;;
   *)
     echo "[FAIL] XPATH value /xep/header/approver/text() ('$header_approver') does not equals 'Board' or 'Council' (but should)."
-    validation_result=1;
+    validation_result=1
     ;;
 esac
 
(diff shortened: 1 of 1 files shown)


guusdk:
I'm not much of a scripter. I went with the simplest tooling that I could think of: a basic Bash script with minimal dependencies, that should be easy to maintain.

This PR is in draft state, as not all validation checks as requested in the issue have been implemented.

I have performed cursory, but not full tests.

tools/validate-xep0001-conformance.sh
```diff
@@ -0,0 +1,132 @@
...
+# Expected to be executed from the directory that holds all xep XML files.
+# Requires one argument: the file name of the xep to be validated, eg:
+#
+# $ tools/validate-xep0001-conformance.sh xep-0010.xml
+#
+# exit status will be non-zero upon validation failure.
+#
+# requires: bash, xmllint
+
+set -e
```
horazont, line 15: Please change this to `set -euo pipefail` and fix any resulting issues. `-u` shouts at you if you use an undefined variable (very useful to catch typos) and `-o pipefail` improves error handling by also failing if any of the commands in a pipe fail.

tools/validate-xep0001-conformance.sh
```diff
@@ -0,0 +1,132 @@
...
+
+VALIDATION_RESULT=0
+
+# 1. Check DTD conformance against xep.dtd
+XEP_DTD="xep.dtd"
+if (xmllint --noout --dtdvalid "$XEP_DTD" "$1")
```
horazont, line 21: You do not need `()` around the command in an `if`. I think that causes a subshell to spawn, the effects of which I'm not certain about. Scripts are arcane.

tools/validate-xep0001-conformance.sh
```diff
@@ -0,0 +1,132 @@
...
+then
+  echo "[PASS] DTD conformance against xep.dtd"
+else
+  echo "[FAIL] DTD conformance against xep.dtd"
+  VALIDATION_RESULT=1;
+fi
+
+FILE_NAME="$(basename -- $1)"
```
horazont, line 29: Whenever you reference a variable (`$foobar`) and unless you know what you're doing, put it in double quotes.

But very good catch with `--` :-). Few people think of that (I often forget it, too).

tools/validate-xep0001-conformance.sh
```diff
@@ -0,0 +1,132 @@
...
+#
+# $ tools/validate-xep0001-conformance.sh xep-0010.xml
+#
+# exit status will be non-zero upon validation failure.
+#
+# requires: bash, xmllint
+
+set -e
+
+VALIDATION_RESULT=0
```
horazont, line 17: Matter of taste: I tend to use `UPPER_SNAKE_CASE` only for environment variables which are either passed into the script or exported to called programs later on. For script-local variables, I use `lower_snake_case`.

horazont (commented):
This is looking pretty great all over all :-).

I found a few nits. Some could be addressed for robustness, others are a matter of taste. Make of this what you like :-).

guusdk:
Added all but the validation for 'legal notice'. I both need to understand exactly what are valid definitions, and have more time than what I have right now to finish that.

I'm not particularly proud of the duplication in the implementation of the 'header type' validation, and am open to suggestions for improving things.

guusdk:
All validation checks (see footnote) have now been implemented. 

For testing, I have tried executing it against all XEP XML files in the root of this repository. I found that this test approach is a bit problematic, as the validation should be applied to _changes_ to XEPs. When applied to _preexisting_ XEPs, not all validation issues that are found can be fix.

One example of this is the validation that checks for the existence of an approver. Many XEPs already in the repository do not define one. Note that the stylesheet that is applied will use a default approver, when the XML does not define one: the rendered output will still have an approver.

With the approver-validation disabled, these files fail the validation script:

- xep-0015.xml - revision version numbers in wrong order
- xep-0017.xml - entity reference for the legal notice has not been detected
- xep-0024.xml - entity reference for the legal notice has not been detected
- xep-0037.xml - entity reference for the legal notice has not been detected
- xep-0041.xml - revision version numbers in wrong order (I suspect a copy/paste issue in the XEP)

I do not think that these failures necessarily need to be blocking issues for this PR to be accepted, which is why I have dropped the 'draft' status of this PR.

footnote: For the validation of the legalese text in the XEP, I have opted to _only_ check for the entity reference to be present, not for the entire text to be. Interesting challenges with markup pop up if you do that. In practise, existing XEPs all seem to use the entity reference anyway.

guusdk:
I wonder if the validation that checks for the Approver to be either Board or Council covers the case of a XEP [being Retracted by its author](https://xmpp.org/extensions/xep-0001.html#states-Retracted). The act of retracting does not seem to require approval of either Board or Council.

tools/validate-xep0001-conformance.sh
```diff
@@ -0,0 +1,205 @@
...
+  echo "[FAIL] DTD conformance against xep.dtd"
+  validation_result=1;
+fi
+
+file_name=$(basename -- "$1")
+
+# The test for exit code equalling 10 detects when the XPATH evaluation yields no results. In that case, the execution
+# should not fail immediately, but use an empty value instead (which will likely cause a validation failure further down).
+header_number=$(xmllint --xpath '/xep/header/number/text()' --nowarning --dtdvalid "$xep_dtd" "$1") || test $?=10
+header_status=$(xmllint --xpath '/xep/header/status/text()' --nowarning --dtdvalid "$xep_dtd" "$1") || test $?=10
```
horazont, line 34: It might be sufficient running the `--dtdvalid` check only once. No need to run the entire document through several rounds of DTD checking.

tools/validate-xep0001-conformance.sh
```diff
@@ -0,0 +1,205 @@
...
+      Historical | Humorous | Informational | Organizational | 'Standards Track' )
+        echo "[PASS] XPATH value /xep/header/type/text() ('$header_type') equals a defined type."
+        ;;
+      *)
+        echo "[FAIL] XPATH value /xep/header/type/text() ('$header_type') does not equals a defined type (but should)."
+        validation_result=1
+        ;;
+    esac
+    ;;
+esac
```
horazont, line 143: I don't quite get what this case is about. If you're special casing individual old documents (why?), why not check for a specific type?

When I wrote that requirement, I meant that it would probably be ok to degrade this check to a warning instead of a hard error for XEPs with numbers < 400.

tools/validate-xep0001-conformance.sh
guusdk: I agree that validation should be needed only once, but without it, I couldn't get xmllint to process the file. That's why I kept it in for the subsequent invocations.

tools/validate-xep0001-conformance.sh
guusdk: My reasoning: the script validates types against a set of valid values. For some specific documents, the set of valid values is a bit larger.

PR #1243, opened 2022-11-09
Pubsub Targeted Encryption: Fix wording and typos

vanitasvitae:
This PR adds some fixes for typos.
I'm not sure, what the process is for proposing patches for inbox'd XEPs, since those have not yet been accepted.
Let me know, if for example a new revision is needed.


vanitasvitae:
Ping @goffi-contrib for review :)

PR #1246, opened 2022-11-09
Use PEP (and XEP-0222) for XEP-0292

singpolyma:
While the existing version says this was rejected by community consensus, since then the consensus of most implementations is to use PEP directly even though it was never specified to work, and that the raw-IQ protocol adds complexity.  This update removes the raw-IQ mode and specifies the reuse of PEP.

### DIFF
--- a/xep-0292.xml
+++ b/xep-0292.xml
@@ -33,6 +33,12 @@
     <jid>samizzi@cisco.com</jid>
   </author>
   <revision>
+    <version>0.12.0</version>
+    <date>2023-03-14</date>
+    <initials>spw</initials>
+    <remark><p>Removes raw-IQ mode and specifies the reuse of PEP</p></remark>
+  </revision>
+  <revision>
     <version>0.11</version>
     <date>2020-06-08</date>
     <initials>ka</initials>
--- a/xep-0292.xml
+++ b/xep-0292.xml
@@ -155,16 +155,18 @@
 
   <p>This section describes the use of the vCard format for self-publication and retrieval of publicly-accessible information about any entity on an XMPP network, thus fulfilling all the use cases of the old vcard-temp format.</p>
 
-  <section2 topic='IQ-Based Publication and Retrieval' anchor='self-iq'>
-    <p>As in <cite>XEP-0054</cite>, the primary method for publishing and retrieving vCards is the XMPP &IQ; stanza. (Although it would have been possible to use &xep0222; for public storage and retrieval, community consensus is that storage via IQ is more backward-compatible with <cite>XEP-0054</cite>, and that publish-subscribe is more appropriate only for event notifications.)</p>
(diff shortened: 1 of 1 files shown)


xep-0292.xml
```diff
@@ -173,155 +175,160 @@
...
-      <p>Because Juliet has sent presence to Romeo including Entity Capabilities data that includes the "urn:xmpp:vcard4+notify" feature, Romeo's XMPP server will send a PEP notification to Juliet. The notification can include an XMPP message body for backward-compatibility with XMPP clients that are not pubsub-capable. This is in accordance with XEP-0060, second 6.1.7.</p>
+      <p>Because Juliet has sent presence to Romeo including Entity Capabilities data that includes the "urn:xmpp:vcard4+notify" feature, Romeo's XMPP server will send a PEP notification to Juliet.</p>
       <example caption="Receiving a vCard publication/update"><![CDATA[
   <message from='romeo@montague.lit'
            to='juliet@capulet.lit'
            type='headline'>
     <event xmlns='http://jabber.org/protocol/pubsub#event'>
       <items node='urn:xmpp:vcard4'>
-        <item id='current'/>
+        <item id='1e3fdd90-f542-41d3-82ae-33d183467a7a'/>
```
Zash, line 326: Why change this? [Asked in xsf@](https://logs.xmpp.org/xsf/2022-11-09?p=h#2022-11-09-4dffc3f70d5f479d)

singpolyma: I thought it was more clear for the ID here to match the ID in the publication example above, as this is an example notification about that publication, in effect.

horazont: But shouldn't the publication also use `id='current'` to avoid to accumulate ever more items?

singpolyma: Probably `pubsub#max_items` should be set to `1` for the node.

Zash: (tho you'd probably limit the node to 1 item)

Kev:
@stpeter You're probably the easier of the two Authors to get hold of to approve this :)

Kev:
This also needs a version block.

xep-0292.xml
stpeter: Did this conversation get resolved?

stpeter:
+1 LGTM

xep-0292.xml
Zash: Dunno, should it follow the [singleton](https://xmpp.org/extensions/xep-0060.html#impl-singleton) pattern or nah?

singpolyma: Given what the XEP already says

> Note: There is no payload, because this is a pure notification (the receiver needs to retrieve the vCard using an IQ-get as described earlier).

I think as it was written (using `current`) is basically not usable because you will always get a notification on connect that says "vcard is current" which... is just obviously true and tells you no information?

PR #1247, opened 2022-11-13
Upgrade XEP-xxxx: Pubsub Social Feed based on the feedbacks

edhelas:
- Rephrase the introduction
- Drop XHTML-IM section
- Formatting
- Change the default pubsub#type profiles


horazont:
@edhelas Pubsub Social Feed will be published as 0472 today, please apply your changes to that document once it's done, thanks.

Ppjet6:
Comments are also unspecified, but it seems there are bits remaining about them in https://xmpp.org/extensions/xep-0472.html#rich_content and https://xmpp.org/extensions/xep-0472.html#comment_author

PR #1250, opened 2022-12-01
Add MUC service shutdown example

truenicoco:
This is a minimal change discussed in jdev with @Zash. I think it is useful to have as many examples as possible.


horazont:
Merged as b4bacd8f415cacc633d750f59ea760d28854396d, sorry for the confusion.

PR #1251, opened 2022-12-09
XEP 0461 improvements.

truenicoco:
I think there is no reason that this XEP shouldn't have its associated disco feature.

I am rather confident that direct messages have no reason to refer to the quoted author with a full JID and that a bare JID makes more sense in this context.

I am less confident about permitting no `to` attribute in general and must confess this is something that would make it is easier to comply to the spec for my gateway implementation, but maybe this is not a valid reason. @edhelas since you implemented it in movim, do you use the 'to' attribute? Any opinion on allowing its absence?

### DIFF
--- a/xep-0461.xml
+++ b/xep-0461.xml
@@ -15,6 +15,7 @@
     <status>Experimental</status>
     <type>Standards Track</type>
     <sig>Standards</sig>
+    <approver>Council</approver>
     <dependencies/>
     <supersedes/>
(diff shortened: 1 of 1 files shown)


horazont:
@mar-v-in Any objections? If so, to the list please, otherwise just ack here.

mar-v-in (approved):

mar-v-in (approved):

PR #1252, opened 2022-12-15
Changes a XEP to test validation of XEPs

guusdk:


guusdk:
Eek. This never should've ended up here. Artifact of my tests of #1253

Issue #1240, opened 2022-11-05
Integrate new tooling into CI


horazont:
## Summary

In order to reduce the manual load on the editor, it is important that steps run automatically as far as reasonably possible.

## Work proposal

Create a CI workflow which:

- On pull requests, runs the tools created in #1235, #1237, #1238, #1239 and sends diagnostics in a comment on the GitHub PR, adding labels and tagging either the author or the editor as appropriate.

- On pushes to the main branch, uses #1236 to trigger the existing tooling for archiving and sending emails.

PR #1253, opened 2022-12-15
#1240 Integrate XEP validation in Continuous Integration

guusdk:
Adds a Github workflow that executes `tools/validate-xep0001-conformance.sh` against any changed XEP that was changed in a pull request.

### DIFF
--- a/.github/workflows/xep-validation.yml
+++ b/.github/workflows/xep-validation.yml
@@ -1,0 +1,35 @@
+name: XEP validation
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Validate any XEP changes
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Detect changes to XEP files
+        id: changed-xeps
+        uses: tj-actions/changed-files@v34
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/xep-validation.yml
```diff
@@ -0,0 +1,35 @@
+name: XEP validation
+
+on:
+  pull_request:
+    branches:
+      - master
+
```
wurstsalat3000, line 7: Might be helpful, so this action can be run on old PRs.

```suggestion
  # Allow to run this workflow manually from the Actions tab
  workflow_dispatch:
```

guusdk: I'm happy to accept that suggestion if you tested that the workflow remains functional for the original intended target: pull requests.

I am reasonably sure that the original worked for PRs, as that's what I tested. I don't want to spend the time to re-test this though. :) #lazy

wurstsalat3000: I tested `workflow_dispatch:` in [my fork](https://github.com/wurstsalat3000/) and it continues to work with pull requests. It additionally allows to manually run this workflow on any branch (but only branches from https://github.com/xsf/xeps/ , which isn't what I originally intended with my suggestion, since PRs are based on forks most of the time). Feel free to ignore :)

PR #1254, opened 2022-12-17
New ProtoXEP: Stream Limits Advertisement

Zash:
Negotiated stanza size limits allowing senders to abort early, leading to improved reliability.


inbox/xep-sla.md
```diff
@@ -0,0 +1,228 @@
...
+responding entity advertises its limits in the \<stream:features/\>
+element that it sends at the start of the stream.
+
+The limits are enclosed in a \<limits/\> element qualified by the
+\'urn:xmpp:stream-limits:0\' namespace. This element SHOULD contain the
+following child elements:
+
+\<max-bytes/\>
+:   Contains an integer representing the maximum size of any first-level
+    stream elements (including stanzas), in bytes. Guidance on
```
Flowdalic, line 86: While reading this, I was wondering in which direction this limit is enforced. Is it the maximum the announcing entity is willing to accept or is going to send. I think only the former makes sense, so maybe clarify by

```suggestion
    stream elements (including stanzas), in bytes the announcing entity is willing to accept. Guidance on
```

PR #1255, opened 2022-12-20
Add tools/triage.sh

### LINKED ISSUE #1237: Create tool to execute the triaging workflow
## Summary

The Editor has a more-or-less formalised triaging workflow described in [docs/TRIAGING.md](https://github.com/xsf/xeps/blob/master/docs/TRIAGING.md#flowchart). Doing these steps manually is tedious and error-prone.

## Work proposal

A tool should be created which, given two git commits (the PR head and the merge base between the PR head and the current main branch), triages the changes contained in that range and provides a machine-readable judgement and human-readable action items.

## Requirements

- MUST be accurate: In case of doubt, the tool should delegate to a human (by providing next steps on the output) instead of guessing.

- MUST provide machine-readable output: This is for future CI integration which may eventually post notices to GitHub or assign labels.

moparisthebest:
Fixes #1237

So for a first go this can be ran manually, or ran by a github action and dumped into a comment, or grep'd for "^tag" and used to automatically apply tags by a github action etc

### DIFF
--- a/tools/triage.sh
+++ b/tools/triage.sh
@@ -1,0 +1,207 @@
+#!/bin/bash
+
+# this should enforce as much of docs/TRIAGING.md as possible, run in top directory of xeps repo
+
+# usage: triage.sh $pr_title [$pr_commit] [$merge_base]
+# the last two will be HEAD and `git merge-base $pr_commit master` by default if not sent in
+
+# for testing you can:
+# git fetch origin +refs/pull/1254/merge
+# git checkout FETCH_HEAD
+# and only send in pr_title
+
+# test cases:
+# multiple xeps changed: git checkout 58cec1b6fb30ccc3b44c9dff34b94f53e958c1ac; ./tools/triage.sh 'XEP-0072, XEP-0096, XEP-0214: fix xmlns in sipub examples' 58cec1b6fb30ccc3b44c9dff34b94f53e958c1ac e4d3f721c2cf9fc7ed42c3428d098b2b3be3c8ed
+# protoxep: git fetch origin +refs/pull/1254/merge; git checkout FETCH_HEAD; ./tools/triage.sh "ProtoXEP: Stream Limits Advertisement"
+# one xep changed: git fetch origin +refs/pull/1249/merge; git checkout FETCH_HEAD; ./tools/triage.sh 'XEP-0444: bla'
+
+set -euo pipefail
+
+pr_title="$1"
+shift
+
+pr_commit=""
+if [ $# -ge 1 ]
+then
+    pr_commit="$1"
+    shift
+else
+    pr_commit="HEAD"
+fi
+
+merge_base=""
+if [ $# -ge 1 ]
+then
+    merge_base="$1"
+    shift
+else
+    merge_base="$(git merge-base "$pr_commit" master)"
+fi
+
+xpath() {
+    set -euo pipefail
+    xmllint --nonet --noent --xpath "$2" "$1" --nowarning --dtdvalid xep.dtd
+}
+
+files_changed="$(git --no-pager diff --name-only "$pr_commit" "$merge_base")"
+
+set +e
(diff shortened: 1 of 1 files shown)


tools/triage.sh
```diff
@@ -0,0 +1,205 @@
...
+        echo "error: multiple protoxeps cannot be created in 1 PR"
+        exit 1
+    fi
+    expected_title="ProtoXEP: $(xpath "$protoxep" '/xep/header/title/text()')"
+    if [ "$pr_title" != "$expected_title" ]
+    then
+        echo "tag: bad-title"
+        echo "error: expected title '$expected_title' got title '$pr_title'"
+        exit 1
+    fi
```
horazont, line 69: I wouldn't want to make this a hard error--that seems just superfluous bureaucracy to me. 

Enforcing the ProtoXEP prefix is useful (though labelling could replace that eventually), but let's not get too nitpicky about the overall title.

tools/triage.sh
```diff
@@ -0,0 +1,205 @@
...
+    echo "tag: ready-to-merge"
+    echo "manual: be sure type '$(xpath "$protoxep" '/xep/header/type/text()')' is appropriate for XEP content"
+    exit 0
+fi
```
horazont, line 73: Is this handling the case where someone mixes a protoxep submission with other XEP changes?

tools/triage.sh
```diff
@@ -0,0 +1,205 @@
...
+    
+    status="$(xpath "$xep" '/xep/header/status/text()')"    
+    approvers=""
+    if echo "$status" | grep -E '^(Experimental|Deferred)$' &>/dev/null
+    then
+        # Experimental or Deferred
+        if [ "$status" == 'Deferred' ]
+        then
+            echo "tag: needs-editor-action"
+            echo "manual: move $xep_num status to Experimental when merged"
```
horazont, line 120: The status change should happen in the same commit which writes the revision block.

tools/triage.sh
```diff
@@ -0,0 +1,205 @@
...
+
+if echo "$all_approvers" | grep '^Board$' &>/dev/null
+then
+    echo "tag: needs-board"
+fi
+
+if echo "$all_approvers" | grep '^Unknown$' &>/dev/null
+then
+    echo "tag: needs-unknown-approver"
+    echo "warning: approver is unknown, perhaps old XEP that needs fixed?"
```
horazont, line 197: This should be error/manual, shouldn't it?

horazont (commented):
Not fully awake anymore. Mostly looking good I think, some thoughts I left.

May need another round of more careful looking later, but maybe you already have thoughts on those thoughts.

tools/triage.sh
moparisthebest: I'd prefer to just put an `<approver/>` in all old XEPs and remove this

tools/triage.sh
moparisthebest: Currently this doesn't go by commit, it looks at the entire PR as 1 change :/

tools/triage.sh
moparisthebest: This is just a warning now

tools/triage.sh
moparisthebest: Oops good catch, now mixing these is a hard error

tools/triage.sh
horazont: Right, I meant that this should be part of the note for the editor. "when merged" can easily be read as "change status after merging" which would not be the correct action.

moparisthebest: I changed this language, let me know

tools/triage.sh
moparisthebest: 205 XEPs are missing approver :skull:

horazont: In favour of "error" then.

tmolitor-stud-tu:
What is missing/needed to move this one forward?

PR #1262, opened 2023-01-05
XEP-0353: Add recommendation for UUID v4 for id attributes

tmolitor-stud-tu:
I think explicit is better than implicit, hence this PR.

Since v0.4 of this XEP isn't implemented by any client yet (to my knowledge) and because the XEP is still experimental, I think we can do this update without bumping the namespace.
What do you thing @fippo @stpeter ?

### DIFF
--- a/xep-0353.xml
+++ b/xep-0353.xml
@@ -28,6 +28,14 @@
   &stpeter;
   &tmolitor;
   <revision>
+    <version>0.5.0</version>
+    <date>2022-01-05</date>
+    <initials>tm</initials>
+    <remark>
+      Recommend usage of UUID v4 for id attributes.
+    </remark>
+  </revision>
+  <revision>
     <version>0.4.0</version>
     <date>2021-11-27</date>
     <initials>tm</initials>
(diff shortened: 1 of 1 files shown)


fippo:
This requirement would need to go into XEP-0166, no?

tmolitor-stud-tu:
> This requirement would need to go into XEP-0166, no?

No, why should it? as long as the id requirements of 0166 and 0353 do not contradict each other having stricter requirements in 0353 is just fine.

tmolitor-stud-tu:
@fippo @stpeter Is this ok to be merged in its current state?

stpeter:
LGTM.

@fippo this is the `id` on the propose element, so it seems appropriate to specify it here (although of course we could update XEP-0166, too).

fippo:
I get that but it still introduces normative requirements on something defined in a different spec which is typically a bad idea in terms of procedure. I'll leave it to the council to decide whether that needs to be addressed so +0

Kev:
This has only a patch version bump, which is reserved for editorial changes, but this is a change to normative language.

tmolitor-stud-tu:
@Kev ah sorry, I did not know this. I'll fix this later today.

Kev:
That'll make life easier for the Editors, thanks. Otherwise I'll change it myself when merging.

tmolitor-stud-tu:
@Kev done

Kev:
Ta muchly.

PR #1265, opened 2023-01-20
Add Council as approver to all Standards Track XEPs where it was missing

moparisthebest:
```
# find all missing approvers
comm -12 <(grep -Lo '<approver>.*</approver>' xep-0* | sort -u) <(ls xep-0* | sort -u) > xepsmissing_approver.txt
# inspect types
cat xepsmissing_approver.txt | xargs grep -ho '<type>.*</type>' | sort -u > xepsmissing_uniquetypes.txt
# add council as approver to all with type Standards Track
cat xepsmissing_approver.txt | xargs grep -lo '<type>Standards Track</type>' | xargs sed -ri 's@^(.*)<sig>(.+)</sig>@\1<sig>\2</sig>\n\1<approver>Council</approver>@'
```

For clarification I think this should be considered a "tooling change" and no versions or anything bumped, this triggered it:
https://github.com/xsf/xeps/pull/1255#discussion_r1082096252

### DIFF
--- a/xep-0004.xml
+++ b/xep-0004.xml
@@ -13,6 +13,7 @@
     <status>Final</status>
     <type>Standards Track</type>
     <sig>Standards</sig>
+    <approver>Council</approver>
     <dependencies>
       <spec>XMPP Core</spec>
     </dependencies>
--- a/xep-0009.xml
+++ b/xep-0009.xml
@@ -13,6 +13,7 @@
   <status>Final</status>
   <type>Standards Track</type>
   <sig>Standards</sig>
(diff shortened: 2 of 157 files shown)


tmolitor-stud-tu:
@Kev I think after merging #1255, this should be merged, too, no?

Kev:
This doesn't pass CI at the moment.

iNPUTmice:
Council has approved this PR and asked if we could do this without a version bump. Council considers this just a bug in the meta data / an editorial change

PR #1269, opened 2023-02-15
Historical ProtoXEP: MUC Avatars

linkmauve:
The original ProtoXEP, submitted in https://github.com/xsf/xeps/pull/700, [got rejected](https://mail.jabber.org/pipermail/standards/2018-September/035351.html) due to there being too many XEPs.

Five years later, the situation hasn’t changed at all, nobody defined a new better protocol (my previous attempt [never went anywhere](https://github.com/xsf/xeps/pull/760) after I forgot to submit it to the list), and people still rely on tribal knowledge and from [Ejabberd](https://docs.ejabberd.im/tutorials/muc-vcard/) to write their implementations.

I therefore propose the exact same ProtoXEP, this time on the historical track. In my opinion, numbers are cheap, we can deprecate this one once we do have a better proposal.

### DIFF
--- a/inbox/muc-avatars.xml
+++ b/inbox/muc-avatars.xml
@@ -11,7 +11,7 @@
   &LEGALNOTICE;
   <number>xxxx</number>
   <status>ProtoXEP</status>
-  <type>Standard Track</type>
+  <type>Historical</type>
   <sig>Standards</sig>
   <approver>Council</approver>
   <dependencies>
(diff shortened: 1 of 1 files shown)


Kev:
Have pinged Edhelas to check that removal as author was desired. (Changes on inbox XEPs are horrid).

linkmauve:
The revert commit (the first one) just puts the ProtoXEP back to its original state.  I should add that I (as the original author) never got asked about merging the changes to make it non-specific to MUC, which imo was the point of the original ProtoXEP (documenting what implementations are already doing).

PR #1270, opened 2023-02-19
XEP-0424: Remove dependency on XEP-0422 'Message Fastening'

jcbrand:
Most recent discussion here: https://mail.jabber.org/pipermail/standards/2023-February/039164.html

CC @iNPUTmice

### DIFF
--- a/xep-0424.xml
+++ b/xep-0424.xml
@@ -20,13 +20,24 @@
     <spec>XMPP IM</spec>
     <spec>XEP-0313</spec>
     <spec>XEP-0421</spec>
-    <spec>XEP-0422</spec>
   </dependencies>
   <supersedes/>
   <supersededby/>
   <shortname>message-retract</shortname>
   &lance;
   &jcbrand;
+  <revision>
+    <version>0.4.0</version>
+    <date>2023-02-19</date>
+    <initials>jcb</initials>
+    <remark>
+        <ul>
+          <li>Remove the dependency on XEP-0422 Message Fastening.</li>
+          <li>Use 'id' attribute on 'retracted' element instead of a child element.</li>
+          <li>Clarify business rlies regarding which 'id' to use to refer to the retracted message.</li>
+        </ul>
+    </remark>
+  </revision>
   <revision>
     <version>0.3.0</version>
     <date>2020-05-17</date>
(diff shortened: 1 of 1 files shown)


xep-0424.xml
```diff
@@ -20,13 +20,18 @@
...
     <spec>XEP-0421</spec>
-    <spec>XEP-0422</spec>
   </dependencies>
   <supersedes/>
   <supersededby/>
   <shortname>message-retract</shortname>
   &lance;
   &jcbrand;
+  <revision>
+    <version>1.0.0</version>
```
mar-v-in, line 30: Hmm. Next version is 0.4.0. 1.0.0 will be used once the XEP moves to stable.

xep-0424.xml
```diff
@@ -120,9 +125,7 @@
 
   <example caption="The client sends out a retraction message"><![CDATA[
 <message type='chat' to='lord@capulet.example' id='retract-message-1'>
-  <apply-to id="origin-id-1" xmlns="urn:xmpp:fasten:0">
-    <retract xmlns='urn:xmpp:message-retract:0'/>
-  </apply-to>
+  <retract id="origin-id-1" xmlns='urn:xmpp:message-retract:1'/>
```
mar-v-in, line 128: You are changing the example here without changing the normative description a few lines above which is still pointing at XEP-0422.

mar-v-in, line 128: I think it would be preferably if the id used is:
1. XEP-0359 origin-id or id from the message id attribute in direct chats
2. XEP-0359 stanza-id assigned by the MUC in MUCs.

As I mentioned on the mailing list, I'll write these rules into a informational XEP soonish (which once available can be referenced from here), but I'd want to make sure we update the rule at the same time we namespace bump (so we won't need another ns bump). Feel free to just copy the rules from https://xmpp.org/extensions/xep-0444.html#business-id

xep-0424.xml
```diff
@@ -167,7 +170,7 @@
     <forwarded xmlns='urn:xmpp:forward:0'>
       <delay xmlns='urn:xmpp:delay' stamp='2019-09-20T23:08:25Z'/>
       <message type="groupchat" from="romeo@montague.example" to="lord@capulet.example" id="wrong-recipient-1">
-        <retracted stamp='2019-09-20T23:09:32Z' xmlns='urn:xmpp:message-retract:0'>
+        <retracted stamp='2019-09-20T23:09:32Z' xmlns='urn:xmpp:message-retract:1'>
           <origin-id xmlns='urn:xmpp:sid:0' id='origin-id-1'/>
```
mar-v-in, line 174: This should probably be changed to using a stanza-id instead of origin-id in cases where the MUC assigned stanza-id was used.

iNPUTmice: The id should probably just be an attribute on `<retracted/>` called `id` and then depending on context it easier refers to the stanza-id or the message id.

xep-0424.xml
jcbrand: Thanks, fixed.

xep-0424.xml
jcbrand: Fixed.

jcbrand: Thanks, updated accordingly.

xep-0424.xml
jcbrand: > The id should probably just be an attribute on <retracted/> called id and then depending on context it easier refers to the stanza-id or the message id.

Done.

iNPUTmice:
Council has voted to move XEP-0424 back to Experimental

jcbrand:
@Kev is there something preventing this PR from being merged? Can I go ahead and just do it?

lovetox:
The usage of stanza-id in MUCs is not necessary for this XEP

This XEP always references only the senders own messages. As long as the sender uses stable-id as message-id all is good.
There needs to be no mention of origin-id or stable-id.
All messages will be namespaced with the senders JID, there will be no conflicts, and if the sender uses non-stable message-ids he will only shoot himself in the foot, it can do no harm to the receiver.

A simply note for the implementation that a client SHOULD use stable message-id is enough, and we lose a whole dependency on a XEP, and make implementation easier.

PR #1271, opened 2023-02-19
XEP-0425: Updates based on list feedback

jcbrand:
- Remove the dependency on XEP-0422 Message Fastening
- Rename to 'Moderated Message Retraction' and focus only on the retraction use-case
- Ensure compatibility with clients that only implement XEP-0424
- Clarify the purpose of the `reason` element

Mailing list thread is here:
https://mail.jabber.org/pipermail/standards/2021-December/038660.html


iNPUTmice:
I think something is wrong with the examples. I assume `oldhag` is the sender of the original message? And the room 'fakes' a retract on behalf of `oldhag`? Then it should look something like this:

(Previously the moderation messages was validated by coming from the bare room jid; but if you want to be compatible with retraction it will need to be validated by the sender (+ occupant id maybe)) 
```xml
<message type="groupchat" id='retraction-id-1' from='room@muc.example.com/oldhag' to="account@example.com/resource">
  <retract id="stanza-id-1" xmlns='urn:xmpp:message-retract:1'>
    <moderated by='room@muc.example.com/macbeth' xmlns='urn:xmpp:message-moderate:1'>
      <reason>This message contains inappropriate content for this forum</reason>
    </moderated>
  </retract>
</message>
```

xep-0425.xml
```diff
@@ -182,8 +194,8 @@
 
 <xs:schema
     xmlns:xs='http://www.w3.org/2001/XMLSchema'
-    targetNamespace='urn:xmpp:message-moderate:0'
-    xmlns='urn:xmpp:message-moderate:0'
+    targetNamespace='urn:xmpp:message-moderate:1'
+    xmlns='urn:xmpp:message-moderate:1'
     elementFormDefault='qualified'>
 
   <xs:element name='moderate'></xs:element>
```
Zash, line 201: Nit: Shouldn't the schema gain an `id` attribute now?

xep-0425.xml
```diff
@@ -84,30 +96,28 @@
...
+    <retract xmlns='urn:xmpp:message-retract:1'/>
+    <reason>This message contains inappropriate content for this forum</reason>
+  </moderate>
 </iq>]]></example>
 
   <section2 topic='Success case' anchor='usecase-success'>
 
   <p>If the moderator is allowed to moderate the message, the groupchat service will send a message stanza to all participants (including the moderator), indicating that the message has been retracted and by whom.</p>
-  <p>This message will use &xep0422; to indicate that it applies to the moderated message, by referring to its XEP-0359 Stanza ID.</p>
+  <p>This message will use &xep0422; to indicate that it is a retraction, and will refer to the retracted message by its XEP-0359 Stanza ID. Inside the &lt;retract/&gt; element is a &lt;moderated/&gt; element indicating that the retraction was a moderator action and containing the optional reason.</p>
```
Flowdalic, line 112: A reference to a xep359 stanza-id should always use the 'id' *and* the 'by' attribute (not just the 'id'). See https://mail.jabber.org/pipermail/standards/2023-February/039168.html

xep-0425.xml
jcbrand: Thanks, added.

jcbrand:
> I assume `oldhag` is the sender of the original message? And the room 'fakes' a retract on behalf of `oldhag`? Then it should look something like this:

As mentioned in the XSF MUC, I don't think it's appropriate to pretend that a retraction comes from someone it doesn't actually come from and I doubt that maintaining compatibility with clients that only implement XEP-0424 is worth it.

singpolyma:
> Remove the dependency on XEP-0422 Message Fastening

Doesn't that just make it more complex to implement for no benefit?

jcbrand:
@singpolyma: No, I think it's the contrary. I added message fastening in the anticipation that it would become  a widely implemented standard. That didn't happen, in fact nobody besides myself made any attempt (that I'm aware of) to actually use fastenings. So it's a dead end and unnecessary cruft.

xep-0425.xml
mar-v-in, line 112: I think you wanted to refer to &xep0424; instead of &xep0422;

jcbrand: Thanks, fixed.

mar-v-in (changes requested):

xep-0425.xml
```diff
@@ -19,12 +19,24 @@
...
     <spec>XMPP IM</spec>
     <spec>XEP-0313</spec>
-    <spec>XEP-0422</spec>
   </dependencies>
   <supersedes/>
   <supersededby/>
   <shortname>message-moderation</shortname>
   &jcbrand;
+  <revision>
+    <version>0.2.2</version>
```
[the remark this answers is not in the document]
jcbrand: Makes sense, thanks.

xep-0425.xml
jcbrand: Doesn't look to me like there's uniformity on this. I've added more text to the XEP to clarify that the `by` attribute of the original stanza-id should be from the MUC.

jcbrand:
@Kev, @iNPUTmice Is there anything still preventing this from being merged?

Kev:
> @Kev, @iNPUTmice Is there anything still preventing this from being merged?

Waiting for Council to get back to me on this.

xep-0425.xml
```diff
@@ -84,30 +97,29 @@
...
-  </apply-to>
+  <moderate id="stanza-id-1" xmlns='urn:xmpp:message-moderate:1'>
+    <retract xmlns='urn:xmpp:message-retract:1'/>
+    <reason>This message contains inappropriate content for this forum</reason>
+  </moderate>
 </iq>]]></example>
 
   <section2 topic='Success case' anchor='usecase-success'>
 
   <p>If the moderator is allowed to moderate the message, the groupchat service will send a message stanza to all participants (including the moderator), indicating that the message has been retracted and by whom.</p>
```
moparisthebest, line 112: Is this the only protection against a non-moderator issuing a retraction? If so it's not enough because the MUC may not understand this protocol and therefore won't be able to perform this filtering.

Regardless, how, when, and where retractions are supposed to be validated should be in the security considerations section.

iNPUTmice:
Council has voted to move XEP-0425 back to experimental (which gives the author control over those changes)

xep-0425.xml
```diff
@@ -84,30 +97,29 @@
...
+  <p>This message will use &xep0424; to indicate that it is a retraction, and will refer to the retracted message by its XEP-0359 Stanza ID. Care must be taken to use the stanza ID that was added by the MUC, i.e. which has a "by" attribute set to the MUC JID, and not some other stanza ID. The client MUST check that the correct stanza ID was used and ignore the stanza if not.</p>
+  <p>Inside the &lt;retract/&gt; element is a &lt;moderated/&gt; element indicating that the retraction was a moderator action and containing the optional reason.</p>
 
 <example caption='The groupchat service sends a moderation message to all participants'><![CDATA[
 <message type="groupchat" id='retraction-id-1' from='room@muc.example.com' to="room@muc.example.com/macbeth">
-  <apply-to id="stanza-id-1" xmlns="urn:xmpp:fasten:0">
-    <moderated by='room@muc.example.com/macbeth' xmlns='urn:xmpp:message-moderate:0'>
-      <retract xmlns='urn:xmpp:message-retract:0' />
+  <retract id="stanza-id-1" xmlns='urn:xmpp:message-retract:1'>
+    <moderated by='room@muc.example.com/macbeth' xmlns='urn:xmpp:message-moderate:1'>
```
lovetox, line 119: here should be a <occupant-id> element added in the example

xep-0425.xml
```diff
@@ -84,30 +97,29 @@
...
+    <reason>This message contains inappropriate content for this forum</reason>
+  </moderate>
 </iq>]]></example>
 
   <section2 topic='Success case' anchor='usecase-success'>
 
   <p>If the moderator is allowed to moderate the message, the groupchat service will send a message stanza to all participants (including the moderator), indicating that the message has been retracted and by whom.</p>
-  <p>This message will use &xep0422; to indicate that it applies to the moderated message, by referring to its XEP-0359 Stanza ID.</p>
+  <p>This message will use &xep0424; to indicate that it is a retraction, and will refer to the retracted message by its XEP-0359 Stanza ID. Care must be taken to use the stanza ID that was added by the MUC, i.e. which has a "by" attribute set to the MUC JID, and not some other stanza ID. The client MUST check that the correct stanza ID was used and ignore the stanza if not.</p>
+  <p>Inside the &lt;retract/&gt; element is a &lt;moderated/&gt; element indicating that the retraction was a moderator action and containing the optional reason.</p>
```
lovetox, line 114: Please describe the elements that should or must be in the message here when the message is first introduced

- Is the "by" attributed a MUST, or optional?
- Is the "by" attribute set to a real JID in non-anon rooms? Or always the occupant JID?
- Should a server that understands the occupant-id XEP add a <occupant-id> tag inside the moderated tag?

Some of these things are defined later down in the Tombstone section, which someone may not even read because they have no interest in implementing tombstones, and my interpret only relevant for the tombstone use case.

lovetox:
What logic does the `<moderate/d>` and `<retract/ed>` follow? It does not seem to make much sense for me currently.

##  Sending IQ
```
<moderate>
    <retract />
</moderate>
```

Why is there an empty retract element in there? What purpose does this serve? Its not optional, and conveys no additional information that is not already clear from receiving the `<moderate>`

## Receiving the message live

```
<retract>
    <moderated/>
</retract>
```

This is ok i guess if we think moderation as special case of retracting.

## Receiving a Tombstone

```
<moderated>
    <retracted/>
</moderated>
```

This is now the complete opposite from above, now we turned it around again, why?

In my opinion, the tombstone case should be turned around to

```
<retracted>
    <moderated/>
</retracted>
```

and in the IQ we can lose the `<retract>` because it serves no purpose.

lovetox:
I just discovered there seems to be no protection against forging tombstones.

xep-0425.xml
lovetox: > To prevent message spoofing, it's very important to check that the moderation message comes from the MUC service 

its already in the security considerations, this is enough, don’t you think?

Kev:
@jcbrand So I think this is ok for you to update to change back to Experimental in the same change, and then I can merge.

jcbrand:
> What logic does the <moderate/d> and <retract/ed> follow? It does not seem to make much sense anymore now.

It's the difference between a request (or command) and a response. Seems to me it's mainly a matter of taste whether one should distinguish like this or not. I prefer it, since it removes ambiguity once you understand the point. If you get `retracted`, you know this is a message that has already been retracted. If you get `retract` it means you (as the client) are being asked to retract a message in the chat history.

> My Proposal, move id to retract element.

Thanks, but I don't really see the need to change this. The message is being moderated, and the moderation action happens to be a retraction. That was the idea behind putting the `id` on the `moderate` element.

>  later in the tombstone use case we depend on the already defined stamp attribute on the retract element

That is on the `retracted` element, which is something else. I think I prefer the `id` attribute on the `moderate` element.

> The <apply-to> was replaced with the <retract> but now its inconsistent with the tombstone and the IQ where moderated is always the parent.

Ok, I updated the tombstone so that `retracted` is the parent of `moderated`.

> Further please add <occupant-id> into the example and add a sentence maybe.

Done.

> Example looks good and consistent, but on the <moderated> element the namespace is missing !

Thanks, fixed.

> Please add a sentence how to discover a tombstone

There already is a sentence above the tombstone example which explains this.

To try and reduce confusion, I've added the following sentence:

> Note the past tense in the element names, e.g. "retracted" instead of just "retract", to indicate that the retraction has already been executed and that this is therefore a tombstone and not an action that should be applied.

jcbrand:
@Kev I've moved the status back to `experimental`, can you please go ahead and merge?

xep-0425.xml
```diff
@@ -142,11 +157,12 @@
       <delay xmlns='urn:xmpp:delay' stamp='2019-09-20T23:18:41Z'/>
       <message type="groupchat" from="room@muc.example.com/oldhag" to="room@muc.example.com/macbeth" id="message-id-1">
         <occupant-id xmlns="urn:xmpp:occupant-id:0" id="ef73b09de9c90a38ba552f7c68cbeef1dca24429" />
-        <moderated by="witch@shakespeare.example">
-          <occupant-id xmlns="urn:xmpp:occupant-id:0" id="dd72603deec90a38ba552f7c68cbcc61bca202cd" />
-          <retracted stamp='2019-09-20T23:19:12Z' xmlns='urn:xmpp:message-retract:0'/>
+        <retracted stamp='2019-09-20T23:19:12Z' xmlns='urn:xmpp:message-retract:0'>
```
truenicoco, line 160: Update to `urn:xmpp:message-retract:1`?

PR #1272, opened 2023-02-20
XEP-0359 (STANZA-ID): Add security consideration and <referenced-stanza/>

Flowdalic:


Kev:
Could you make the version block 0.7.0 rather than 0.7 so the new triage.sh script is happy for me, please?

Flowdalic:
> Could you make the version block 0.7.0 rather than 0.7 so the new triage.sh script is happy for me, please?

Guus did so, I took the liberty to squash his commit into mine as I relaxed the requirements for the 'by' attribute a bit.

PR #1278, opened 2023-04-02
XEP-0402: Replace arbitrary max items with 'max'

linkmauve:


linkmauve:
@Kev I added a version block here as well.

iNPUTmice:
Council has approved this PR

PR #1281, opened 2023-04-06
XEP-0372: urn:xmpp:reference:1

edhelas:
- Redefine types
- Bump the namespace.
- Make uri parameter non mandatory
- Remove the mentions chapter (defined in another XEP).
- Add reference to XEP-0426
- Add the XML schema
- Add restrictions around the begin and end attributes
- Complete the discovery section
- Add the optional hreflang attribute


singpolyma:
Why are mentions being removed? Isn't that most of the point of this XEP?

I'm also not sure we need to bump the namespace as this will break everyone using references today.

Ppjet6:
The goal for it was to make the spec generic enough and without specific features, so that other specs could define their own types.

Mentions were supposed to be re-added with something like https://bouah.net/specs/mentions.html but I'm not going to pursue on this.

I guess personally I'd settle with just the following for the moment:
- Add reference to XEP-0426
- Add the XML schema
- Add restrictions around the begin and end attributes
- Add the optional hreflang attribute

I do think making it really generic is the most interesting part of the PR, but that also makes me wonder if the spec just isn't too broad.

Anyway this is still "Needs Author".

Kev:
Hadn't clicked that this is waiting on my review, sorry. Have added it to my TODO.

PR #1286, opened 2023-06-28
2xep tool fixes

mwild1:
This tool is handy to import from e.g. Markdown to XEP XML. Unfortunately I had problems with it emitting XML that wouldn't build. The changes in this branch allowed me to build the resulting XML using pandoc 2.9.2.1.


tools/2xep.lua
```diff
@@ -180,8 +183,7 @@ function Strikeout(s)
 end
 
 function Link(s, src, tit, attr)
-  return "<a href='" .. escape(src,true) .. "' title='" ..
-         escape(tit,true) .. "'>" .. s .. "</a>"
```
Zash, line 184: I take it the XEP schema doesn't have a `@title` equivalent?

PR #1289, opened 2023-07-19
XEP-0333: Change 'discover support' example

guusdk:
The 'discover support' example should not use a server/domain JID, as there's precious little for servers to support. Instead, a client JID is used.


iNPUTmice:
Council has approved this (but agrees that this is editorial anyway)

PR #1292, opened 2023-09-04
roominfo_changesubject does not exist

singpolyma:
The spec everywhere else specifies it should be called roomconfig_changesubject and this is also present in this example as well lower down.


Neustradamus:
@singpolyma: Badly, the problem is not new: You can see here a @zinid and @iNPUTmice talk:
- https://github.com/xsf/xeps/pull/667#issuecomment-398299921

iNPUTmice:
Council has approved this PR

PR #1293, opened 2023-09-07
XEP-0060: Release version 1.26.0

melvo:
Fix and add examples for publishing items:

* Remove 'item' element specifying its ID from example response for publishing item with ID
* Add examples for publishing item without ID

### DIFF
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -51,6 +51,14 @@
   &stpeter;
   &ralphm;
   <revision>
+    <version>1.26.0</version>
+    <date>2023-09-07</date>
+    <initials>melvo</initials>
+    <remark>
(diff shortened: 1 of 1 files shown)


melvo:
@Kev Does that bug fix release really needs the council? If so, when will they decide on it?

Kev:
By definition a bug fix is a substantive change, which means it needs Council, yes, sorry. Will poke again.

iNPUTmice:
Council has approved this PR under the condition that the change log gets changed to
> Add examples for publishing item without ID

melvo:
@Kev @iNPUTmice Thanks for the feedback. It should be ready to be merged now.

Kev:
This isn't an editorial change, so it should advance x.Y.z instead of x.y.Z.

PR #1299, opened 2023-11-25
XEP-0077: Release version 2.5

melvo:
Specify account creation via HTTP as non-exclusive alternative to XMPP

### DIFF
--- a/xep-0077.xml
+++ b/xep-0077.xml
@@ -13,6 +13,7 @@
   <status>Final</status>
   <type>Standards Track</type>
   <sig>Standards</sig>
+  <approver>Council</approver>
   <dependencies>
     <spec>XMPP Core</spec>
   </dependencies>
@@ -23,6 +24,12 @@
     <url>http://www.xmpp.org/schemas/iq-register.xsd</url>
   </schemaloc>
   &stpeter;
+  <revision>
+    <version>2.5</version>
+    <date>2023-11-25</date>
+    <initials>melvo</initials>
+    <remark><p>Specify account creation via HTTP as non-exclusive alternative to XMPP</p></remark>
+  </revision>
   <revision>
     <version>2.4</version>
     <date>2012-01-25</date>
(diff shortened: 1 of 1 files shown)


xep-0077.xml
```diff
@@ -517,9 +524,12 @@
   </ol>
   <p>Support for extensibility via <cite>Data Forms</cite> is RECOMMENDED but is not required for compliance with this document.</p>
 </section1>
-<section1 topic='Redirection' anchor='redirect'>
-  <p>A given deployment MAY wish to redirect users to another medium (e.g., a website) for registration, rather than allowing in-band registration. The recommended approach is to include only the &lt;instructions/&gt; element rather than the required fields or a data form in the IQ result, as well as a URL encoded using &xep0066; (see the <link url="#precedence">Precedence Order</link> section below for further details).</p>
-  <example caption='Host Redirects Entity to Web Registration'><![CDATA[
+<section1 topic='Alternative Registration Medium' anchor='alternative-registration-medium'>
+  <p>A given deployment MAY wish to provide an alternative medium (e.g., a website) for registration.</p>
+  <section2 topic='Web Registration' anchor='web-registration'>
```
mar-v-in, line 529: This invalidates prior links to this subsection (https://xmpp.org/extensions/xep-0077.html#redirect). I propose we keep that URL as is (by keeping `anchor='redirect'`) as is. But this is probably editorial and thus is on the editor to decide.

melvo: I thought about that as well when I changed the anchor but keeping it would result in a misleading anchor. Is there any solution for aliases/redirections from old URLs to new ones even if they do not exactly match?

melvo: @Kev What should I do?

iNPUTmice:
Council has rejected this for the following reasons:
* -1 from my side. The language is weird. The justification for the change is now in the xep. I don't like changing final xeps without prior discussion
* I even have doubts about the usecase in general. That itself makes it rather unacceptable for me to do this change to something that's already Final. -1, but might change after discussion and wording change

PR #1301, opened 2023-11-30
XEP-0458 Last Call feedback

stpeter:
Please do not merge this until @dwd and @stpeter agree on the changes.


Kev:
(The checking script complains about 0.4 rather than 0.4.0 BTW, although I'm not sure how much we should care)

stpeter:
I changed 0.4 to 0.4.0. And yes, I think this can be merged now, thanks.

PR #1294, opened 2023-09-12
XEP-0198: Add section defining SASL2 and BIND2 interaction

iNPUTmice:

PR #1303, opened 2023-12-07
XEP-0388: fix stream managment example

iNPUTmice:
After merging  #1294 XEP-0388 needs it's example updated.


Kev:
(Matt okays out of band)

mwild1:
I approve. Thanks!

PR #1304, opened 2023-12-11
Use link to immutable versioned XEP in biblatex's "url" entry

Flowdalic:
One major problem when citing XEPs is that they are often cited using links to the living, that is mutable, version of the XEP. For example: https://xmpp.org/extensions/xep-0123.html

However, we really want people to be able to cite an immutable document. Luckily, we have the [`attic/`](https://xmpp.org/extensions/attic/) where we keep immutable versioned XEPs that can be referenced.

This PR changes the XSL so that the immutable versioned XEP URL is used for the generated biblatex entry.

However, it is unfortunately not read yet. Primarily because we do not seem to keep a immutable versioned variant of the *latest* version of a XEP. If we could change the editor workflow to also create a versioned XEP in `attic/` for the latest version of the XEP, then this could be merged.

That said, purely for aesthetic reasons I would prefer to not have the term `attic` in the part of the URL. Maybe we can change this to `versioned` (or something else)?

### DIFF
--- a/xep.xsl
+++ b/xep.xsl
@@ -528,6 +528,23 @@ content: "XEP-<xsl:value-of select='/xep/header/number'/>";
             </xsl:otherwise>
           </xsl:choose>
         </xsl:variable>
+        <xsl:variable name="bibtex-url">
+          <xsl:choose>
+            <!-- XEPs after a certain number have an immutable version to which we preferably link -->
+            <xsl:when test='/xep/header/number &gt; 300'>
+              <xsl:text>https://xmpp.org/extensions/attic/xep-</xsl:text>
+              <xsl:value-of select='/xep/header/number'/>
+              <xsl:text>-</xsl:text>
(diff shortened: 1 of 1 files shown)


Kev:
> That said, purely for aesthetic reasons I would prefer to not have the term attic in the part of the URL. Maybe we can change this to versioned (or something else)?

You'd like to change the immutable URLs from `/attic/` to `/versioned/` ? ;)

(This is above my pay grade, in any case)

Kev:
I'm happy for someone to patch the tooling so it archives the latest version too, BTW, but am not inclined to make the change myself at the moment.

Flowdalic:
> I'm happy for someone to patch the tooling so it archives the latest version too, BTW, but am not inclined to make the change myself at the moment.

Then we should disable the logic for now: https://github.com/xsf/xeps/pull/1308

PR #1309, opened 2023-12-11
Makefile: remove .SILENT, make the targets loud

Flowdalic:
The Makefile was set globally to silent, which caused the failing commands to be *not* displayed. For example

```bash
flo@neo-pc xeps-xsf $ make html
make: *** [Makefile:112: build/xep-0001.html] Error 1
```

Removing this provides usefull information about which command failed:

```bash
flo@neo-pc xeps-xsf $ make html
xmllint --nonet --noout --noent --loaddtd --valid "xep-0001.xml" ! xmllint --nonet --noout --noent --loaddtd --xpath "//img/@src[not(starts-with(., 'data:'))]" xep-0001.xml 2>/dev/null && true make: *** [Makefile:110: build/xep-0001.html] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,3 @@
-.SILENT:
-
 OUTDIR?=build
 REFSDIR?=$(OUTDIR)/refs
 EXAMPLESDIR?=$(OUTDIR)/examples


deuill:
Seems that the underlying issue here is the same as encountered in #1316, were you able to solve these somehow @Flowdalic?

PR #1310, opened 2023-12-11
Extract the copyrigh years of a XEP from its revision

Flowdalic:
In https://github.com/xsf/xeps/pull/1300#issuecomment-1850475278 it was said that we do not longer simply bump the copyright year range. But keeping the copyright year at 2021 of all XEPs, including newer ones from 2023, is also far from ideal.

So let's simply extract the copyright year from the revision information we already have.


mwild1:
This looks like a very reasonable solution, thanks!

Kev:
Thanks Flow.

PR #1315, opened 2024-01-15
advance XEP-0458 to Active

stpeter:
This PR advances XEP-0458 to Active per the Board vote on January 5, 2024.

@Kev - let me know if changes are needed for proper advancement.


Kev:
The tooling requires versions to look semverish (three parts). I'll fix this up.

PR #1320, opened 2024-02-02
XEP-0227: Add support for MUC export, and clean up examples

linkmauve:
A room export can contain:
- the affiliations of the users
- the configuration of the room
- the MAM archive
- a vCard-temp containing its avatar

Examples for all of those are included, I don’t expect any further data for now but newer versions of this XEP can obviously add more.

PR #1321, opened 2024-02-02
XEP-0313: Fix MUC archive example

linkmauve:
It had duplicated ids, and the JIDs of the witches weren’t consistent with their definition in XEP-0045.

These issues were found while writing #1320.

### DIFF
--- a/xep-0313.xml
+++ b/xep-0313.xml
@@ -31,6 +31,18 @@
   &mwild;
   &ksmith;
   <revision>
+    <version>1.1.1</version>
+    <date>2023-12-15</date>
+    <initials>egp</initials>
+    <remark>
+      <ul>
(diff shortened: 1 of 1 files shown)


xep-0313.xml
```diff
@@ -30,6 +30,18 @@
   </schemaloc>
   &mwild;
   &ksmith;
+  <revision>
+    <version>1.1.1</version>
+    <date>2023-12-15</date>
+    <initials>egp</initials>
+    <remark>
+      <ul>
+        <li>Fix JID and affilitation of the first two witches in the MUC example.</li>
```
Fishbowler, line 39: Small typo
```suggestion
        <li>Fix JID and affiliation of the first two witches in the MUC example.</li>
```

PR #1322, opened 2024-02-04
xep-template: Require accessibility and privacy sections

mathieui:
The Privacy section is a new one, and Accessibility was previously optional.

Rationale for the new section: it makes sense to distinguish between security and privacy, even if there is an overlap.

Rationale for requiring both: not making accessibility and privacy part of the core things required for a XEP is bad in this day and age, even if I expect that most XEPs will not have much applicable on those topics. It will make sure the authors have to consider those aspects before submitting.


iNPUTmice:
Council has accepted this modification

PR #1327, opened 2024-02-29
XEP-0447: Release version 0.3.1

mar-v-in:


mar-v-in:
The validation failure seems to be on the side of GitHub actions...

PR #1335, opened 2024-03-20
XEP-0428: Change schema int types to xs:unsignedInt

lnjX:
The previous type was xs:integer, which allows for arbitrary large
integers, and also negative integers. Negative values do not make sense
when referencing characters and limiting the size of the integer to 32
bit is, I think, reasonable as that would still allow referencing in 4
GiB of text (and that should probably exceed any stanza size limit).

### DIFF
--- a/xep-0428.xml
+++ b/xep-0428.xml
@@ -25,6 +25,12 @@
   &dcridland;
   &larma;
   <revision>
+    <version>0.3.0</version>
+    <date>2024-03-20</date>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
ping @mar-v-in and @dwd

mar-v-in:
Change is fine, except that schema is not normative and therefore this is an editorial change, thus should be version 0.2.1, not 0.3.0

PR #1350, opened 2024-06-26
Tag roster-related XEPs

avidseeker:
See: https://wiki.xmpp.org/web/Roster for more info


avidseeker:
I agree.

Signed.

- Avid Seeker

PR #1361, opened 2024-07-10
XEP-0045: Improved wording of status code purpose

guusdk:
Previously, the wording of the status code purpose incorrectly hinted that the user receiving the presence stanza with the status code is the user that is affected by the corresponding change. This is incorrect, as these status code are not only used in stanzas sent to the affected user. They are also used to inform remaining occupants of the room of the status change.


guusdk:
There was a brief discussion about this in the XSF chatroom. Logs: https://logs.xmpp.org/xsf/2024-07-10#2024-07-10-947697cc820aca29

PR #1365, opened 2024-07-13
XEP-0045: Presence sent to occupants of a destroyed room includes a `<destroy/>` element

guusdk:
Even if no reason or alternate venue is provided, a `<destroy/>` element is needed to distinguish between other presence unavailable.


iNPUTmice:
Council said the MUST is better changed to a SHOULD so it is backward compatible with implementations that don’t include it.

PR #1369, opened 2024-08-06
XEP-0388: Fix the XML Schema and examples

linkmauve:


xep-0388.xml
```diff
@@ -686,131 +695,141 @@
...
+
+  <xs:simpleType name="Jid">
+    <xs:restriction base="xs:string">
+      <xs:maxLength value="3071"/>
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="Uuid">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="([0-9]|[a-f]|[A-F]){8}-([0-9]|[a-f]|[A-F]){4}-([0-9]|[a-f]|[A-F]){4}-([0-9]|[a-f]|[A-F]){4}-([0-9]|[a-f]|[A-F]){12}"/>
```
Fishbowler, line 819: Is there a historical reason that `([0-9]|[a-f]|[A-F]){8}` isn't written as `([0-9a-fA-F]){8}` ?
Maybe compatibility with certain tools?

linkmauve: I probably just copy/pasted it from somewhere else, feel free to PR that change as it makes it slightly more readable. :)

PR #1370, opened 2024-08-14
XEP-0045: Add explicit error definition when non-owners attempt to use owner-specific functionality

guusdk:
XEP-0045 explicitly states that modifying the Owner List (section 10.5) is 'forbidden' to any non-owner, but does not explicitly state that error in 10.3 (Granting Owner Status) and 10.4 (Revoking Owner Status). The same issue is there in the sections 10.6, 10.7 and 10.8 (granting/revoking Admin Status).

This commit copies the text that's used throughout the specification ("If the \<user@host> of the 'from' address does not match the bare JID of a room owner, the service MUST return a <forbidden/> error to the sender.") to the sections where it's missing.

### DIFF
--- a/xep-0045.xml
+++ b/xep-0045.xml
@@ -47,6 +47,12 @@
   <registry/>
   &stpeter;
   <revision>
+    <version>1.35.1</version>
+    <date>2024-09-17</date>
+    <initials>gk</initials>
+    <remark><p>Add explicit error definition when non-owners attempt to use owner-specific functionality.</p></remark>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Hi @guusdk 

council has approved those changes. Would you mind rebasing this on top of current master?

guusdk:
Hello @iNPUTmice - It is now rebased! Thanks for all of the editorial work you've been putting in this, lately!

PR #1377, opened 2024-08-29
Replace all HTML `cite` elements with XML entities

guusdk:
Prior to this change, some XEPs are being referenced by an XML entity (eg `&xep0030;`) while others use a HTML cite element (eg `<cite>XEP-0030</cite>`). The latter doesn't seem to do anything with regards to rendering (there's no link to the reference list, for example).

All references to XEPs (and RFCs) _should_ be based on XML entities, as (briefly) discussed in the XSF Discussion chatroom: https://logs.xmpp.org/xsf/2024-08-29#2024-08-29-67941f0944ae8365

The changes in this commit are the result of executing these two commands:

```
sed -i -E 's/<cite>XEP-([0-9]{4})<\/cite>/\&xep\1\;/' xep-*.xml
sed -i -E 's/<cite>RFC ([0-9]{4})<\/cite>/\&rfc\1\;/' xep-*.xml
```


guusdk:
This PR needs the fix for validation of multi-file PRs as provided in https://github.com/xsf/xeps/pull/1378

PR #1378, opened 2024-08-29
CI: improve iteration over multiple changed files

guusdk:
The quotation made the list of files be interpreted as one very long filename.

Split off in a environment variable to better match the examples provided by tj-actions/changed-files

### DIFF
--- a/.github/workflows/xep-validation.yml
+++ b/.github/workflows/xep-validation.yml
@@ -24,10 +24,12 @@ jobs:
 
       - name: Validate changed file(s)
         if: steps.changed-xeps.outputs.any_changed == 'true'
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-xeps.outputs.all_changed_files }}
         run: |
           sudo apt-get install -y libxml2-utils
           result=0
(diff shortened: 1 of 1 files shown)


.github/workflows/xep-validation.yml
```diff
@@ -24,10 +24,12 @@ jobs:
 
       - name: Validate changed file(s)
         if: steps.changed-xeps.outputs.any_changed == 'true'
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-xeps.outputs.all_changed_files }}
         run: |
           sudo apt-get install -y libxml2-utils
           result=0
-          for xep in "${{ steps.changed-xeps.outputs.all_changed_files }}"; do
+          for xep in ${ALL_CHANGED_FILES}; do
```
linkmauve, line 32: I believe you could make this change simpler just by removing the quotes here (untested):
```suggestion
          for xep in ${{ steps.changed-xeps.outputs.all_changed_files }}; do
```

guusdk: ... possibly, but I definitely tested the commit as-is, and I don't think the proposed modification makes things considerably more easy. I'm happy to go with the commit as-is, unless someone feels strongly about this and wants to test the alternative.

PR #1382, opened 2024-08-29
XEP-0188: Add missing 'approver' element.

guusdk:
I'm not sure if this needs a revision block.


guusdk:
Triage validation fails with:
> move XEP-0188 status to Experimental before or with merge
I would argue that we can ignore this. The change in this PR is a formalization of an implied previous approver only.

PR #1387, opened 2024-08-30
XEP-0004: Empty and absent values

guusdk:
The pre-existing specification does not differentiate between empty or absent values. As the XEP has reached the 'final' status, such definitions can not be added to this XEP now.

This commit doesn't (intend to) add new definitions. It merely documents consequences of what is the current final spec.

The intention of this change is to make explicit that users of this XEP cannot expect this distinction to be present (unless it is explicitly defined in dependant XEPs). As a result implementations can be expected to use a variety of representations.

### DIFF
--- a/xep-0004.xml
+++ b/xep-0004.xml
@@ -29,6 +29,12 @@
     &temas;
     &stpeter;
     <revision>
+      <version>2.13.2</version>
+      <date>2024-08-30</date>
+      <initials>gk</initials>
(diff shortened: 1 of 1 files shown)


guusdk:
This PR is a direct result of a discussion that was had in the XSF Standards MUC room: https://logs.xmpp.org/xsf/2024-08-30#2024-08-30-02021f86c64bb4d4

PR #1398, opened 2024-10-29
XEP-0384 add an updated attribute to the device list

edhelas:
This allows XMPP clients to have a hint of which device bundle was updated since the last refresh. It greatly reduces the amount of bundles to refresh on login.

### DIFF
--- a/xep-0384.xml
+++ b/xep-0384.xml
@@ -58,6 +58,14 @@
     <jid>jabber@larma.de</jid>
   </author>
   <revision>
+    <version>0.9</version>
+    <date>2024-10-29</date>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Can you please raise this on the mailing list first? Personally I see no reason to refresh a bundle. You either already have a session with that device or you don’t.

Neustradamus:
@edhelas: Thanks!
It is a very important update.

edhelas:
After some explanations from the author it seems that I misunderstood some flow regarding how the OMEMO devices and keys are retrieved. This PR is not required anymore.

PR #1405, opened 2024-12-03
New ProtoXEP: Pubsub Stories

edhelas:
This new ProtoXEP extends XEP-0472: Pubsub Social Feed by defining a new Profile called Stories.

An HTML version is available here: https://movim.eu/xeps/stories.html


singpolyma:
This does not define any namespace, though it says it does. It seems to register a well known pubsub node with the intent to store atom content in it.

Also WRT stories, are you sure it should be limited to picture? I think eg video stories are a thing? Perhaps just one atom entry with a SHOULD contain an enclosure would be closer.

edhelas:
@singpolyma Thanks for the feedbacks !

PR #1406, opened 2024-12-06
Uniformize the registrar anchor on all XEPs

linkmauve:
This change is purely editorial, it makes all XEPs with an XMPP Registrar Considerations section all have the same anchor.


linkmauve:
Please bypass CI, I won’t make one commit per XEP just to satisfy it.

guusdk:
I _think_ I addressed the multiple-files thing in https://github.com/xsf/xeps/pull/1378

PR #1409, opened 2024-12-20
Compliance Suite 2024: copied from XEP-0479

sonnyp:
Consider using squash and merge

I know it's a bit late for 2024, but I'm planning on submitting a 2025 proposal and it might be confusing to have a missing year.


Neustradamus:
@sonnyp: I propose to rename "Compliance Suite 2025" :)

jcbrand:
Is there any difference between the compliance suites for 2023 and 2024? If not, I don't think we need a 2024 suite just to avoid a gap in the years.

PR #1411, opened 2024-12-22
XEP-0388: Fix failure example sequence

sonnyp:
The schema is as follow

```xml
  <xs:element name="failure">
    <xs:complexType>
      <xs:sequence>
        <xs:element name="text" type="xs:string" minOccurs="0"/>
        <xs:any namespace='##other' minOccurs='0' maxOccurs='unbounded' processContents='lax'/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
```

Which matches RFC 6120 similar stream error https://xmpp.org/rfcs/rfc6120.html#rfc.section.6.5

### DIFF
--- a/xep-0388.xml
+++ b/xep-0388.xml
@@ -27,6 +27,14 @@
   &tmolitor;
   &mwild;
   <revision>
+    <version>1.0.3</version>
+    <date>2024-12-22</date>
+    <initials>sp</initials>
+    <remark><ul>
+      <li>Fix failure example sequence.</li>
+    </ul></remark>
+  </revision>
+  <revision>
     <version>1.0.2</version>
     <date>2024-08-06</date>
     <initials>egp</initials>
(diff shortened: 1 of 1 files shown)


lnjX:
Hmm, but it looks like the XML schema of `failure` actually misses the reference to the SASL error condition element.

So the schema should actually look like this:
```xml
  <xs:element name="failure">
    <xs:complexType>
      <xs:sequence>
        <xs:choice>
          <xs:element ref="sasl:aborted"/>
          <xs:element ref="sasl:account-disabled"/>
          <xs:element ref="sasl:credentials-expired"/>
          <xs:element ref="sasl:encryption-required"/>
          <xs:element ref="sasl:incorrect-encoding"/>
          <xs:element ref="sasl:invalid-authzid"/>
          <xs:element ref="sasl:invalid-mechanism"/>
          <xs:element ref="sasl:malformed-request"/>
          <xs:element ref="sasl:mechanism-too-weak"/>
          <xs:element ref="sasl:not-authorized"/>
          <xs:element ref="sasl:temporary-auth-failure"/>
        </xs:choice>
        <xs:element ref="text" minOccurs="0" maxOccurs="1"/>
        <xs:any namespace='##other' minOccurs='0' maxOccurs='unbounded' processContents='lax'/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
```
Note that SASL error condition is a required and text and additional errors are optional.

```xml
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
           xmlns:sasl="urn:ietf:params:xml:ns:xmpp-sasl"
```
(and the reference to `sasl` is needed of course)

sonnyp:
> Council has decided that we should just fix the schema. Editor will do that.

https://github.com/xsf/xeps/commit/3728a980a025f5185b3bf668739532481133101f

PR #1413, opened 2025-01-08
XEP-0421 v0.3.0

mar-v-in:
- Adjust wording to use semi-anonymous or pseudonymous instead of anonymous
- Explicitly mention issues arising from occupant id matching across rooms
- Add example with server secret instead of room secret
- Add some pseudocode

Diff is here: https://larma.de/xeps/xep-0421-diff.html
Note that as part of not using the term anonymous anymore, I also adjusted the title to make it more descriptive.

### DIFF
--- a/xep-0421.xml
+++ b/xep-0421.xml
@@ -6,8 +6,8 @@
 <?xml-stylesheet type='text/xsl' href='xep.xsl'?>
 <xep>
 <header>
-  <title>Anonymous unique occupant identifiers for MUCs</title>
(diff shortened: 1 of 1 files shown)


mar-v-in:
This PR addresses feedback from the list. I don't think council is needed for the author to incorporate list feedback during Proposed status, but this case wasn't handled in the triage script. I created a PR to adjust the triage script to treat Proposed the same as Experimental. See #1414

PR #1423, opened 2025-01-26
XEP-0482: Add the XML Schema

linkmauve:


xep-0482.xml
```diff
@@ -204,4 +212,75 @@
...
+    <xs:documentation>
+      The protocol documented by this schema is defined in
+      XEP-0482: https://xmpp.org/extensions/xep-0482.html
+    </xs:documentation>
+  </xs:annotation>
+
+  <xs:element name='invite'>
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element ref='jingle'/>
```
mar-v-in, line 234: `<jingle/>` element is meant to be optional. This would implicitly be solved by changing to `<xs:sequence><xs:choice maxOccurs='unbounded'>`.

xep-0482.xml
```diff
@@ -204,4 +212,75 @@
...
+  <xs:annotation>
+    <xs:documentation>
+      The protocol documented by this schema is defined in
+      XEP-0482: https://xmpp.org/extensions/xep-0482.html
+    </xs:documentation>
+  </xs:annotation>
+
+  <xs:element name='invite'>
+    <xs:complexType>
+      <xs:sequence>
```
mar-v-in, line 233: I think `<xs:sequence><xs:choice maxOccurs='unbounded'>` might be more appropriate here.

While currently not specified, the idea was that the order of elements inside `<invite/>` indicates the sending clients priority for each of the ways to join the call, meaning a recipient client should pick the first option it deems appropriate, when of course the receiving user might be empowered to override this.

For regular calls this is likely always first `<jingle/>` and then `<external/>`, but for an multi-party audio call it might be as complex as `<meeting type='jitsi' /><external uri='https://jitsi...' /><jingle jid='mixer@jitsi...' /><external uri='tel:...' />` ,where the first two receive individual audio streams from each participant whereas the third and fourth go through a mixer that mixes them in a single audio stream and thus are inferior.

xep-0482.xml
```diff
@@ -204,4 +212,75 @@
...
+        <xs:element ref='jingle'/>
+        <xs:element ref='external' minOccurs='0' maxOccurs='unbounded'/>
```
mar-v-in, line 235: Maybe add explicit `<xs:any namespace='##other' />` here to indicate that this is explicitly intended as an extension point.

PR #1430, opened 2025-03-28
XEP-0421: Fix typo

Fishbowler:


Fishbowler:
I suspect that neither of these labels are true for this change :)

Kev:
> I suspect that neither of these labels are true for this change :)

FWIW, the Needs Version Block one was correct - every XEP update needs a new version block, with typos/editorial fixes being the third component.

PR #1436, opened 2025-04-26
XEP-0353/JMI: Fix specification and xml schema inconsistencies

lnjX:
Currently the schema says some elements MUST contain a '&lt;reason/&gt;' from `urn:xmpp:jingle-message:0` and the examples and specification text said that some elements SHOULD contain a `&lt;reason/&gt;` element from `urn:xmpp:jingle:1`.

This does:
 * SHOULD -> MUST contain reason
 * Use reason from `urn:xmpp:jingle:1` in both XML schema and examples

- **XEP-0353/JMI: Remove redefinition of jingle reason element in schema**
- **XEP-0353/JMI: Make usage of reason element a MUST**
- **XEP-0353/JMI: Add version block for 0.7.0**


lnjX:
Ping @tmolitor-stud-tu :)

lnjX:
Ping @fippo @stpeter @tmolitor-stud-tu

and @iNPUTmice :)

lnjX:
I see. :/ However, the XML schema already had `minOccurs=1 maxOccurs=1` before, so the schema already had a MUST before. This is inconsistent, but both conditions, those from the text and those from the XML schema need to be met, or am I wrong?

If so, then a `muji` element without the `reason` element was allowed by the text but not by the XML schema, so it was not valid.
If we also change the text to `MUST`, then this does not change the behavior, an element without the `reason` element is still invalid.

However, this is maybe a bit pedantic and I can totally understand if you say this leads to confusion and making the MUST from the schema a SHOULD would be better than the current situation. :)

stpeter:
What do the existing implementations generate and expect?

lnjX:
Sorry that this took so long, did the adjustments now. :)

PR #1438, opened 2025-05-03
XEP-0272/Muji: Use <content/> element from jingle & add schema

lnjX:
- **XEP-0272/Muji: Add jingle xmlns to <content/> elements**
- **XEP-0272/Muji: Add XML schema**
- **XEP-0272/Muji: Add revision block**

### DIFF
--- a/xep-0272.xml
+++ b/xep-0272.xml
@@ -39,6 +39,17 @@
   </author>
   &larma;
   <revision>
+    <version>0.2.1</version>
+    <date>2025-05-03</date>
+    <initials>lnj</initials>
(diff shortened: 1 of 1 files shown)


lnjX:
Ping @mar-v-in :)

mar-v-in:
Sorry for the delay. Indeed a good catch that the namespace is missing. I believe existing implementations still have it correct, right?

lnjX:
Hmm, only Movim (@edhelas), Dino and QXmpp advertise Muji in their DOAP file.

PR #1439, opened 2025-06-06
XEP-0485: Fixed references to XEP identifier.

guusdk:
The original specification referenced `XEP-0XXX` as no XEP number was assigned at the time of writing. This commit updates those references to XEP-0485.


guusdk:
GitHub automation added the label 'Needs Author'. As I am the editor and this is an editorial change, I have replaced that with 'Needs Editor'.

PR #1453, opened 2025-09-03
XEP-0070: extend security section

pulkomandy:
- Explain that the transaction identifier helps mitigating prompt bombing/fatigue attacks (where an attacker sends a lot of authentication requests and the user ends up accepting the wrong one)
- Add a note about the default login prompt in web browsers, that may mislead users into using their XMPP account password as a transaction identifier

### DIFF
--- a/xep-0070.xml
+++ b/xep-0070.xml
@@ -31,6 +31,12 @@
   &dizzyd;
   &stpeter;
   <revision>
+    <version>1.0.2</version>
+    <date>2025-09-30</date>
+    <initials></initials>
+    <remark>Mention risk of users entering their XMPP password instead of a token in security considerations.</remark>
+  </revision>
+  <revision>
     <version>1.0.1</version>
     <date>2016-12-09</date>
     <initials>mp (XEP Editor: ssw)</initials>


xep-0070.xml
```diff
@@ -348,6 +348,9 @@ Content-Length: 3032
   <section2 topic='End-to-End Encryption' anchor='security-e2e'>
     <p>For added security, the XMPP Server and XMPP Client may wish to communicate using end-to-end encryption. Methods for doing so are outside the scope of this proposal.</p>
   </section2>
+  <section2 topic='Misleading password prompt' anchor='security-password'>
+    <p>The default login prompt in most web browsers will ask the user for a password where they should instead enter a transaction identifier in the case of XMPP authentication. This may lead users to accidentally enter their XMPP account password as a transation identifier. Websites using this authentication method in an user facing way SHOULD NOT use the standard web browser login dialog, and let the user enter their credentials by other means.</p>
```
pulkomandy, line 352:
```suggestion
    <p>The default login prompt in most web browsers will ask the user for a password where they should instead enter a transaction identifier in the case of XMPP authentication. This may lead users to accidentally enter their XMPP account password as a transation identifier. Websites using this authentication method in an user facing way should avoid the standard web browser login dialog, and let the user enter their credentials by other means.</p>
```

pulkomandy:
OK, removed it then.

xep-0070.xml
```diff
@@ -348,6 +348,9 @@ Content-Length: 3032
   <section2 topic='End-to-End Encryption' anchor='security-e2e'>
     <p>For added security, the XMPP Server and XMPP Client may wish to communicate using end-to-end encryption. Methods for doing so are outside the scope of this proposal.</p>
   </section2>
+  <section2 topic='Misleading password prompt' anchor='security-password'>
+    <p>The default login prompt in most web browsers will ask the user for a password where they should instead enter a transaction identifier in the case of XMPP authentication. This may lead users to accidentally enter their XMPP account password as a transation identifier. Websites using this authentication method in an user facing way should avoid the standard web browser login dialog, and let the user enter their credentials by other means.</p>
```
Fishbowler, line 352:
```suggestion
    <p>The default login prompt in most web browsers will ask the user for a password where they should instead enter a transaction identifier in the case of XMPP authentication. This may lead users to accidentally enter their XMPP account password as a transation identifier. Websites using this authentication method in a user facing way should avoid the standard web browser login dialog, and let the user enter their credentials by other means.</p>
```

Tiny and unimportant nitpick of 'a' vs 'an'.

PR #1466, opened 2025-10-03
XEP-0060: Remove ambiguity in disco#info feature usage

guusdk:
This change addresses two issues:
1. Section 10 previously defined the `http://jabber.org/protocol/pubsub#publish` feature to be REQUIRED, while section 7.1.1 defines it to be optional.
2. The value `http://jabber.org/protocol/pubsub`, which isn't registered as a feature in this document or with the XMPP Registrar, should not be primarily used in examples of this XEP. Instead, add `http://jabber.org/protocol/pubsub#subscribe` (which is a required feature), keeping `http://jabber.org/protocol/pubsub` as a recommendation for compatiblity.

These changes have been discussed in the XMPP Standards chatroom (logs: https://logs.xmpp.org/xsf/2025-10-03#2025-10-03-f173fd843da7c3ee ) and on the standards mailinglist: https://mail.jabber.org/hyperkitty/list/standards@xmpp.org/thread/R4K2OAPVBQ4IC757F2YBZDV4PPHFGRNS/

### DIFF
--- a/xep-0060.xml
+++ b/xep-0060.xml
@@ -51,6 +51,15 @@
   &stpeter;
   &ralphm;
   <revision>
+    <version>1.27.0</version>
+    <date>2025-10-03</date>
+    <initials>gdk</initials>
+    <remark>
+      <p>Removed non-normative 'http://jabber.org/protocol/pubsub' disco#info feature from examples</p>
+      <p>Resolved ambiguity around the mandatory status of the 'publish' feature by no longer defining it to be REQUIRED.</p>
+    </remark>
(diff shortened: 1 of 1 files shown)


guusdk:
Given new insights on the mailinglist (subscribe doesn't seem to be required either) I'm closing this proposal while evaluating a better solution.

guusdk:
For posterity: this has been replaced by:
- https://github.com/xsf/xeps/pull/1467
- https://github.com/xsf/xeps/pull/1468

PR #1521, opened 2026-04-01
Various fixes for `xep.ent`

guusdk:
The changes in this PR all relate to the declaration of XML entities in `xep.ent`. 

The changes relate to removing duplicate entities, and adding two new ones.

With the changing of the duplicates, minor changes to published XEPs are introduced (notably for the stoxcore entity). I am not sure if we can blindly accept these changes, without a version bump of the affected XEPs.

I do believe that there is more work that can be done, which more likely leads to a version bump of affected XEPs. I've left that out of scope for this change, and raised a new issue for that work here: https://github.com/xsf/xeps/issues/1520


stpeter (commented):
Other than updating stoxcore, everything else looks fine to me.

PR #1525, opened 2026-04-05
ProtoXEP Submission: Emoji Markup

techmetx11:
Adds custom emojis in messages, by using XEP-0394 Message Markup. This was co-authored by @snit-guckfung

### DIFF
--- a/inbox/emoji-markup.xml
+++ b/inbox/emoji-markup.xml
@@ -7,7 +7,7 @@
 <xep>
     <header>
         <title>Emoji Markup</title>
-        <abstract>This specification leverages &xep0394; and &xep0447; or &xep385; to send custom emojis</abstract>
+        <abstract>This specification leverages &xep0394; and &xep0447; (or &xep0385;) to send custom emojis</abstract>
             &LEGALNOTICE;
         <number>xxxx</number>
         <status>ProtoXEP</status>
--- a/inbox/emoji-markup.xml
+++ b/inbox/emoji-markup.xml
@@ -7,7 +7,7 @@
 <xep>
     <header>
         <title>Emoji Markup</title>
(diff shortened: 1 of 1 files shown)


mar-v-in:
- SFS `<file-sharing>` can have an `id` attribute (see example in [§4.2](https://xmpp.org/extensions/xep-0447.html#multi-file)). You can reference the file sharing by id rather than the hash (makes things easier and shorter)
- Existing SFS implementations that don't support this specification will render the original body as a message with attachment(s) (see [§4.1](https://xmpp.org/extensions/xep-0447.html#body)), which is likely not intended. We probably need a new disposition type in SFS to accommodate for this kind of use (something that has a meaning of "don't display at all, leave interpretation to other specifications").

snit-guckfung:
Yeah, originally we were going to use SFS with the 'id' attribute, but I think that got generalised to just always using the hash when we also allowed SIMS.

techmetx11:
> SFS `<file-sharing>` can have an `id` attribute (see example in [§4.2](https://xmpp.org/extensions/xep-0447.html#multi-file)). You can reference the file sharing by id rather than the hash (makes things easier and shorter)

Yes, but SIMS doesn't have the same property, so we had to use a hash of the file. We're trying to maintain compatibility with SFS/SIMS/any file sharing XEP

> * Existing SFS implementations that don't support this specification will render the original body as a message with attachment(s) (see [§4.1](https://xmpp.org/extensions/xep-0447.html#body)), which is likely not intended. We probably need a new disposition type in SFS to accommodate for this kind of use (something that has a meaning of "don't display at all, leave interpretation to other specifications").

I thought no disposition set meant the file attachment would be completely invisible to the user? Then again, there is no defined default in the SFS XEP

PR #1531, opened 2026-04-24
New MUC: Fix typo in Bassanio

Fishbowler:
Should've done this yesterday, for Shakespeare's birthday.

_This PR assumes that @dwd made a typo, rather than made a joke I didn't understand_


dwd:
Clearly my Shakespare is rusty.

dwd:
But also, this should be fixed after Council voting - movable targets for a ProtoXEP isn't fun.

PR #1535, opened 2026-05-01
XEP-0463 v0.2.0

mar-v-in:
Use a dedicated child element of `<x/>` rather than namespaced attributes.

### DIFF
--- a/xep-0463.xml
+++ b/xep-0463.xml
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <!DOCTYPE xep SYSTEM 'xep.dtd' [
   <!ENTITY % ents SYSTEM 'xep.ent'>
-  <!ENTITY mav0 'urn:xmpp:muc:affiliations:0'>
+  <!ENTITY mav 'urn:xmpp:muc:affiliations:1'>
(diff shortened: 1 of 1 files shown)


singpolyma:
What is the motivation for this change? The namespaced attributes seemed like an elegant solution here... Certainly moreso than a childless element with just attributes?

mar-v-in:
@singpolyma people expressed they don't like this use of namespaced attributes.

Specifically, the previous version used an empty mav:since attribute inside the join presence to signal support for the specification, which is very confusing. Now the existence of the element signals support and the since attribute is optional.

Given that there is likely wish or need to add further things to the join presence to enable features (see MUC2/GC3 discussions), it would be elegant to use the same structure for all of them. However, namespaced attributes are limited in their capabilities (as they can't have structure), meaning they are not going to suffice all use cases other XEPs may have.

I agree that for the responses, the namespaced attributes were somewhat elegant as they added description to what the `<x/>` element contains. However if for the reasons above the join presence is changed to use an element, it would only be consistent to do the same elsewhere. Further, the `<x/>` element in presences/messages that update affiliations may (especially in a future with new MUC features) contain other things than the affiliation version refers to.

PR #1540, opened 2026-05-07
ProtoXEP (Emoji Markup): Removed outdated BoB reference

techmetx11:


mar-v-in:
@techmetx11 the proto XEP is currently being voted on in the council (there is a 2 weeks voting window). After the voting is completed, it is likely to become a XEP with a number and you would then need to file changes against that XEP, not against the proto XEP in the inbox.

techmetx11:
Oh, well I'll keep the changes for if it becomes a XEP then

PR #1544, opened 2026-06-02
XEP-0301: Update STOX core reference to RFC 7247

edwtie:
## Summary

This updates XEP-0301's STOX core reference from the old draft-ietf-stox-core document to the published RFC 7247.

XEP-0301 section 9.1 uses `&stoxcore;` when discussing guidance for XMPP/SIP address translation and conveyance. Peter pointed out that this should reference RFC 7247 instead of the old draft, so this PR updates the shared entity while keeping the existing entity usage intact.

## Authorship

This also adds Edward Tie / Tiedragon as a later author for this maintenance update. The original authors remain first in the author list; Edward is added after them for the current update work.

## Compatibility note

The shared `&stoxcore;` entity currently expands in XEP-0301 and the SoX ProtoXEP. Both uses refer to SIP/XMPP address translation or mapping, which is the same subject area covered by the final STOX core RFC.

## Notes

- RFC 7247 is the published version of the STOX core architecture document.
- Updating the shared entity also updates any other XEP/ProtoXEP text that expands the same reference.

## Validation

- `git diff --check` passed.
- GitHub XEP validation passed.
- Local `make .xep-0301.check.ok` could not be run because `make` is not installed on this Windows environment.


edwtie:
Authorship note updated: this PR now adds Edward Tie / Tiedragon as a later author for this maintenance update, while keeping the original XEP-0301 authors first. The change is submitted for XEP Editor / Approving Body review together with the RFC 7247 reference update requested for XEP-0301 section 9.1.

guusdk:
Thanks for this! We should consider making this change part of a larger effort: there's more updates that are likely required, as documented in #1520

One concern that we should be watchful of is that these entity definitions are re-used across all XEPs. It is unlikely that we want to keep one XEP on an older version of the specification, while we want another on a newer version, but still, we should check.

Additionally, we ideally should make plausible that the updated RFC is still compatible with how they're used by the XEPs that currently depend on the older version/draft of the RFC. There may, after all, be material changes in the update of the RFC.

edwtie:
Thanks, that makes sense.

I checked the current `&stoxcore;` expansion sites in this tree. They are limited to:

- `xep-0301.xml`: SIP/XMPP address translation and conveyance guidance in the RFC 4103/T.140 interoperability section.
- `inbox/sox.xml`: SIP/XMPP address mapping guidance.

Both uses point at the same subject area covered by the final STOX core document: architecture, addresses, and error handling for SIP/XMPP interworking. The IETF Datatracker also marks `draft-ietf-stox-core-11` as the older Internet-Draft that was ultimately published as RFC 7247, so I think the shared entity update is appropriate here.

If you prefer, I can narrow this PR to avoid the shared entity and instead update only the XEP-0301 text, but that would leave the SoX text pointing at the stale draft even though it uses the same SIP/XMPP address-mapping reference.

guusdk:
I am not the boss of anything, but instead of _narrowing_ the PR, I'd love to see it be used to replace _more_ outdated RFC drafts.

My uncertainty is around process. Is this an editorial change? Does this require approval from each XEP author that is affected?

PR #1549, opened 2026-06-23
XEP-0420: First bigger update with implementation experience and author addition

Syndace:
Hi!

With the experience of implementing SCE twice, I have prepared this update. It mostly fills out TODOs that were left in the XEP, clarifies a few details and removes one or two warnings that turned out not to be relevant in practice.

I don't believe a namespace bump should be required, given the experimental nature of this XEP and that none of the tag names or broad semantics were changed.

Additionally, I have added myself as an author with the original author's approval, as Paul isn't as active in the XMPP community any more nowadays.

### DIFF
--- a/xep-0420.xml
+++ b/xep-0420.xml
@@ -31,9 +31,10 @@
   <supersededby/>
   <shortname>SCE</shortname>
   &paulschaub;
+  &syndace;
   <revision>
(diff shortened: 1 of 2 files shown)


vanitasvitae:
Looks good to me. I'm fine with @Syndace taking over co-authorship :)

PR #1551, opened 2026-07-03
XEP-0424: Modifications in response to Last-Call feedback.

jcbrand:
* Clarify that messages whose user-visible content lies outside the <body> element (e.g. XEP-0447 Stateless File Sharing, where the body is only a fallback) have a messaging payload and may be retracted.
* Fix leftover reference to "origin ID" in the Use Case section (now: message ID).
* State that a retraction message's fallback body SHOULD carry a XEP-0428 fallback indication, and that supporting clients MUST NOT render the body of a retraction message, including when no message matching the retraction 'id' is found.
* Add security considerations regarding sender-controlled fallback bodies.
* Add the ability to advertise a retraction time limit via a XEP-0128 'max-age' field.
* Upgrade the XEP-0334 <store/> hint from MAY to SHOULD, to ensure retractions are archived also by services that are unaware of this specification.

Feedback from Goffi:
https://mail.jabber.org/hyperkitty/list/standards@xmpp.org/message/YWS2IIEP5KFDQ7ILZZBQWDUXHD2FHGFH/

Feedback from Daniel:
https://mail.jabber.org/hyperkitty/list/standards@xmpp.org/message/SVY7PHKUKAS5FJCKXIP7JEO4AOGOQAWI/


jcbrand:
This XEP is not experimental (it's "proposed") and I am an author of the XEP, so the github-actions bot is doubly wrong.

PR #1552, opened 2026-07-08
XEP-0143: Add author responsibility to submission process

guusdk:
Adds a requirement that authors understand, verify, and take responsibility for every part of their contribution.

Suggested by Ralph Meijer on the standards list, in the context of the discussion on AI-assisted contributions. The wording is deliberately not specific to AI: the obligation applies regardless of the tools an author uses.

This does not attempt to settle the wider discussion about AI use in specification authoring.

### DIFF
--- a/xep-0143.xml
+++ b/xep-0143.xml
@@ -23,6 +23,12 @@
   <shortname>N/A</shortname>
   &stpeter;
   <revision>
+    <version>1.1.4</version>
+    <date>2026-07-08</date>
+    <initials>gdk</initials>
+    <remark><p>Add author responsibility to the submission process.</p></remark>
+  </revision>
+  <revision>
     <version>1.1.3</version>
     <date>2024-12-24</date>
     <initials>dwd</initials>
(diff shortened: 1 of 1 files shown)


silverhook:
IMHO this looks like a good addition.

I have seen something in the lines of “professional responsibility” in other guidelines, but cannot recall the exact wording or where I’ve seen it, before.

The closest I could found though is this CoC I helped draft some years ago – I think several points there coincide with the AI (and other) slop concerns that are being discussed (happy to take this to the mailing list, if it’s the better place).
https://commonsconservancy.org/dracc/0015/

goffi-contrib:
Hi @guusdk . Thanks for you contribution. I'm not sure about the wording "take responsibility", it seems a bit vague to me. Does that means responsible of legal liability? Plagiarism of existing specs? Or responsible if things are not working? Something else?

guusdk:
@goffi-contrib this was intended as taking responsibility as the author of the document (as opposed to waiving off responsibility because it was generated by AI, or somesuch). I'm open to better words. That discussion is probably best done on the mailing list though.

goffi-contrib:
Sure, moving my comment to standard@. With my council hat, I'll wait for feedback before voting on this PR. Thanks.

goffi-contrib:
https://mail.jabber.org/hyperkitty/list/standards@xmpp.org/message/EN3X4VWEIJ52BGLHID7IP5BMQSGLATNA/

PR #1557, opened 2026-07-29
Fix bind2 namespace

mdosch:
XEP-0386 defines it as `urn:xmpp:bind:0`:
https://xmpp.org/extensions/xep-0386.html#sect-idm45625558119104


iNPUTmice:
Please add a version block even if this is only editorial.

PR #1562, opened 2026-09-11
New ProtoXEP: SASL2 Terms of Service Acceptance Task

guusdk:
Adds inbox/sasl2-tos.xml, a draft SASL2 task (XEP-0388 §2.5) for gating authentication on terms-of-service acceptance.

Cross-posted to standards@ for discussion. This is an early draft, and feedback on the basic direction is as welcome as wording nits.

### DIFF
--- a/inbox/sasl2-tos.xml
+++ b/inbox/sasl2-tos.xml
@@ -23,7 +23,7 @@
     <shortname>NOT_YET_ASSIGNED</shortname>
     &gdk;
     <revision>
-      <version>0.1.2</version>
+      <version>0.1.3</version>
       <date>2026-09-12</date>
       <initials>gdk</initials>
       <remark><p>First draft.</p></remark>
@@ -110,6 +110,7 @@
       <li>A server MUST NOT complete SASL2 authentication for an account for which this task applies (see <link url='#applicability'>Determining Applicability</link>) unless the task has completed successfully during the same negotiation.</li>
       <li>A server MUST offer this task on every anonymous authentication attempt for which it applies (see <link url='#applicability'>Determining Applicability</link>), since acceptance cannot be reliably recorded against an account that does not exist.</li>
       <li>A server SHOULD treat "currently accepted version" as a per-account fact rather than a per-session one: once durably recorded, the same account is not asked again on a later connection unless the applicable version changes or no durable record was actually made (see <link url='#outcome'>Outcome</link>).</li>
+      <li>A server MAY treat acceptance obtained through a channel other than this task (for example a support interaction or a signed agreement) as satisfying acceptance of a version, provided it is recorded per-account in the same way. This task exists to enforce acceptance for accounts that have not already accepted through such a channel, not to be the sole route by which acceptance can be recorded.</li>
       <li>A server SHOULD treat the <tt>version</tt> identifiers it issues as opaque and SHOULD compare them for exact equality, both when checking whether a previously-recorded acceptance is still current and when validating a peer's <tt>&lt;accept/&gt;</tt> against the version it was offered in the same round.</li>
     </ol>
   </section1>
--- a/inbox/sasl2-tos.xml
+++ b/inbox/sasl2-tos.xml
@@ -23,8 +23,8 @@
     <shortname>NOT_YET_ASSIGNED</shortname>
     &gdk;
     <revision>
-      <version>0.1.0</version>
-      <date>2026-09-11</date>
+      <version>0.1.1</version>
+      <date>2026-09-12</date>
       <initials>gdk</initials>
       <remark><p>First draft.</p></remark>
     </revision>
@@ -41,7 +41,7 @@
     <ol>
       <li>A server MUST be able to prevent an account from completing authentication until the current version of its terms of service has been accepted by that account.</li>
       <li>A server SHOULD be able to determine, per account, whether the currently-applicable version has already been accepted, so that a user who has already accepted is not asked again.</li>
-      <li>A client that does not implement this specification MUST be able to determine, on abort, that its inability to proceed is due to a terms-of-service requirement, rather than an unrelated authentication failure, so that it can inform its user rather than simply reporting a failed login.</li>
+      <li>A server MUST provide a human-readable explanation of why authentication has not yet completed, so that a client that does not implement this specific task, but does implement SASL2 authentication, can still inform its user rather than simply reporting a failed login.</li>
     </ol>
   </section1>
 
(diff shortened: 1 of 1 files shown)


inbox/sasl2-tos.xml
```diff
@@ -0,0 +1,184 @@
...
+    <p>This document defines only the task itself: its namespace, its elements, and the rules governing when it applies and how its outcome is determined. How a task is offered, selected, and driven to completion or failure in general is defined by &xep0388;.</p>
+  </section1>
+
+  <section1 topic='Requirements' anchor='reqs'>
+    <ol>
+      <li>A server MUST be able to prevent an account from completing authentication until the current version of its terms of service has been accepted by that account.</li>
+      <li>A server SHOULD be able to determine, per account, whether the currently-applicable version has already been accepted, so that a user who has already accepted is not asked again.</li>
+      <li>A client that does not implement this specification MUST be able to determine, on abort, that its inability to proceed is due to a terms-of-service requirement, rather than an unrelated authentication failure, so that it can inform its user rather than simply reporting a failed login.</li>
```
Fishbowler: How would a client comply with this requirement without implementing this specification?
Would a client that was "TOS aware" be able to shrug off the rest of the spec, say "I can't display the TOS, sorry", and declare compliance in the DOAP?

I think what I'm saying is: is the inability to comply with a task an omission from the parent XEP rather than something that should be dealt with at a Task level?

inbox/sasl2-tos.xml
```diff
@@ -0,0 +1,184 @@
...
+    </ol>
+  </section1>
+
+  <section1 topic='Protocol' anchor='protocol'>
+    <p>This specification defines the <tt>urn:xmpp:tos:0</tt> namespace, used to qualify the elements defined below, and one SASL2 task, identified in <tt>&lt;task/&gt;</tt> and <tt>&lt;next/&gt;</tt> elements by the token <tt>TOS-ACCEPT</tt>.</p>
```
Fishbowler, line 49: Vague nitpick: Should the namespace include sasl?

inbox/sasl2-tos.xml
```diff
@@ -0,0 +1,184 @@
...
+      <p>On receiving the client's response, the server determines the outcome per the following rules:</p>
+      <ol>
+        <li>If the response contains no <tt>&lt;accept/&gt;</tt> element, the server MUST fail the SASL2 negotiation with a <tt>&lt;not-authorized/&gt;</tt> condition.</li>
+        <li>If the response contains an <tt>&lt;accept/&gt;</tt> element whose <tt>version</tt> attribute does not match the version that was offered in this round, the server MUST fail the SASL2 negotiation with a <tt>&lt;malformed-request/&gt;</tt> condition (rejecting, for example, a value cached from an earlier attempt).</li>
+        <li>Otherwise, the task completes for this negotiation. The server SHOULD persist the acceptance against the account, so that it is not asked again on a future connection. If the account is authenticating anonymously, or the acceptance cannot be persisted (for example due to a transient failure of the server's own backing store), no durable record results; the task still completes, but the server SHOULD offer it again on the account's next authentication attempt, since there is nothing on record to rely on next time.</li>
```
Fishbowler, line 102: Should this section specifically call out (again) the version of the terms that the acceptance was made against?

Fishbowler: Similarly, should this allow that a server MAY offer an external route to accept and persist acceptance of a version of these terms (e.g. 1:1 message, email, etc)

inbox/sasl2-tos.xml
guusdk: I don't really know. Not doing that would make it reusable for related (non-SASL2) functionality, which may be desirable or may not at all, depending on how you look at it. I've got no strong argument either way.

inbox/sasl2-tos.xml
guusdk: Agreed. DOAP compliance shouldn't be claimable on this alone. But more fundamentally, the requirement itself was wrong to be phrased as binding a client at all: a spec can't obligate a party who doesn't implement it. It's really a server requirement, aimed at making a user of a non-compliant client understand why authentication is failing. I've added a change to reflect this.

inbox/sasl2-tos.xml
guusdk: On the version: I think this is already implicit. Rule 3 says the task applies "if the account has not previously accepted the version that is currently applicable," which does tie acceptance to a specific version rather than a bare yes/no flag, but it's easy for an implementer to read past this, and cheap to guard against. I have now added improved words.

On an external acceptance route: I think this is already allowed as written, just not stated. Rule 3 only checks whether the account has accepted the applicable version. It doesn't require that acceptance to have come through this specific task. A server that recorded acceptance via, say, a support interaction, and stored it the same way, would already cause this task to be skipped. I have now add an explicit business rule to that effect, since otherwise an implementer might reasonably assume this task is the only legitimate channel, when it's meant as a fallback for accounts that haven't accepted some other way.

inbox/sasl2-tos.xml
Fishbowler: Nice

inbox/sasl2-tos.xml
Fishbowler: Fair

inbox/sasl2-tos.xml
Fishbowler: Thanks!

singpolyma:
This seems like it would be well described by a data form. A piece of text and a boolean that is submitted. Which leads me to think if a more generic FORM task would serve the purpose. The user must fill out this form to continue.

guusdk:
> This seems like it would be well described by a data form. A piece of text and a boolean that is submitted. Which leads me to think if a more generic FORM task would serve the purpose. The user must fill out this form to continue.

Valid suggestions, but I deliberately chose different.

I prefer to tie acceptance of TOS into authentication (hence the SASL2 Task-based architecture). Other routes certainly exist (there are even other protoxeps that define those), but those, in my opinion, leave wiggle room that I think using SASL2 rules out (granted, at the expense of other characteristics).

A dataform would be nice, but I don't think that those are currently specified in SASL2 (whether or not they should is a distinct matter). For the purpose of accepting or rejecting the TOS, the mechanic provided by SASL2 today suffices: it is a very binary accept/reject choice that's baked into the mechanism already. There's no _need_ for a dataform there, and it's unlikely that in the common case, more options than the binary accept/reject are needed. 

This first version of the specification uses a URL to reference the TOS text, rather than inlining the TOS texts in the data exchange. Of all things described here, I am least married to this approach. Inlining the TOS text could be valuable and has undeniable benefits. The choice to go with a URL was based on the facts that TOS texts tend to be large, and we'd likely want to be flexible with media types when requesting them. Using HTTP gives a mechanism that allows a client to request the most appropriate representation (plaintext, html, pdf, ...) and prevents the need for such negotiation to happen in SASL2 or to have the server send all available representations.

singpolyma:
What I propose would still be a SASL2 task and still bound to authentication, etc, as you say. A dataform is "just" XML so specifying a task that uses one is ultimately the same as what's here, just slightly different namespace/syntax

I do think inlining some text needs to at least be an option, though if some prefer to include just a link that's up to their local policy of course.

